PR #10, opened 2022-01-18
Bugfixes handling of MatchingRequestParameter in response-prefixes

floroks:
When a positive or negative response used MatchingRequestParameter,
the parameter was ignored, and the resulting calculated prefixes
for the responses were incomplete.

This caused the wrong service to be matched when using find
for a response in these cases (e.g. if a RecordDataIdentifier
was used for a ReadDataByIdentifier-response)


KLBa (approved):
Looks good, thank you for your contribution!

PR #11, opened 2022-01-18
Read unit spec and add CI

KLBa:
This PR ...

* adds a GitHub workflow that runs unit tests on every push and pull request
* removes git lfs
* adds handling of the odx element `DIAG-DATA-DICTIONARY-SPEC/UNIT-SPEC`. This includes:
  * Reading of `UNIT-SPEC` from odx
  * Adding the `UNIT-SPEC` to the pdx template
  * Extending the somersault ECU to use units

  Examples:

  Access the `UNIT-SPEC` directly:
  ```python
  >>> import odxtools
  >>> db = odxtools.load_pdx_file("examples/somersault.pdx")
  >>> db.diag_layers.somersault.local_diag_data_dictionary_spec.unit_spec
  UnitSpec(unit_groups=[european_duration], units=[second, minute], physical_dimensions=[second])
  ```
  Access the unit via a DOP that references it:
  ```python
  >>> dop = db.ecus.somersault_assiduous.services.headstand.request.parameters.duration.dop
  >>> dop.unit  # Resolved reference to the unit
  Unit(id='somersault.unit.second', short_name='second', display_name='s', oid=None, long_name='Second', description='<p>SI unit for the time</p>', factor_si_to_unit=1.0, offset_si_to_unit=0.0, physical_dimension_ref='somersault.PD.second')
  ``` 

Katrin Bauer <[katrin.b.bauer@daimler.com](mailto:katrin.b.bauer@daimler.com)>, on behalf of [MBition GmbH](https://mbition.io/).
[Provider Information](https://github.com/Daimler/daimler-foss/blob/master/PROVIDER_INFORMATION.md)


odxtools/units.py
```diff
@@ -0,0 +1,290 @@
+# SPDX-License-Identifier: MIT
+# Copyright (c) 2021 MBition GmbH
```
andlaus, line 2: 2022?!

odxtools/units.py
```diff
@@ -0,0 +1,290 @@
...
+
+    ```
+    Unit(
+        id="kilometre",
+        short_name="kilometre",
+        display_name="km"
+    )
+    ```
+
+    Additionally, a unit may reference a SI unit (`.physical_dimension`)
```
andlaus: This is not a big deal, since I frequently observe native speakers getting this wrong: I guess this should be "an SI unit". Like any other natural language English has weird quirks, in this case it's that if something is *pronounced* as starting with a vowel, "an" is used, if it is not, "a" will win out even if the word in question starts with a vowel. Examples: "This is an SI unit" but "a unit may reference".

andlaus (commented):
neat! I just have two tiny nitpicks..

odxtools/units.py
KLBa: Good point! Should we changed that in all files actually?

odxtools/units.py
KLBa: Thanks!

andlaus:
ok, now I don't even have nitpicks anymore :). merging...

odxtools/units.py
andlaus: I guess only when they get modified. Anyway, I think the factual legal relevance of this comment is (close to?) zero anyway...

PR #14, opened 2022-01-20
Add PhysicalType and rework handling of data types

KLBa:
This PR does 3 main things (where the first two cause a lot of commit noise):

* Add an enum for data types, i.e., magical strings like `"A_UINT32"` are replaced by `DataType.A_UINT32`
  * The enum also offers methods to `cast` values, so the dicts `ODX_TYPE_PARSER` and `ODX_TYPE_TO_PYTHON_TYPE` shouldn't be used anymore
* Add a class `PhysicalType` for `PHYSICAL-TYPE` elements.
  Previously the physical type of a DOP was just represented by the string for the data type (e.g. `dop.physical_data_type == "A_UINT32"`). Besides the data type, odxfiles can also specify a display radix or precision in the physical type.

This does make accessing the data type a bit more complex, i.e., the previous
```python
dop.physical_data_type
```
is now
```python
dop.physical_type.base_data_type.value
```

* Additionally, this PR fixes the computation of `LinearCompuMethod.physical_limit`s when the interval type of the internal limit is `OPEN` and the data type is an integer. In these cases, the physical limit now has the interval type `CLOSED` and only complies to physical values that map to valid internal values.

Katrin Bauer <[katrin.b.bauer@daimler.com](mailto:katrin.b.bauer@daimler.com)>, on behalf of [MBition GmbH](https://mbition.io/).
[Provider Information](https://github.com/Daimler/daimler-foss/blob/master/PROVIDER_INFORMATION.md)

### DIFF
--- a/examples/somersaultecu.py
+++ b/examples/somersaultecu.py
@@ -27,6 +27,7 @@ from odxtools.dataobjectproperty import DataObjectProperty
 from odxtools.diagdatadictionaryspec import DiagDataDictionarySpec
 
 from odxtools.diagcodedtypes import StandardLengthType
+from odxtools.physicaltype import PhysicalType
 
 from odxtools.units import UnitSpec
 from odxtools.units import PhysicalDimension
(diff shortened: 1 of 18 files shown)


andlaus:
looks good, thanks!

PR #18, opened 2022-01-28
Feature: add pre conditions to find

froks:
Adds display of state-preconditions to the find command
Adds enabled audiences to the find command
Makes DiagService hashable, so it can be used in a dict (used to group services by variants)

--

Florian Roks \<florian.roks@daimler.com\>, Daimler TSS GmbH
[Imprint](https://github.com/Daimler/daimler-foss/blob/master/LEGAL_IMPRINT.md)


froks:
Used wrong github account, will recreate with the correct one

PR #21, opened 2022-02-05
Fix some encode/decode logic operations

kayoub5:


andlaus (commented):
Thanks for the fixes, they look good to me, agreed @KLBa?

before we can merge this, can you please send in a signed [CLA](https://github.com/mercedes-benz/odxtools/blob/main/CONTRIBUTING.md#contributions)? (this hopefully will not take longer than five minutes.)

andlaus:
(to explain this a bit: the CLA is only required because internal Mercedes-Benz regulations. For MIT-licensed projects like `odxtools`, it IMO does not really make sense to demand it because the MIT license seems to be *more* liberal than the CLA, but IANAL...)

odxtools/diagcodedtypes.py
```diff
@@ -315,6 +315,8 @@ def convert_internal_to_bytes(self, internal_value, encode_state: EncodeState, b
             return value_byte
         else:
             termination_char = self.__termination_character()
+            if self.termination == 'END-OF-PDU':
+                termination_char = bytes()
```
KLBa, line 319: Do you have an example where this is needed?

I thought a parameter with `termination == 'END-OF-PDU'` must be the last parameter in a request. In this case the flag `encode_state.is_end_of_pdu` should be True and this else-branch is never reached.
In this case, I would prefer to raise an error here, as this shouldn't happen..
(It's entirely possible that I misunderstand the usage of `MinMaxLengthType`s though)

odxtools/parameters.py
```diff
@@ -407,18 +407,17 @@ def get_coded_value_as_bytes(self, encode_state: EncodeState):
...
         if phys_val != self.physical_constant_value:
             raise DecodeError(
                 f"Physical constant parameter does not match! "
                 f"The parameter {self.short_name} expected physical value {self.physical_constant_value!r} but got {phys_val!r} "
-                f"at byte position {decode_state.next_byte_position} "
+                f"at byte position {next_byte_position} "
                 f"in coded message {decode_state.coded_message.hex()}."
             )
-        return decode_state
+        return decode_state, next_byte_position
```
KLBa, line 420: Here `decode_state` should be `phys_val` too

KLBa (commented):
Thanks for the fixes!
I just added two comments :)

odxtools/diagcodedtypes.py
kayoub5: I don't know what exactly happened, but I do know that with the pdx I had the parser reached this point with `termination == 'END-OF-PDU'` and `termination_char == None`.

odxtools/parameters.py
kayoub5:
```suggestion
        return phys_val, next_byte_position
```

odxtools/diagcodedtypes.py
KLBa: Hm, ok, I would be interested in an example (ideally a unit test case) because I can't reproduce it..
That shouldn't stop this from being merged though.

KLBa (approved):

KLBa:
> Hi, It seems the mailbox [cla-mbition@daimler.com](mailto:cla-mbition@daimler.com) no longer exists, who should I send the email to?

Sorry for the confusion, please send it to me: <katrin.b.bauer@daimler.com>

kayoub5:
> > Hi, It seems the mailbox [cla-mbition@daimler.com](mailto:cla-mbition@daimler.com) no longer exists, who should I send the email to?
> 
> Sorry for the confusion, please send it to me: [katrin.b.bauer@daimler.com](mailto:katrin.b.bauer@daimler.com)

sent

KLBa:
> > > Hi, It seems the mailbox [cla-mbition@daimler.com](mailto:cla-mbition@daimler.com) no longer exists, who should I send the email to?
> > 
> > 
> > Sorry for the confusion, please send it to me: [katrin.b.bauer@daimler.com](mailto:katrin.b.bauer@daimler.com)
> 
> sent

Thank you!

PR #24, opened 2022-02-21
somersaultlazy: Convert it to the 'isotp' module

andlaus:
It turns out that with a few tricks this module can also be used in conjunction with asyncio. Compared to aioisotp, we get much better support for configuring the parameters of the in-kernel isotp subsystem, e.g., being able to instruct it to pad the last CAN frame of an ISO-TP telegram to 8 bytes. On the flipside, we now require a relatively recent linux kernel (>= 5.10), and we lose the user space ISO-TP implementation, i.e. the unittest mode is now done using custom code that works around the `isotp` module instead of simply using a "virtual" bus from `python-can`.

Andreas Lauser &lt;<andreas.lauser@daimler.com>&gt;, on behalf of [MBition GmbH](https://mbition.io/).


KLBa (approved):
Looks good. Special thanks for the sterile mode that can be used in WSL!

PR #26, opened 2022-02-22
support float COMPU-NUMERATOR

yassin92:
referring to ODX spec COMPU-NUMERATOR could be a float.

### DIFF
--- a/odxtools/compumethods/linearcompumethod.py
+++ b/odxtools/compumethods/linearcompumethod.py
@@ -82,7 +82,7 @@ class LinearCompuMethod(CompuMethod):
                                               IntervalType.INFINITE)
 
         assert self.internal_lower_limit is not None and self.internal_upper_limit is not None
-        assert denominator > 0 and type(denominator) == int or type(denominator) == float
+        assert denominator > 0 and isinstance(denominator, (int, float))
 
         self.__compute_physical_limits()
 
--- a/odxtools/compumethods/linearcompumethod.py
+++ b/odxtools/compumethods/linearcompumethod.py
@@ -82,7 +82,7 @@ class LinearCompuMethod(CompuMethod):
                                               IntervalType.INFINITE)
 
         assert self.internal_lower_limit is not None and self.internal_upper_limit is not None
-        assert denominator > 0 and type(denominator) == int
+        assert denominator > 0 and type(denominator) == int or type(denominator) == float
 
         self.__compute_physical_limits()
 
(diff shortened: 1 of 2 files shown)


odxtools/compumethods/readcompumethod.py
```diff
@@ -48,8 +48,14 @@ def _parse_compu_scale_to_linear_compu_method(scale_element,
...
     factor = computation_python_type(next(nums).text)
     if coeffs.find("COMPU-DENOMINATOR/V") is not None:
-        kwargs["denominator"] = int(
-            coeffs.find("COMPU-DENOMINATOR/V").text)
+        try:
+            kwargs["denominator"] = int(
+                coeffs.find("COMPU-DENOMINATOR/V").text)
+        except ValueError as ex:
+            pass
+        finally:
```
kayoub5, line 56: the float parsing needs to be in the catch block, not in the finally
```suggestion
```

kayoub5:
@KLBa CLA sent by Technica Engineering GmbH apply to this PR as well since @yassin92 is an employee there.

odxtools/compumethods/readcompumethod.py
andlaus: I'd like to propose something like this:
```python
        denom_elem = coeffs.find("COMPU-DENOMINATOR/V")
        kwargs["denominator"] = int(denom_elem.text) if denom_elem is not None else 1
```

kayoub5: the point of the pr is to address the fact that the value could be a float instead of an int, not sure how the default 1 value helps with that.

andlaus:
>  the point of the pr is to address the fact that the value could be a float instead of an int, not sure how the default 1 value helps with that.

Yeah, I was a bit lost in diffs. my real question is the following, though: Is there a reason not to use `float()` instead of `int()` unconditionally?

kayoub5:
> > the point of the pr is to address the fact that the value could be a float instead of an int, not sure how the default 1 value helps with that.
> 
> Yeah, I was a bit lost in diffs. my real question is the following, though: Is there a reason not to use `float()` instead of `int()` unconditionally?

for us: not really. for you: backward compatibility (especially the assert part)

andlaus:
> for us: not really. for you: backward compatibility (especially the assert part)

for us, `float()` is okay as well (it most likely won't cause any troubles and fixing the screwups will most likely be quite simple anyway), and it would make the code simpler, so I'm in favor for this...

odxtools/compumethods/linearcompumethod.py
```diff
@@ -82,7 +82,7 @@ def __init__(self,
                                               IntervalType.INFINITE)
 
         assert self.internal_lower_limit is not None and self.internal_upper_limit is not None
-        assert denominator > 0 and type(denominator) == int
+        assert denominator > 0 and type(denominator) == int or type(denominator) == float
```
kayoub5, line 85:
```suggestion
        assert denominator > 0 and isinstance(denominator, (int, float))
```

kayoub5:
@andlaus PR updated

andlaus:
cool, thanks. all good now, merging.

PR #28, opened 2022-03-04
Do not raise an exception when encountering an unknown trouble code

andlaus:
Sometimes it happens that the PDX files are not 100% complete and they miss a DTC which the ECU can occasionally send. Bailing out in this case makes dealing with reading such errors quite cumbersome, so let's refrain from this.

The approach taken for these DTCs is to simply provide the trouble code number without any additional context. (An external tool can then be used to it interpret it.)


KLBa (approved):
Looks good, merging.

PR #39, opened 2022-04-23
Better DESC extraction

kayoub5:
Current implimentation fails for sample below DESC, since the attribute will cause `raw_string.startswith("<DESC>")` to fail

```
<DESC TI="some_attribute_value">some value <p> other value</p> end</DESC>
```

Solution based on https://stackoverflow.com/a/3446055/1438522

### DIFF
--- a/odxtools/utils.py
+++ b/odxtools/utils.py
@@ -12,7 +12,9 @@ def read_description_from_odx(et_element: ElementTree.Element):
     if et_element is None:
         return None
  
-    raw_string = (et_element.text or '') + ''.join(ElementTree.tostring(e, encoding='unicode') for e in et_element)
+    raw_string = et_element.text
+    for e in et_element:
+        raw_string += ElementTree.tostring(e, encoding='unicode')
 
     return raw_string.strip()
 
--- a/odxtools/utils.py
+++ b/odxtools/utils.py
@@ -4,8 +4,6 @@
 from typing import Dict, Literal
 from xml.etree import ElementTree
 
-def content(tag):
-    return 
 
 def read_description_from_odx(et_element: ElementTree.Element):
     """Read a DESCRIPTION element. The element usually has the name DESC."""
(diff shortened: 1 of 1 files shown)


odxtools/utils.py
```diff
@@ -11,15 +11,9 @@ def read_description_from_odx(et_element: ElementTree.Element):
     #       This just represents it as XHTML string. 
     if et_element is None:
         return None
+ 
+    raw_string = (et_element.text or '') + ''.join(ElementTree.tostring(e, encoding='unicode') for e in et_element)
```
andlaus, line 15: hm, I guess this is a good change, but I think that statement is quite hard to read and the `or` in `(et_element.text or '')` is quite superfluous. would you mind changing it to something like this?
```suggestion
    raw_string = et_element.text
    for e in et_element:
        raw_string += ElementTree.tostring(e, encoding='unicode')
```

odxtools/utils.py
```diff
@@ -11,15 +11,11 @@ def read_description_from_odx(et_element: ElementTree.Element):
     #       This just represents it as XHTML string. 
     if et_element is None:
         return None
+ 
+    raw_string = et_element.text
```
kayoub5, line 15:
```suggestion
    raw_string = et_element.text or ''
```

odxtools/utils.py
kayoub5: ` or ''` is required since `et_element.text` could be None, and concatenation of None and a string is not valid

andlaus:
nice! thanks for the contribution :)

odxtools/utils.py
andlaus: Ok, fair enough. I just thought that all ElementTree nodes come with a `.text` attribute. maybe not...

andlaus: Ah, I have a suspicion: The type checker probably declares the `.text` attribute as `Optional[str]` because for tags without content like `<FOO />` it is (probably) `None`. I suppose not having the `or ''` would thus not bite at runtime but I think it is good to have it to keep `mypy` happy...

PR #53, opened 2022-08-17
Support COMPU-NUMERATOR with only one V

kayoub5:
ODX XSD does not indicate an upper limit for the number of V elements in COMPU-NUMERATOR, but it does indicate that the minimum number of V elements is 1 and not 2

### DIFF
--- a/odxtools/compumethods/readcompumethod.py
+++ b/odxtools/compumethods/readcompumethod.py
@@ -46,7 +46,8 @@ def _parse_compu_scale_to_linear_compu_method(scale_element,
     nums = coeffs.iterfind("COMPU-NUMERATOR/V")
 
     offset = computation_python_type(next(nums).text)
-    factor = computation_python_type(next(nums).text)
+    factor_el = next(nums, None)
+    factor = computation_python_type(factor_el.text if factor_el else '0')
     if coeffs.find("COMPU-DENOMINATOR/V") is not None:
         kwargs["denominator"] = float(
             coeffs.find("COMPU-DENOMINATOR/V").text)


andlaus:
interesting. If the number of values is different from two, what is the semantic of these values? I always thought the first is the numerator for the offset and the second is the numerator of the scaling factor...

odxtools/compumethods/readcompumethod.py
```diff
@@ -46,7 +46,8 @@ def _parse_compu_scale_to_linear_compu_method(scale_element,
     nums = coeffs.iterfind("COMPU-NUMERATOR/V")
 
     offset = computation_python_type(next(nums).text)
-    factor = computation_python_type(next(nums).text)
+    factor_el = next(nums, None)
+    factor = computation_python_type(factor_el.text if factor_el else '0')
```
andlaus, line 50: letting the scaling factor default to 0 does not really make sense because the physical value would not depend of the internal value anymore. did you mean 1?

kayoub5: I though so at first, but the formula from ODX specs suggests otherwise, basically it could be that there are more than one denominator, and denominator[1] is the one that depends on internal value.

image

I think that the correct implementation should be done with a list of numerator and a list of denominator, but it would change the current API too much.

andlaus: > I though so at first, but the formula from ODX specs suggests otherwise, [...]

I see: all compu methods are basically seen as special cases of (piecewise?) rational functions. If we ever implement this fully, providing the inverse functions is going be a lot of fun...

> I think that the correct (generic) implementation should be done with a list of numerator and a list of denominator, but it would change the current API too much.

agreed. Maybe we should put it on a list for odxtools 2.0 features...

andlaus:
ok, thanks for lifting my confusion. merging.

PR #54, opened 2022-08-17
Support TABLE-ROW-REF

kayoub5:
Not sure of how you want to handle TABLE-ROW-REF, this PR is a proposal of how to handle them

Note:
* This will convert TABLE-ROW-REF internally to a normal TABLE-ROW, once written back the ODX will contain TABLE-ROW-REF converted to TABLE-ROW

### DIFF
--- a/odxtools/table.py
+++ b/odxtools/table.py
@@ -2,8 +2,8 @@
 # Copyright (c) 2022 MBition GmbH
 
 import abc
-from dataclasses import dataclass, astuple
-from typing import Optional, List, Dict, Any
+from dataclasses import dataclass
+from typing import Optional, List, Dict, Any, Tuple
 
 from odxtools.utils import read_description_from_odx
 
@@ -55,21 +55,6 @@ class TableRow:
             + ")"
         )
 
-
-@dataclass
-class TableRowRef(TableRow):
-    """This class represents a TABLE-ROW-REF."""
-
-    def __init__(self, id: str):
-        self._id_ref = id
-        super().__init__("", "", "", 0, "")
-
-    def _resolve_references(self, id_lookup: Dict[str, Any]) -> None:
-        row: TableRow = id_lookup.get(self._id_ref)
-        super().__init__(*astuple(row))
-        super()._resolve_references(id_lookup)
-
-
 class Table(TableBase):
     """This class represents a TABLE."""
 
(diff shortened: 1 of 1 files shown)


odxtools/table.py
```diff
@@ -143,9 +157,13 @@ def read_table_from_odx(et_element):
...
     ]
 
+    table_row_refs = [
+        TableRowRef(el.get('ID-REF')) for el in et_element.iterfind("TABLE-ROW-REF")
+    ]
+
     table = Table(
         key_dop_ref=key_dop_ref,
-        table_rows=table_rows,
+        table_rows=table_rows + table_row_refs,
```
andlaus, line 166: I think it would be better to add a `table_row_refs` attribute to the table class, rename the current `table_rows` attribute to `_local_table_rows` (or similar) and introduce a new `table_rows` attribute which returns the union of the resolved table row references and the rows directly specified for the table. this is analogous to how inheritance is handled, cf. e.g. `ecu._local_services` vs `ecu.services`.

kayoub5:
@andlaus any idea how to make mypy happy?

odxtools/table.py
```diff
@@ -82,6 +84,11 @@ def key_dop(self) -> Optional[DopBase]:
         """The key data object property associated with this table."""
         return self._key_dop
 
+    @property
```
andlaus, line 87: This fixes mypy:

```python
    @property
    def table_rows(self) -> Iterable[TableRow]:
        """The table rows (both local and referenced) in this table."""
        return self._local_table_rows + self._ref_table_rows
```

odxtools/table.py
```diff
@@ -91,9 +98,14 @@ def _resolve_references(self, id_lookup: Dict[str, Any]) -> None:
         self._key_dop = id_lookup.get(self.key_dop_ref)
         if self._key_dop is None and self.key_dop_ref is not None:
             logger.warning(f"KEY-DOP-REF '{self.key_dop_ref!r}' could not be resolved.")
-        for table_row in self.table_rows:
+
+        for table_row in self._local_table_rows:
             table_row._resolve_references(id_lookup)
 
+        self._ref_table_rows = [
```
andlaus, line 105: This fixes mypy:

```python
        self._ref_table_rows = []
        for ref in self._table_row_refs:
            tr = id_lookup.get(ref)
            if isinstance(tr, TableRow):
                self._ref_table_rows.append(tr)
```

andlaus:
all good now. thanks for the contribution!

PR #57, opened 2022-08-18
Don't fail parsing if one ODX file have no DIAG-LAYER-CONTAINER

kayoub5:
According to the XSD, DIAG-LAYER-CONTAINER is one choice of what the ODX could contain

The ODX could contain FLASH, ECU-CONFIG, VEHICLE-INFO-SPEC...

If an ODX file in the PDX does not have a DIAG-LAYER-CONTAINER the parser will now ignore it instead of failing the whole PDX.

### DIFF
--- a/odxtools/database.py
+++ b/odxtools/database.py
@@ -39,12 +39,13 @@ class Database:
                     logger.info(f"Processing the file {zip_member}")
                     d = pdx_zip.read(zip_member)
                     root = ElementTree.fromstring(d)
-
-                    dlc_elements.append(root.find("DIAG-LAYER-CONTAINER"))
+                    dlc = root.find("DIAG-LAYER-CONTAINER")
+                    if dlc is not None:
+                        dlc_elements.append(dlc)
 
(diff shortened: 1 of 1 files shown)


odxtools/database.py
```diff
@@ -44,7 +44,7 @@ def __init__(self,
 
             tmp = [
                 read_diag_layer_container_from_odx(dlc_el, enable_candela_workarounds=enable_candela_workarounds) \
-                  for dlc_el in dlc_elements
+                  for dlc_el in filter(None, dlc_elements)
```
andlaus, line 47: to avoid confusion, please take care of this using an `if` on line 43, i.e.,

```python
dlc = root.find("DIAG-LAYER-CONTAINER")
if dlc is not None:
  dlc_elements.append(dlc)
```

andlaus: (note for a future version: add support for non-DLC files...)

andlaus:
good stuff. merging!

PR #59, opened 2022-08-19
Allow CodedConstParameter to have a bytes coded_value

kayoub5:
I have few OEM PDX files where the CODED-CONST is defined as a A_BYTEFIELD, not sure if other types like STRING should be considered as well.


andlaus:
```patch
--- a/odxtools/parameter_info.py
+++ b/odxtools/parameter_info.py
@@ -23,7 +23,7 @@ def parameter_info(param_list : Iterable[Union[Parameter, EndOfPduField]]) -> st
     result = ""
     for param in param_list:
         if isinstance(param, CodedConstParameter):
-            result += f"{param.short_name} : const = {param.coded_value}\n"
+            result += f"{param.short_name} : const = {str(param.coded_value)}\n"
             continue
         elif isinstance(param, MatchingRequestParameter):
             result += f"{param.short_name} : <matches request>\n"
```

should make mypy happy...

andlaus:
probably better solution:
```patch
diff --git a/odxtools/parameter_info.py b/odxtools/parameter_info.py
index c6b3383..78b48c2 100644
--- a/odxtools/parameter_info.py
+++ b/odxtools/parameter_info.py
@@ -23,7 +23,7 @@ def parameter_info(param_list : Iterable[Union[Parameter, EndOfPduField]]) -> st
     result = ""
     for param in param_list:
         if isinstance(param, CodedConstParameter):
-            result += f"{param.short_name} : const = {param.coded_value}\n"
+            result += f"{param.short_name} : const = {param._coded_value_str}\n"
             continue
         elif isinstance(param, MatchingRequestParameter):
             result += f"{param.short_name} : <matches request>\n"
```

andlaus:
all good now.

PR #50, opened 2022-07-29
Support parsing ENV-DATA-DESC / ENV-DATA / MUX

smileprem:

PR #60, opened 2022-08-20
Use findtext function

smileprem:


smileprem:
Please review after the MR https://github.com/mercedes-benz/odxtools/pull/50 is merged.

andlaus:
thanks for the PR. I've just merged #50. With that out of the way, I think the the 30-odd "old" patches are getting a bit unwieldy. can you remove them? AFAICS, this basically comes down to
```bash
git fetch origin
git rebase -i origin/main
# delete all lines except the one concerning the last commit
```

smileprem:
Hi @andlaus - I fixed the merge conflicts. Please review

odxtools/multiplexer.py
```diff
@@ -149,16 +149,8 @@ def __repr__(self) -> str:
...
-        int(et_element.find("BYTE-POSITION").text)
-        if et_element.find("BYTE-POSITION") is not None
-        else None
-    )
-    bit_position = (
-        int(et_element.find("BIT-POSITION").text)
-        if et_element.find("BIT-POSITION") is not None
-        else 0
-    )
+    byte_position = int(et_element.findtext("BYTE-POSITION"))
```
kayoub5, line 152: byte position is optional, int(None) will fail

odxtools/multiplexer.py
```diff
@@ -149,16 +149,8 @@ def __repr__(self) -> str:
...
+    bit_position = int(et_element.findtext("BIT-POSITION", 0))
```
kayoub5, line 153: bit position is optional, int(None) will fail

odxtools/multiplexer.py
```diff
@@ -208,15 +200,11 @@ def read_case_from_odx(et_element):
...
+    short_name = et_element.findtext("SHORT-NAME")
     long_name = et_element.find("LONG-NAME")
     if long_name is not None:
         long_name = long_name.text
-    byte_position = (
-        int(et_element.find("BYTE-POSITION").text)
-        if et_element.find("BYTE-POSITION") is not None
-        else None
-    )
+    byte_position = int(et_element.findtext("BYTE-POSITION"))
```
kayoub5, line 207: same here

odxtools/parameters/readparameter.py
```diff
@@ -28,10 +28,8 @@ def read_parameter_from_odx(et_element):
 
     semantic = et_element.get("SEMANTIC")
 
-    byte_position = int(et_element.find(
-        "BYTE-POSITION").text) if et_element.find("BYTE-POSITION") is not None else None
-    bit_position = int(et_element.find(
-        "BIT-POSITION").text) if et_element.find("BIT-POSITION") is not None else 0
+    byte_position = int(et_element.findtext("BYTE-POSITION"))
```
kayoub5, line 31: same here

odxtools/multiplexer.py
andlaus: while new solution will work, it is IMO nicer if a function call always returns objects of the same type, i.e.,
```suggestion
    bit_position = int(et_element.findtext("BIT-POSITION", "0"))
```

a slightly more severe issue with this: if the input file does not specify a bit position, it is impossible to detect that fact anymore, i.e., all output files written must either skip the tag for bit_position == 0 or the tag must always be written.

andlaus: adding a `findint(et_element, location, default=None)` utility function analogous to `findtext()` would be a workable solution. I'm just not sure if that is a good idea, though...

andlaus (changes requested):
in principle this is good, and I appreciate the PR. That said, the PR is probably a bit over-eager as constructs like `foo = int(et_elem.findtext("BAR", "0"))` will slightly change the contents of files written by odxtools.

also, please get rid of the  commits containing already merged functionality (i.e., all but the last two as far as I can see)

smileprem:
Hi @andlaus @kayoub5 - Thanks for the suggestion. This MR is supposed to be a simple search and replace of find() with findtext(). Along the way, I made minor tweaks, which of course would produce un-intended consequences.

Shall we please scope this MR to search and replace only?. We can implement the other improvements in subsequent MRs. I'll fix all the MR comments today. Please let me know, if this is OK?.

odxtools/multiplexer.py
smileprem: fixed

odxtools/multiplexer.py
smileprem: reverted the code

odxtools/multiplexer.py
smileprem: fixed

odxtools/parameters/readparameter.py
smileprem: fixed

smileprem:
I'm used to git merge. I think, I messed up the git rebase. Let me close this and open another MR. Sorry about the trouble.

PR #64, opened 2022-08-22
Warn instead of assert when for unkown parent ref type

kayoub5:
* ref_type is never actually used by the code, making an assert is an overkill
* Using warnings module ensure that we still see the error, but we can choose to ignore it or suppress it
* This servers as reference idea for future cases where checks needs to be made but can be tolerated in some edge cases


andlaus:
thanks

PR #65, opened 2022-08-22
Support for int written as float

kayoub5:
It may sound strange, but I got PDX files from a big OEM, where values are integer but written as float, example `15.0` instead of `15`

### DIFF
--- a/odxtools/odxtypes.py
+++ b/odxtools/odxtypes.py
@@ -16,7 +16,9 @@ def parse_int(value: str) -> int:
     try:
         return int(value)
     except ValueError:
-        return int(float(value))
+        v = float(value)
+        assert v.is_integer()
+        return int(v)
 
 _ODX_TYPE_PARSER: Dict[str, Callable[[str], PythonType]] = {
     "A_INT32": parse_int,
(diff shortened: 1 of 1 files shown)


odxtools/odxtypes.py
```diff
@@ -12,9 +12,15 @@ def bytefield_to_bytearray(bytefield: str) -> bytearray:
 PythonType = Union[str, int, float, bytearray]
 LiteralPythonType = Type[Union[str, int, float, bytearray]]
 
+def parse_int(value: str) -> int:
+    try:
+        return int(value)
+    except ValueError:
+        return int(float(value))
```
andlaus, line 19: thanks. A while ago, I had this issue in a slightly different context: ARXML files written by PREEVISION sometimes do this as well. I would recommend checking that the floating point number ends in .0, though:
```patch
diff --git a/odxtools/odxtypes.py b/odxtools/odxtypes.py
index 0506b94..c192c6e 100644
--- a/odxtools/odxtypes.py
+++ b/odxtools/odxtypes.py
@@ -16,7 +16,9 @@ def parse_int(value: str) -> int:
     try:
         return int(value)
     except ValueError:
-        return int(float(value))
+        v = float(value)
+        assert v == int(v)
+        return int(v)
 
 _ODX_TYPE_PARSER: Dict[str, Callable[[str], PythonType]] = {
     "A_INT32": parse_int,
```

(use a warning instead of the assertation if you prefer.)

kayoub5:
@andlaus review?

andlaus:
perfect. thanks!

PR #66, opened 2022-08-22
Support for encoding and decoding Multiplexer. Resolves #61

### LINKED ISSUE #61: Multiplexer does not inherit from DopBase
This causes any encode/decode to break with an unexpected kind of errors for example `AttributeError: 'Multiplexer' object has no attribute 'bit_length'`

Call stack
```
..\odxtools\odxtools\structures.py:33: in bit_length
    elif all(p.bit_length is not None for p in self.parameters):
..\odxtools\odxtools\structures.py:33: in <genexpr>
    elif all(p.bit_length is not None for p in self.parameters):
..\odxtools\odxtools\parameters\parameterwithdop.py:70: in bit_length
    return self.dop.bit_length
..\odxtools\odxtools\structures.py:33: in bit_length
    elif all(p.bit_length is not None for p in self.parameters):
..\odxtools\odxtools\structures.py:33: in <genexpr>
    elif all(p.bit_length is not None for p in self.parameters):

    @property
    def bit_length(self):
        if self.dop is not None:
>           return self.dop.bit_length
E           AttributeError: 'Multiplexer' object has no attribute 'bit_length'
```

kayoub5:


kayoub5:
@andlaus review?

andlaus:
looks good, thanks!

PR #67, opened 2022-08-23
Better check for ODX file extension

kayoub5:


andlaus:
thanks

andlaus:
(this now works because of #57, right @kayoub5 ?)

kayoub5:
> (this now works because of #57, right @kayoub5 ?)

Yes

PR #68, opened 2022-08-23
Better overlap and gap handling

kayoub5:
Handle the case were parameters overlap and have gap between them a bit better

Currently the encoding logic do encode them correctly but still bails if it encounter such parameters due to bit_length checking

One simple example is a struct defining a DTC code, it can be treated as a whole number, or have its individual components each treated as a parameter, for convenience, user could define both in the same struct.

image

### DIFF
--- a/odxtools/structures.py
+++ b/odxtools/structures.py
@@ -140,8 +140,7 @@ class BasicStructure(DopBase):
             bit_length: Optional[int]):
 
         if self._byte_size is not None:
-            # We definetly broke something if we didn't respect the explicit
-            # byte_size
+            # We definitely broke something if we didn't respect the explicit byte_size
             assert len(coded_rpc) == self._byte_size, self._get_encode_error_str('was', coded_rpc, self._byte_size * 8)
             # forget what we calcualted, ODX defined an explicit byte_size
             # if bit_length was less than _byte_size then we have padding or gaps between parameters
--- a/odxtools/structures.py
+++ b/odxtools/structures.py
@@ -1,17 +1,20 @@
 # SPDX-License-Identifier: MIT
 # Copyright (c) 2022 MBition GmbH
 
+import math
 from typing import Any, List, Dict, Iterable, Optional, OrderedDict, Union
+import warnings
 
 from .dataobjectproperty import DataObjectProperty, DopBase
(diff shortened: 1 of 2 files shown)


kayoub5:
@andlaus review?

odxtools/structures.py
```diff
@@ -122,14 +125,60 @@ def convert_physical_to_internal(self,
...
+
+    def _get_encode_error_str(self,
+            verb: str,
+            coded_rpc: bytearray,
+            bit_length: int):
+
+        return str(f"Structure {self.short_name} {verb} encoded uncorrectly:" +
+                f" actual length is {len(coded_rpc)}," +
+                f" computed byte length is {bit_length // 8}," +
+                f" computed_rpc is {coded_rpc.hex()}" +
```
kayoub5, line 178:
```suggestion
                f" computed_rpc is {coded_rpc.hex()}\n" +
```

odxtools/structures.py
```diff
@@ -122,14 +125,60 @@ def convert_physical_to_internal(self,
...
 
         return bytearray(coded_rpc)
 
+    def _validate_coded_rpc(
+            self,
+            coded_rpc: bytearray,
+            bit_length: Optional[int]):
+
+        if self._byte_size is not None:
+            # We definetly broke something if we didn't respect the explicit
```
andlaus, line 143:
```suggestion
            # We definitely broke something if we didn't respect the explicit
```

andlaus (commented):
looks good. If possible, I would like to see the somersault example to be extended by such a case, though...

andlaus:
> I have limited time at the moment for writing detailed tests, I could try to add those afterwards (different PR)

ok, let's get this merged. be aware though, that not having this in the example PDX, fixing issues like #70 is difficult for people who don't have datasets that use this feature.

> Tip: you may want to add Codecov https://about.codecov.io/ to the CI, it would help with telling what changes requires tests and what changes are already covered with existing tests.

Yep, this has been on my TODO list for quite a while, but more pressing matters kept interfering and even my time is limited ;)

maybe soon.

kayoub5:
@andlaus could you publish a version with the PRs merged so far?

andlaus:
np. done.

Issue #70, opened 2022-08-23
__message_format_lines crash with overlapping signals


kayoub5:
I am not sure how that function works to fix the issue myself, but I could provide what the struct that caused the issue look like

All have byte position = 0
| Param  | bit position | bit length |
|---|---|---|
| 1 | 0 | 2 |
| 2 | 2 | 2 |
| 3 | 3 | 2 |
| 4 | 5 | 1 |
| 5 | 6 | 1 |
| 6 | 7 | 1 |

Stack trace
```
            next_line = f"{(len(indent_for_byte_numbering) - 1 - len(str(byte))) * ' '}{byte} "

            for bit in range(8):
                assert 8 * byte + bit <= breakpoint

                if 8 * byte + bit == breakpoint:
                    # END-OF-PDU fields do not exhibit a fixed bit
                    # length, so they need special treatment here
                    dct = None
>                   if hasattr(params[i], 'dop'):
E                   IndexError: list index out of range
```

andlaus:
I suppose that in order to help you here I would need a dataset which uses this feature (e.g. an extension of the somersault example. (hint, hint ;)))

kayoub5:
@andlaus do you have enough now to make a fix?

PR #71, opened 2022-08-26
Fix BasicStructure bit_length

kayoub5:
Include test case for #70, but not a fix since I still can't get my head around how that function works


andlaus:
thanks for the catch. (and also for the unit tests.)

PR #72, opened 2022-08-29
Centralized struct bit length calculation

kayoub5:
* BasicStructure.bit_length do not take into consideration EndOfPduField
* BasicStructure.convert_physical_to_internal recalculates bit_length without taking into consideration overlapped parameters and gaps

Centralize the two implementations into one that takes into consideration the special cases of both


kayoub5:
@andlaus review?

andlaus:
looks good, thanks for the catch. merging...

PR #73, opened 2022-09-07
Use bytearray for scale limit

kayoub5:
Comparison for bytearray is well defined in Python
```python
a = bytearray.fromhex('aa')
b = bytearray.fromhex('bb')
print(a < b) # True
print(a > b) # False
```

Using `int` breaks the assert in [odxtools\compumethods\texttablecompumethod.py:](https://github.com/mercedes-benz/odxtools/blob/b56eb61b54fbfca4fa38247c8d5e188344a82efc/odxtools/compumethods/texttablecompumethod.py#L27)


andlaus:
thanks

PR #74, opened 2022-09-07
Support for DOCREF

kayoub5:


andlaus:
in principle, this looks good but I think it is a bit incomplete: There's no unit test which uses DOCREF and (more importantly), this feature does not seem to be supported by the writing code. It would be great to keep the property that all PDX files which can be read by odxtools can be written back to disk without a loss of information (modulo bugs). the writing code is located here:

https://github.com/mercedes-benz/odxtools/blob/main/odxtools/write_pdx_file.py
where the more important jinja2 templates being in the (pdx_stub/macros)[https://github.com/mercedes-benz/odxtools/tree/main/odxtools/pdx_stub/macros) subfolder.

kayoub5:
> in principle, this looks good but I think it is a bit incomplete: There's no unit test which uses DOCREF and (more importantly), this feature does not seem to be supported by the writing code. It would be great to keep the property that all PDX files which can be read by odxtools can be written back to disk without a loss of information (modulo bugs). the writing code is located here:
> 
> https://github.com/mercedes-benz/odxtools/blob/main/odxtools/write_pdx_file.py with the more important jinja2 templates being in the [pdx_stub/macros](https://github.com/mercedes-benz/odxtools/tree/main/odxtools/pdx_stub/macros) subfolder.

I am aware of that, but the current logic for DOCREF have no side effects on documents that do not use DOCREF.
(If the document was using DOCREF then it was written back incorrectly even without the changes in this PR)

Adding writing support to this already big PR would make way too massive.
I think it would be best to split writing DOCREF to its own PR.

kayoub5:
For test I would recommend using an existing (known to be correct) PDX file
For example
https://github.com/sebastianwilczek/CANoe-Configurations/blob/master/Ethernet/Diagnostics/EthernetDiagnostics/ODX/UDS-ExampleEcu-8.0.1.pdx

kayoub5:
@andlaus should I move the `make_ref` changes to its own PR, so that this one is smaller and have more room for adding writing updates?

andlaus:
> @andlaus should I move the make_ref changes to its own PR, so that this one is smaller and have more room for adding writing updates?

no it's okay as is IMO.

I read this up a bit: the DOCREF stuff is not really relevant w.r.t. writing for us because we only generate a single odx-d file anyway. The only thing which should be fixed is to avoid collisions between IDs stemming from different odx-d files, but that's quite a niche problem IMO. let's merge this..

PR #76, opened 2022-09-13
Support compu scale with int written as float

kayoub5:
Same as #65 but for compu scale

Note: a PDX that reproduce the issue could be found in https://github.com/speedbyte/code_composer_build/tree/a34f54cdee679d0b1104b527aafde5f17054634c/doc/Diagnose/PDX


andlaus:
thanks. merging (sorry for the delay, I was AFK for the last two weeks)

PR #77, opened 2022-09-20
Implicit LENGTH-KEY calculation

kayoub5:
Auto calculate the value of LENGTH-KEY when possible, still honer explicit values for the parameter


andlaus:
Nice! Thanks!

PR #78, opened 2022-09-23
Fix various issues

andlaus:
This fixes a bunch of smaller issues which crept in recently.

the morale is that we definitely better unit tests soon...

### DIFF
--- a/odxtools/cli/browse.py
+++ b/odxtools/cli/browse.py
@@ -17,7 +17,7 @@ import logging
 # logging.basicConfig(level=logging.DEBUG)
 
 # name of the tool
-name = "browse"
+_odxtools_tool_name_ = "browse"
 
 def _convert_string_to_odx_type(string_value: str, odx_type: DataType):
     """Similar to odx_type.from_string(string_value) but more relaxed to parse user input"""
--- a/odxtools/cli/browse.py
+++ b/odxtools/cli/browse.py
@@ -16,6 +16,8 @@ from . import _parser_utils
 import logging
 # logging.basicConfig(level=logging.DEBUG)
 
+# name of the tool
+name = "browse"
 
 def _convert_string_to_odx_type(string_value: str, odx_type: DataType):
     """Similar to odx_type.from_string(string_value) but more relaxed to parse user input"""
--- a/odxtools/cli/main.py
+++ b/odxtools/cli/main.py
@@ -2,13 +2,21 @@
 # Copyright (c) 2021-2022 MBition GmbH
 
 import argparse
-
-from . import list
-from . import browse
-from . import snoop
-from . import find
+import importlib
 
 from ..version import __version__ as odxtools_version
+from .dummy_sub_parser import DummyTool
+
+# import the tool modules which can be loaded. if a tool
+# can't be loaded, add a dummy one
+tool_modules = []
+for tool_name in ["list", "browse", "snoop", "find"]:
+    try:
+        tool_modules.append(
+            importlib.import_module(f".{tool_name}", package="odxtools.cli"))
+    except Exception as e:
+        tool_modules.append(
+            DummyTool(tool_name, e))
 
 def start_cli():
     argparser = argparse.ArgumentParser(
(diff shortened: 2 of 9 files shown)


andlaus:
@kayoub5: can you review and merge this? (I've added you as a maintainer to the repository...)

odxtools/cli/main.py
```diff
@@ -47,11 +53,6 @@ def start_cli():
...
-    if args.subparser_name == "list":
-        list.run(args)
-    elif args.subparser_name == "browse":
-        browse.run(args)
-    elif args.subparser_name == "snoop":
-        snoop.run(args)
-    elif args.subparser_name == "find":
-        find.run(args)
+    for tool in tool_modules:
+        if tool.name == args.subparser_name:
```
kayoub5, line 57: Just a suggestion, maybe use `tool.__name__ ` instead ?

odxtools/database.py
```diff
@@ -84,7 +84,7 @@ def get_lookups(layer: Union[DiagLayerContainer, DiagLayer], docnames = None) ->
             for key, value in dict(lookup).items():
                 # In case an ID-REF uses a DOCREF
                 for docname in docnames:
-                    result[docname + ">" + key] = value
+                    result[f"{key}_from_{docname}"] = value
```
kayoub5, line 87: Line 100 needs to be updated as well

```
docref_lookup = {k: v for k, v in self.id_lookup.items() if '>' in k}
```

odxtools/cli/main.py
```diff
@@ -2,13 +2,21 @@
...
+from .dummy_sub_parser import DummyTool
+
+# import the tool modules which can be loaded. if a tool
+# can't be loaded, add a dummy one
+tool_modules = []
+for tool_name in ["list", "browse", "snoop", "find"]:
+    try:
+        tool_modules.append(
+            importlib.import_module(f".{tool_name}", package="odxtools.cli"))
+    except Exception as e:
```
kayoub5, line 17:
```suggestion
    except ImportError as e:
```

kayoub5 (changes requested):

odxtools/cli/main.py
andlaus: I'm not so sure about this: if the exception thrown is not an ImportError, it should be ignored as well, right? that said, I'm not sure if this can happen with `importlib.import_module`. any ideas?

andlaus:
thanks for the review. I fixed the first two issues. For the third suggestion, I'm not too sure whether it is a good idea or not. (opinions welcome!)

odxtools/database.py
andlaus: ok. I suppose we should change the DOCREF implementation a bit in the future: I propose a hashable `OdxReference` class that has a total ordering which includes the IDREF and the optional DOCREF instead of doing references using strings. not now, though...

odxtools/cli/browse.py
```diff
@@ -17,7 +17,7 @@
 # logging.basicConfig(level=logging.DEBUG)
 
 # name of the tool
-name = "browse"
+__name__ = "browse"
```
kayoub5, line 20: __name__ is an auto generated variable by python itself, it includes the module name

odxtools/database.py
kayoub5: A named tuple would do the job

odxtools/cli/main.py
```diff
@@ -47,11 +53,6 @@ def start_cli():
...
-    if args.subparser_name == "list":
-        list.run(args)
-    elif args.subparser_name == "browse":
-        browse.run(args)
-    elif args.subparser_name == "snoop":
-        snoop.run(args)
-    elif args.subparser_name == "find":
-        find.run(args)
+    for tool in tool_modules:
+        if tool.__name__ == args.subparser_name:
```
kayoub5, line 57:
```suggestion
        if tool.__name__.endswith(args.subparser_name):
```

odxtools/cli/browse.py
andlaus: sure, but it can be overwritten:
```python
>>> import odxtools.database
>>> odxtools.database.__name__
'odxtools.database'
>>> odxtools.database.__name__ = "bla"
>>> odxtools.database.__name__
'bla'
```

maybe we should rename this to something like `_odxtools_tool_name`...

odxtools/cli/main.py
andlaus: I guess I want this to be a bit more bulletproof, IOW, I don't want to rely on this not breaking in the future.

andlaus:
update pushed...

kayoub5 (approved):

PR #80, opened 2022-10-03
Fix odx tools lib update errors

LLOUIS03:
This contain the changes related of the odxprocessor

### DIFF
--- a/odxtools/communicationparameter.py
+++ b/odxtools/communicationparameter.py
@@ -47,7 +47,7 @@ def _read_complex_value_from_odx(et_element):
 
 
 def read_communication_param_ref_from_odx(et_element):
-    id_ref = make_ref(et_element)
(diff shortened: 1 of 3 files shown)


kayoub5:
@LLOUIS03 could you elaborate on what error you are trying to fix?

odxtools/communicationparameter.py
```diff
@@ -47,7 +47,7 @@ def _read_complex_value_from_odx(et_element):
 
 
 def read_communication_param_ref_from_odx(et_element):
-    id_ref = make_ref(et_element)
+    id_ref = et_element.get("ID-REF")
```
kayoub5, line 50: Please double check in your pdx if the element have a `DOCREF` attribute, if it does, then the ID-REF is not pointing to the same file.

PR #82, opened 2022-10-14
Refactor the ODXLINK reference infrastructure

andlaus:
instead of being a simple string, an ODXLINK ID of an object now consists of the object's document-local ID and the corresponding document fragment. References are the same as IDs, but if no document is specified, it means "same document as the referencing object".

While at it, we fix mypy for the tests, examples and setup.py and as a consequence, run it for the whole project instead of just the `odxtools` sub-directory.

Besides this, this PR also fixes a few random smaller issues which I stumbled over.


andlaus:
@kayoub5: It would be good if you ensured that your files continue to work before merging this (I don't have any files which use the DOCREF feature seriously)

odxtools/audience.py
```diff
@@ -24,10 +27,10 @@ def enabled_audiences(self):
     def disabled_audiences(self):
         return self._disabled_audiences
 
-    def _resolve_references(self, id_lookup):
-        self._enabled_audiences = [id_lookup[ref]
+    def _resolve_references(self, odxlinks):
```
kayoub5, line 30: Some type hinting would be nice
```suggestion
    def _resolve_references(self, odxlinks: OdxLinkDatabase):
```

odxtools/communicationparameter.py
```diff
@@ -33,7 +34,7 @@ def __str__(self) -> str:
         return f"CommunicationParameter('{self.id_ref}', value={val})"
 
     def _python_name(self):
-        return self.id_ref.replace(".", "__")
+        return self.id_ref.ref_id.replace(".", "__")
```
kayoub5, line 37: I found some PDX files that use minus sign in the id string
```suggestion
        return self.id_ref.ref_id.replace(".", "__").replace("-", "_")
```

odxtools/diaglayer.py
```diff
@@ -542,71 +545,78 @@ def read_parent_ref_from_odx(et_element):
...
 
     if et_element.find("DIAG-DATA-DICTIONARY-SPEC"):
         diag_data_dictionary_spec = read_diag_data_dictionary_spec_from_odx(
-            et_element.find("DIAG-DATA-DICTIONARY-SPEC"))
+            et_element.find("DIAG-DATA-DICTIONARY-SPEC"), doc_frag)
     else:
         diag_data_dictionary_spec = None
 
-    import_refs = [make_ref(ref) for ref in et_element.iterfind("IMPORT-REFS/IMPORT-REF")]
+    import_refs = [OdxLinkRef.from_et(ref, doc_frag, doc_frag)
```
kayoub5, line 618:
```suggestion
    import_refs = [OdxLinkRef.from_et(ref, doc_frag)
```

odxtools/parameters/parameterwithdop.py
```diff
@@ -39,20 +40,20 @@ def dop(self) -> Union[DopBase, None]:
...
         else:
             self.dop_snref = dop.short_name
 
-    def resolve_references(self, parent_dl, id_lookup):
+    def resolve_references(self, parent_dl, odxlinks):
         dop = self.dop
         if self.dop_snref:
-            dop = parent_dl.data_object_properties[self.dop_snref]
-            if not dop:
+            dop = parent_dl.data_object_properties.get(self.dop_snref)
```
kayoub5, line 50: get ? data_object_properties is not a dict, it's NamedItemList and does not have a get function

kayoub5:
@andlaus logic seems to be broken, try with this PDX, https://github.com/speedbyte/code_composer_build/tree/a34f54cdee679d0b1104b527aafde5f17054634c/doc/Diagnose/PDX

odxtools/communicationparameter.py
andlaus: hm, ok. I think that fixing this properly will require full internalization of the communication parameters (@smileprem is currently working on this, if I am not mistaken...) I'd say we should keep this in mind, but defer fixing this to another PR?

andlaus:
> @andlaus logic seems to be broken, try with this PDX, https://github.com/speedbyte/code_composer_build/tree/a34f54cdee679d0b1104b527aafde5f17054634c/doc/Diagnose/PDX

Oops, correct. I re-read the spec and it is quite a bit more arcane than I originally understood it. Fixing this will take a few days though, stay tuned.

That said, I hopefully fixed the [other](https://github.com/mercedes-benz/odxtools/pull/82/commits/9561a3d710eb4a545838bf3fa9daac935bd56319) [isses](https://github.com/mercedes-benz/odxtools/pull/82/commits/76de64adbf2e7e1ede57d67bf8c2d03d283f79c2) you've discovered. (Thanks for the review, BTW!)

odxtools/communicationparameter.py
kayoub5: Then revert `assert item_name.isidentifier()` in odxtools/nameditemlist.py until `_python_name` is covering all cases

smileprem: Hi @andlaus - Thanks for the trust on us. This is planned for Q4. I'll put up the MR in the next few weeks. But, in the meantime, please do what's needed to make the library usable. We could always replace the code. Just my opinion.

andlaus: > Then revert assert item_name.isidentifier() in odxtools/nameditemlist.py until _python_name is covering all cases

ah, I didn't make the connection here. That item names are identifiers is imperial for `NamedItemList` because the whole point of this class is to be able to access items via their name like e.g. `ecu.services.Default_Session_Start` instead of having to go through indices. I've amended the PR with your solution. (thanks for the catch.)

odxtools/multiplexer.py
```diff
@@ -318,34 +312,31 @@ def read_case_from_odx(et_element):
...
-    if long_name is not None:
-        long_name = long_name.text
-    byte_position = (
-        int(et_element.find("BYTE-POSITION").text)
-        if et_element.find("BYTE-POSITION") is not None
-        else None
-    )
-    switch_key = read_switch_key_from_odx(et_element.find("SWITCH-KEY"))
+    long_name = et_element.findtext("LONG-NAME")
+    byte_position = int(et_element.findtext("BYTE-POSITION", "0"))
```
kayoub5, line 322: 0 is not the same as None, are you sure?

odxtools/odxlink.py
```diff
@@ -0,0 +1,223 @@
...
+        self._db: Dict[OdxDocFragment, Dict[OdxLinkId, Any]] = {}
+
+    def resolve(self, ref: OdxLinkRef) -> Any:
+        """
+        Resolve a reference to an object
+
+        If the database does not contain any object which is referred to, a
+        KeyError exception is raised.
+        """
+        assert isinstance(ref, OdxLinkRef)
```
kayoub5, line 163: if res is None return None?

odxtools/odxlink.py
```diff
@@ -0,0 +1,223 @@
...
+        if local_id is None:
+            return None
+
+        return OdxLinkId(local_id, doc_fragments)
+
+from typing import Dict, Union, Optional, Any
+from dataclasses import dataclass, field, replace
+from xml.etree.ElementTree import Element
+
+@dataclass(frozen=True)
```
kayoub5, line 80: Maybe named tuple instead of frozen dataclass?

odxtools/parameters/parameterwithdop.py
```diff
@@ -39,20 +40,20 @@ def dop(self) -> Union[DopBase, None]:
...
+            if dop is None:
                 logger.info(
                     f"Param {self.short_name} could not resolve DOP-SNREF {self.dop_snref}")
         elif self.dop_ref:
-            dop = id_lookup.get(self.dop_ref)
-            if not dop:
+            dop = odxlinks.resolve_lenient(self.dop_ref) # TODO: non-lenient!
```
kayoub5, line 55: Only used here, may try/except instead of double implementation?

tests/test_singleecujob.py
```diff
@@ -102,45 +129,24 @@ class Context(NamedTuple):
...
-                )
-            ]
+            input_params=input_params,
+            output_params=output_params,
+            neg_output_params=neg_output_params,
         )
 
         self.singleecujob_odx = f"""
-            <SINGLE-ECU-JOB ID="{self.singleecujob_object.id}">
+            <SINGLE-ECU-JOB ID="{self.singleecujob_object.id.local_id}">
```
kayoub5, line 142: Maybe a helper that takes OdxLinkId as input and return all attributes?

odxtools/odxlink.py
andlaus: this is intentional: it makes implementation errors much more obvious (and if this happens, some calling code needs to be fixed...)

odxtools/parameters/parameterwithdop.py
andlaus: yes, the database is correct: it contains some static fields which are not yet internalized by odxtools but which are referenced. (I'll implement static fields soon.)

tests/test_singleecujob.py
andlaus: I guess that won't make things clearer: The problem is that for IDs, we don't want the document fragment stuff, since this can only be specified for references, but not for IDs.

odxtools/multiplexer.py
kayoub5: If ODX says BYTE-POSITION is mandatory here then we are good

smileprem:
> @kayoub5, @smileprem: did you test if this PR works with the data sets that are relevant for you? If it does, I'd like to merge it in the next few days...

Hi @andlaus - This is a large MR and I didn't get a chance to review it. I ran my codebase with your branch against all the PDX files that I have. I get the below error 

```
  File "odxprocessor/odx/parameter/parameter.py", line 49, in get_table_key
    table_from_param = database.id_lookup[req_param.table_ref]
AttributeError: 'Database' object has no attribute 'id_lookup'
make: *** [run_local] Error 1
```
when I try to do
```
table_from_param = database.id_lookup[req_param.table_ref]
```

I suppose this is introduced by this MR. Could you please update the documentation on how to refer the objects from the map?

andlaus:
can you try if

```python
table_from_param = database.id_lookup.resolve(req_param.table_ref)
```

does the trick for you? I also recommend to rename the `id_lookup` parameter to `odxlink_db` (or similar).

smileprem:
> can you check if
> 
> ```python
> table_from_param = database.odxlinks.resolve(req_param.table_ref)
> ```
> 
> does the trick for you?

```
table_from_param = database.odxlinks.resolve(parameter.table_key_ref)
```

This works in many other places where I refer the id_lookup. But not able to resolve the table needed for TABLE-STRUCT parameter, even though the parameter that I pass is the same in the ref_id in odxlinks.

```
param.id.local_id =  _New_ECU_1_945
parameter.table_key_ref =  OdxLinkRef(ref_id='_New_ECU_1_945', ref_docs=[OdxDocFragment(doc_name='New_ECU_1', doc_type='CONTAINER'), OdxDocFragment(doc_name='New_ECU_1_AVAS', doc_type='LAYER')])
{"message": "NOT IMPLEMENTED YET - Cant find table from table_key_ref", "parameter": "Parameter(short_name='RoutineControlOptionRecord', type='TABLE-STRUCT', semantic=None, byte_position=4)", "level": "ERROR"}
```

Need to spend some time to figure out what's going on. I'll keep you posted.

kayoub5:
> > @andlaus logic seems to be broken, try with this PDX, https://github.com/speedbyte/code_composer_build/tree/a34f54cdee679d0b1104b527aafde5f17054634c/doc/Diagnose/PDX
> 
> Oops, correct. I re-read the spec and it is quite a bit more arcane than I originally understood it. Fixing this will take a few days though, stay tuned.
> 
> That said, I hopefully fixed the [other](https://github.com/mercedes-benz/odxtools/pull/82/commits/9561a3d710eb4a545838bf3fa9daac935bd56319) [isses](https://github.com/mercedes-benz/odxtools/pull/82/commits/76de64adbf2e7e1ede57d67bf8c2d03d283f79c2) you've discovered. (Thanks for the review, BTW!)

The PDX in that link still does not load, Key Error in `OutputParam._resolve_references`

andlaus:
> > > @andlaus logic seems to be broken, try with this PDX, https://github.com/speedbyte/code_composer_build/tree/a34f54cdee679d0b1104b527aafde5f17054634c/doc/Diagnose/PDX
> > 
> > 
> > Oops, correct. I re-read the spec and it is quite a bit more arcane than I originally understood it. Fixing this will take a few days though, stay tuned.
> > That said, I hopefully fixed the [other](https://github.com/mercedes-benz/odxtools/pull/82/commits/9561a3d710eb4a545838bf3fa9daac935bd56319) [isses](https://github.com/mercedes-benz/odxtools/pull/82/commits/76de64adbf2e7e1ede57d67bf8c2d03d283f79c2) you've discovered. (Thanks for the review, BTW!)
> 
> The PDX in that link still does not load, Key Error in `OutputParam._resolve_references`

Darn! I suppose you don't have a shareable pdx file exhibiting this issue? If no, can you please provide the following information (hint: catch `KeyError` in `OdxLinkDatabase.resolve()` and set a breakpoint in the handler):

- what is the `ref_id` of the reference to be resolved?
- which document fragments are considered for resolving it? (the `ref_docs` attribute of the reference to be resolved)
- after unzipping the PDX file, is there an XML tag in some file that exhibits the same `ID=` attribute as mentioned by the `ref_id` of the reference to be resolved?
- what are the names and types of the container and the layer where this XML tag is located in? (These are the  SHORT-NAMEs and XML tag names of the objects of the second and third level below the root tag of the respective XML file. (assuming the reference to be resolved is for an object located in a layer.))

andlaus:
The current main branch does not work with the internal database files due to the current approach to the DOCREF mechanism. I would thus like to get this PR merged this week or at the latest early next week. @kayoub5, @smileprem: Can you please boost the priority of fixing it up for your stuff, so that we do not get too severe regressions?

odxtools/nameditemlist.py
```diff
@@ -40,6 +40,7 @@ def append(self, item: T):
         self._list.append(item)
 
         item_name = self._item_to_name_fn(item)
+        assert item_name.isidentifier()
```
kayoub5, line 43: Please revert this for now, or change it to log a warning

I am getting way too may errors reported by it.

odxtools/database.py
```diff
@@ -23,7 +24,7 @@ def __init__(self,
                  odx_d_file_name: str = None,
                  enable_candela_workarounds: bool = True):
 
-        self._id_lookup: Dict[str, Any] = {}
+        self._odxlinks: Dict[str, Any] = {}
```
kayoub5, line 27: Wrong key type

odxtools/odxlink.py
```diff
@@ -0,0 +1,223 @@
...
+        """
+
+        # put all new objects into the databases for the document
+        # fragments which it specifies
+        for odx_link_id, obj in new_entries.items():
+            for doc_frag in odx_link_id.doc_fragments:
+                if doc_frag not in self._db:
+                    self._db[doc_frag] = dict()
+
+                doc_frag_db = self._db[doc_frag][odx_link_id] = obj
```
kayoub5, line 223: doc_frag_db not used

kayoub5:
Please make sure to bump the major version once this is released, a lot of breaking changes in it

odxtools/nameditemlist.py
andlaus: ok, done. do happen to know where this still happens to you? (with the dot-to-underscore change, the comparam stuff should be fine now?)

odxtools/database.py
andlaus: yep, but also superfluous local variable at this place. removed. thanks for the catch...

odxtools/odxlink.py
andlaus: oops, right. Thanks for the catch!

odxtools/nameditemlist.py
kayoub5: I have item names that start with a number...

odxtools/nameditemlist.py
```diff
@@ -40,7 +40,14 @@ def append(self, item: T):
...
 
         item_name = self._item_to_name_fn(item)
-        assert item_name.isidentifier()
+
+        if not item_name.isidentifier():
+            warnings.warn(f"For NamedItemList objects to work properly, all "
+                          f"item names must be valid python identiers."
+                          f"(Encountered name {item_name} which is not an "
+                          "identifier!")
+            return
```
kayoub5, line 49: Remove the return, user can still use `getattr(obj, "44&weirdname")` as a workaround.

andlaus:
(for me, the linked PDX works fine:

```
odxtools list -a ~/tmp/TelephoneUDS_UHV_NAR.pdx
ECU-SHARED-DATA 'BL_LIBEPHVA' (Receive ID: None, Send ID: None)
 num services: 0, num DOPs: 391, num communication parameters: 0.
 Description: An ECU variant library that defines all RDIDs for measurement values, adjustment channels, basic settings, DTCs and io control actuator tests. The lib is based on the former BV BV\_ParkiAssis2UDS\_002035.
The DOPs of the ECU-SHARED-DATA 'BL_LIBEPHVA' are: 
  DataObjectProperty('DOP_ASCII11Bytes', category='IDENTICAL', diag_coded_type='StandardLengthType(base_data_type='DataType.A_ASCIISTRING', bit_length=88)', physical_type=PhysicalType(base_data_type=<DataType.A_UNICODE2STRING: 'A_UNICODE2STRING'>, display_radix=None, precision=None), unit_ref='None')
[...]
```

odxtools/nameditemlist.py
andlaus: good point. `obj.get("44&weirdname")` and `obj["44&weirdname"]` should work as well...

kayoub5:
> > The PDX in that link still does not load, Key Error in OutputParam._resolve_references
> 
> @kayoub5 is this issue still around? I think that this is the current showstopper..

Works in latest commit, took a while to run the tests

odxtools/nameditemlist.py
```diff
@@ -40,7 +40,13 @@ def append(self, item: T):
         self._list.append(item)
 
         item_name = self._item_to_name_fn(item)
-        assert item_name.isidentifier()
+
+        if not item_name.isidentifier():
+            warnings.warn(f"For NamedItemList objects to work properly, all "
```
kayoub5, line 45: Missing `import warnings`

kayoub5: Add `OdxWarning` as argument, so users can filter on the warning

andlaus:
> Works in latest commit, took a while to run the tests

cool! It might be that loading the database became a bit slower because making and resolving these ODXLINK references becomes quite a bit more complicated. I think that correctness while loading the database is more important than performance as long as it only takes a few seconds...

Unless @smileprem vetoes, I'll merge this tomorrow or on Wednesday...

odxtools/multiplexer.py
```diff
@@ -20,15 +20,15 @@ class MultiplexerCase:
...
 
     def __post_init__(self):
         self._structure: Optional[BasicStructure] = None
 
-    def _resolve_references(self, id_lookup: Dict[str, Any]) -> None:
-        self._structure = id_lookup.get(self.structure_ref)
+    def _resolve_references(self, odxlinks: OdxLinkDatabase) -> None:
+        self._structure = odxlinks.resolve(self.structure_ref)
         if self._structure is None:
             logger.warning(
```
kayoub5, line 33: This warning will not longer trigger (dead code)

Maybe add `if self.structure_ref` ?

odxtools/multiplexer.py
```diff
@@ -54,14 +54,14 @@ class MultiplexerDefaultCase:
...
     def __post_init__(self):
         self._structure: Optional[BasicStructure] = None
 
-    def _resolve_references(self, id_lookup: Dict[str, Any]) -> None:
+    def _resolve_references(self, odxlinks: OdxLinkDatabase) -> None:
         if self.structure_ref is not None:
-            self._structure = id_lookup.get(self.structure_ref)
+            self._structure = odxlinks.resolve(self.structure_ref)
             if self._structure is None:
                 logger.warning(
```
kayoub5, line 66: This warning will not longer trigger (dead code)

Maybe add `if self.structure_ref` ?

odxtools/singleecujob.py
```diff
@@ -59,8 +60,8 @@ class OutputParam:
     def __post_init__(self):
         self._dop: Optional[DopBase] = None
 
-    def _resolve_references(self, id_lookup: Dict[str, Any]) -> None:
-        self._dop = id_lookup.get(self.dop_base_ref)
+    def _resolve_references(self, odxlinks: OdxLinkDatabase) -> None:
+        self._dop = odxlinks.resolve(self.dop_base_ref)
         if self._dop is None:
             logger.warning(
```
kayoub5, line 66: Dead code for warning again

odxtools/singleecujob.py
```diff
@@ -74,15 +75,15 @@ def dop(self) -> Optional[DopBase]:
...
 
     def __post_init__(self):
         self._dop: Optional[DopBase] = None
 
-    def _resolve_references(self, id_lookup: Dict[str, Any]) -> None:
-        self._dop = id_lookup.get(self.dop_base_ref)
+    def _resolve_references(self, odxlinks: OdxLinkDatabase) -> None:
+        self._dop = odxlinks.resolve(self.dop_base_ref)
         if self._dop is None:
             logger.warning(
```
kayoub5, line 88: And again

odxtools/table.py
```diff
@@ -24,28 +25,28 @@ def __init__(self, id: str, short_name: str, long_name=None):
...
         self._structure: Optional[DopBase] = None
         self._dop: Optional[DopBase] = None
 
-    def _resolve_references(self, id_lookup: Dict[str, Any]) -> None:
+    def _resolve_references(self, odxlinks: OdxLinkDatabase) -> None:
         if self.structure_ref is not None:
-            self._structure = id_lookup.get(self.structure_ref)
+            self._structure = odxlinks.resolve(self.structure_ref)
             if self._structure is None:
                 logger.warning(
```
kayoub5, line 45: and again

andlaus: I hope to have removed all of these instances: https://github.com/mercedes-benz/odxtools/pull/82/commits/fab3efa75a16fe82f32bf2ed482c60d3904c1d94

odxtools/multiplexer.py
andlaus: > Maybe add if self.structure_ref ?

isn't this already there? (I think that we can assume that links to objects which should get internalized are valid. agree?)

odxtools/singleecujob.py
```diff
@@ -34,8 +35,8 @@ class InputParam:
...
```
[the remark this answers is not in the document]
andlaus: according to the XSD, `DOP-BASE-REF` is mandatory for `INPUT-PARAM`:

Screenshot_20221024_161553

odxtools/nameditemlist.py
andlaus: > I have item names that start with a number...

they seem to be invalid IDs, though: if I didn't get this wrong (which is very well possible), `xsd:ID` must start with a letter or an underscore:

http://books.xmlschemata.org/relaxng/ch19-77151.html and https://stackoverflow.com/questions/1631396/what-is-an-xsncname-type-and-when-should-it-be-used

andlaus:
@kayoub5: short names which start with a number should be supported with the latest version of this PR. can you check?

andlaus:
okay. seems like no vetos were voiced, let's take the plunge...

Issue #83, opened 2022-10-18
CAN send/receive ids are null


smileprem:
Hi team - We were using the commit `74163b455f070f8fd925955bb33911bfcd83a092` which was a few weeks old. Recently upgraded to the latest version at `3b33829178dc854e8a578ff66be7459137017391`

With the new version of the library, the can send/receive ids appear as `None`. This was working with the previous version.

A sample test program. 
```
import odxtools

database = odxtools.load_file("./Sibros_TCU_V1.0.1.pdx", enable_candela_workarounds=True)
for dlc in database.diag_layer_containers:
    for ecu_variant in dlc.base_variants:
        print("physical_request_id = ", hex(ecu_variant.get_receive_id()))
        print("physical_response_id = ", hex(ecu_variant.get_send_id()))
```
Error:
```
Traceback (most recent call last):
  File "test.py", line 6, in <module>
    print("physical_request_id = ", hex(ecu_variant.get_receive_id()))
TypeError: 'NoneType' object cannot be interpreted as an integer

```

Gitlab does not allow me to upload the PDX files. Probably you can test with the PDX files that you have.
<img width="957" alt="image" src="https://user-images.githubusercontent.com/5611865/196558520-54919311-3b41-4120-a0ca-920668d15783.png">

**Question:** Are these APIs deprecated?. Should I be looking at a different function?. Please advise.

kayoub5:
@smileprem you can put the PDX in a ZIP and upload it

smileprem:
That's brilliant. Why didn't I think of that?. Here is the PDX file that you can test with.

[Sibros_TCU_V1.0.1.pdx.zip](https://github.com/mercedes-benz/odxtools/files/9822735/Sibros_TCU_V1.0.1.pdx.zip)

Also, I put up an quick MR for this. https://github.com/mercedes-benz/odxtools/pull/84

smileprem:
Hi Team - The latest commit from main branch (cdb078019e41c84f4763d15d44166635b86259e4) fixed the issues we observed with the earlier version. Thanks for fixing the errors. I'll close this ticket.

PR #84, opened 2022-10-18
Updated the IDs for get_receive_id() and get_send_id()

smileprem:
Fixes the issue https://github.com/mercedes-benz/odxtools/issues/83

### DIFF
--- a/odxtools/diaglayer.py
+++ b/odxtools/diaglayer.py
@@ -441,7 +441,7 @@ class DiagLayer:
         """CAN ID to which the ECU listens for diagnostic messages"""
         # TODO (?): When using the dict notation to access comparams,
         #           wouldn't it be prettier to allow the dot in the id?
-        com_param = self.communication_parameters["ISO_15765_2__CP_UniqueRespIdTable"]
+        com_param = self.communication_parameters["ISO_15765_2__CP_UniqueRespIdTable_from_ISO_15765_2"]
         if com_param is None:
             return None
         else:
@@ -464,7 +464,7 @@ class DiagLayer:
 
     def get_send_id(self):
         """CAN ID to which the ECU sends replies to diagnostic messages"""
-        com_param = self.communication_parameters["ISO_15765_2__CP_UniqueRespIdTable"]
+        com_param = self.communication_parameters["ISO_15765_2__CP_UniqueRespIdTable_from_ISO_15765_2"]
         if com_param is None:
             return None
         else:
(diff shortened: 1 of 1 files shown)


odxtools/diaglayer.py
```diff
@@ -441,7 +441,7 @@ def get_receive_id(self):
         """CAN ID to which the ECU listens for diagnostic messages"""
         # TODO (?): When using the dict notation to access comparams,
         #           wouldn't it be prettier to allow the dot in the id?
-        com_param = self.communication_parameters["ISO_15765_2__CP_UniqueRespIdTable"]
+        com_param = self.communication_parameters["ISO_15765_2__CP_UniqueRespIdTable_from_ISO_15765_2"]
```
andlaus, line 444: it could be possible that the name of the document fragment for `ISO_15765_2.CP_UniqueRespIdTable` is different from `ISO_15765_2`, so it's probably a good idea to filter:

```python
cps = [ cp for cp in self.communication_parameters if cp.short_name.startswith("ISO_15765_2__CP_UniqueRespIdTable_from")
assert len(cps) == 1, "Communication parameter `ISO_15765_2.CP_UniqueRespIdTable` not found"
com_param = cps[0]
```

That said, once #82 has been merged, this needs to be modified once more...

andlaus (commented):
this is not yet perfect, but IMO sufficient as a stop-gap measure. I'd prefer the slightly more sophisticated approach, but if you want to have this merged as-is I will do so...

odxtools/diaglayer.py
smileprem: Hi @andlaus - This is a blocker for us. We would really appreciate, if you could approve this MR.

andlaus: I see. is there an issue with going with the `.startswith()` approach from above? I really would like to avoid to hard-code the name of the document fragment of the communication parameters...

smileprem: My bad. From your previous comment, I thought you were OK with this MR. I get this error when running the above code snippet. I don't see a short_name attribute for CommunicationParameterRef.

```
  File "test.py", line 6, in <module>
    print("physical_request_id = ", hex(ecu_variant.get_receive_id()))
  File "/Users/smileprem/Code/sibros/odxtools/odxtools/diaglayer.py", line 445, in get_receive_id
    cps = [cp for cp in self.communication_parameters if cp.short_name.startswith('ISO_15765_2__CP_UniqueRespIdTable_from')]
  File "/Users/smileprem/Code/sibros/odxtools/odxtools/diaglayer.py", line 445, in <listcomp>
    cps = [cp for cp in self.communication_parameters if cp.short_name.startswith('ISO_15765_2__CP_UniqueRespIdTable_from')]
AttributeError: 'CommunicationParameterRef' object has no attribute 'short_name'
```

andlaus: sorry that I was a bit unclear: I *prefer* the more comprehensive solution, but if there is a good reason why it won't work, I will (somewhat grudgingly ;)) merge this PR as a quick fix.

> `AttributeError: 'CommunicationParameterRef' object has no attribute 'short_name'`

ok, replace `short_name` by `id_ref`(I have not tested this before making the original suggestion, sorry...)

andlaus: BTW: #82 should also fix this issue...

smileprem: That's awesome. Let's try to merge the other MR and close this one.

PR #86, opened 2022-10-26
No 'id' attributes anymore and remove `read_element_id()`

andlaus:
`id()` is a build-in function and it is terrible style to overwrite it for some contexts!

Besides these large but rather trivial change, the `read_element_id()` utility function is removed: This function was introduced to reduce the amount of duplicated code, but in my experience, the code which `read_element_id()` replaces is rather trivial and the function obfuscates things quite a bit and it does not play nicly with the `mypy` type checker. (Besides that, its name was IMO pretty confusing because it did *not* create an `OdxLinkId`, but parsed the elements featured by the `ELEMENT-ID` XSD     group...)

### DIFF
--- a/odxtools/audience.py
+++ b/odxtools/audience.py
@@ -39,7 +39,7 @@ class AdditionalAudience:
     """
     Corresponds to ADDITIONAL-AUDIENCE.
     """
-    id: OdxLinkId
+    odx_link_id: OdxLinkId
     short_name: str
     long_name: Optional[str] = None
     description: Optional[str] = None
@@ -69,13 +69,13 @@ def read_audience_from_odx(et_element, doc_frags: List[OdxDocFragment]):
 
 def read_additional_audience_from_odx(et_element, doc_frags: List[OdxDocFragment]):
     short_name = et_element.find("SHORT-NAME").text
-    id = OdxLinkId.from_et(et_element, doc_frags)
-    assert id is not None
+    odx_link_id = OdxLinkId.from_et(et_element, doc_frags)
+    assert odx_link_id is not None
 
     long_name = et_element.findtext("LONG-NAME")
     description = read_description_from_odx(et_element.find("DESC"))
 
-    return AdditionalAudience(id=id,
+    return AdditionalAudience(odx_link_id=odx_link_id,
                               short_name=short_name,
                               long_name=long_name,
                               description=description)
--- a/odxtools/audience.py
+++ b/odxtools/audience.py
@@ -39,7 +39,7 @@ class AdditionalAudience:
     """
     Corresponds to ADDITIONAL-AUDIENCE.
     """
-    odx_link_id: OdxLinkId
+    odx_id: OdxLinkId
     short_name: str
     long_name: Optional[str] = None
     description: Optional[str] = None
@@ -69,13 +69,13 @@ def read_audience_from_odx(et_element, doc_frags: List[OdxDocFragment]):
 
 def read_additional_audience_from_odx(et_element, doc_frags: List[OdxDocFragment]):
     short_name = et_element.find("SHORT-NAME").text
-    odx_link_id = OdxLinkId.from_et(et_element, doc_frags)
-    assert odx_link_id is not None
+    odx_id = OdxLinkId.from_et(et_element, doc_frags)
+    assert odx_id is not None
 
     long_name = et_element.findtext("LONG-NAME")
     description = read_description_from_odx(et_element.find("DESC"))
 
-    return AdditionalAudience(odx_link_id=odx_link_id,
+    return AdditionalAudience(odx_id=odx_id,
                               short_name=short_name,
                               long_name=long_name,
                               description=description)
--- a/odxtools/diagcodedtypes.py
+++ b/odxtools/diagcodedtypes.py
@@ -447,12 +447,12 @@ class ParamLengthInfoType(DiagCodedType):
             # if isinstance(param_value.parameter, LengthKeyParameter) would be prettier,
             # but leads to cyclic import...
             if parameter.parameter_type == "LENGTH-KEY" \
-                    and parameter.id == self.length_key_id: # type: ignore
+                    and parameter.odx_link_id == self.length_key_id: # type: ignore
                 # The bit length of the parameter to be extracted is given by the length key.
                 bit_length = value
                 break
 
-        assert bit_length is not None, f"Did not find any length key with ID {self.length_key_id}"
+        assert bit_length is not None, f"Dodx_link_id not find any length key with ID {self.length_key_id}"
 
         # Extract the internal value and return.
         return self._extract_internal(decode_state.coded_message,
--- a/odxtools/diagcodedtypes.py
+++ b/odxtools/diagcodedtypes.py
@@ -447,7 +447,7 @@ class ParamLengthInfoType(DiagCodedType):
             # if isinstance(param_value.parameter, LengthKeyParameter) would be prettier,
             # but leads to cyclic import...
             if parameter.parameter_type == "LENGTH-KEY" \
-                    and parameter.odx_link_id == self.length_key_id: # type: ignore
+                    and parameter.odx_id == self.length_key_id: # type: ignore
                 # The bit length of the parameter to be extracted is given by the length key.
                 assert isinstance(value, int)
                 bit_length = value
--- a/odxtools/diagcodedtypes.py
+++ b/odxtools/diagcodedtypes.py
@@ -3,7 +3,7 @@
 
 import abc
 import math
-from typing import Any, Union, List
+from typing import Any, Optional, Union, List
 
 from .odxtypes import DataType
 from .exceptions import DecodeError, EncodeError
@@ -38,13 +38,13 @@ class DiagCodedType(abc.ABC):
         self.is_highlow_byte_order = is_highlow_byte_order
 
     def _extract_internal(self,
-                          coded_message,
-                          byte_position,
-                          bit_position,
-                          bit_length,
+                          coded_message: bytes,
+                          byte_position: int,
+                          bit_position: int,
+                          bit_length: int,
                           base_data_type: DataType,
-                          is_highlow_byte_order,
-                          bit_mask=None):
+                          is_highlow_byte_order: bool,
+                          bit_mask: Optional[int] = None):
         """Extract the internal value.
 
         Helper method for `DiagCodedType.convert_bytes_to_internal`.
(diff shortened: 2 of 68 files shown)


odxtools/audience.py
```diff
@@ -39,7 +39,7 @@ class AdditionalAudience:
     """
     Corresponds to ADDITIONAL-AUDIENCE.
     """
-    id: OdxLinkId
+    odx_link_id: OdxLinkId
```
kayoub5, line 42: Is the `odx_` prefix really needed?

odxtools/diagcodedtypes.py
```diff
@@ -447,12 +447,12 @@ def convert_bytes_to_internal(self, decode_state: DecodeState, bit_position: int
...
             # but leads to cyclic import...
             if parameter.parameter_type == "LENGTH-KEY" \
-                    and parameter.id == self.length_key_id: # type: ignore
+                    and parameter.odx_link_id == self.length_key_id: # type: ignore
                 # The bit length of the parameter to be extracted is given by the length key.
                 bit_length = value
                 break
 
-        assert bit_length is not None, f"Did not find any length key with ID {self.length_key_id}"
+        assert bit_length is not None, f"Dodx_link_id not find any length key with ID {self.length_key_id}"
```
kayoub5, line 455: Wrong change

odxtools/diagdatadictionaryspec.py
```diff
@@ -217,7 +217,7 @@ def read_diag_data_dictionary_spec_from_odx(et_element, doc_frags: List[OdxDocFr
     ]:
         num = len(list(et_element.iterfind(path)))
         if num > 0:
-            logger.info(f"Not implemented: Did not parse {num} {name}.")
+            logger.info(f"Not implemented: Dodx_link_id not parse {num} {name}.")
```
kayoub5, line 220: Wrong change

odxtools/diaglayer.py
```diff
@@ -751,15 +753,16 @@ def __str__(self) -> str:
 
 
 def read_diag_layer_container_from_odx(et_element, enable_candela_workarounds=True):
-    short_name = et_element.find("SHORT-NAME").text
+    short_name = et_element.findtext("SHORT-NAME")
+    assert short_name is not None
```
kayoub5, line 757: Not really related to the PR, but please raise `OdxError` instead of assert

odxtools/diaglayer.py
```diff
@@ -751,15 +753,16 @@ def __str__(self) -> str:
...
     long_name = et_element.findtext("LONG-NAME")
 
     # create the current ODX "document fragment" (description of the
     # current document for references and IDs)
     doc_frags = [ OdxDocFragment(short_name, "CONTAINER") ]
 
-    id = OdxLinkId.from_et(et_element, doc_frags)
-
+    odx_link_id = OdxLinkId.from_et(et_element, doc_frags)
+    assert odx_link_id is not None
```
kayoub5, line 765: Are you sure id is required here?

odxtools/isotp_state_machine.py
```diff
@@ -314,11 +314,11 @@ def on_consecutive_frame(self, telegram_idx, segment_idx, frame_payload):
         super().on_consecutive_frame(telegram_idx, segment_idx, frame_payload)
 
 
-    def _send_can_message(self, id, payload):
+    def _send_can_message(self, odx_link_id, payload):
         if len(payload) < self._padding_size:
             payload = bytes(payload)+bytes([self._padding_value]*(self._padding_size - len(payload)))
 
-        msg = can.Message(arbitration_id=id,
+        msg = can.Message(arbitration_id=odx_link_id,
```
kayoub5, line 321: Not an odx link id

odxtools/parameters/readparameter.py
```diff
@@ -22,19 +22,13 @@
...
+    description = read_description_from_odx(et_element.find("DESC"))
     semantic = et_element.get("SEMANTIC")
-
-    byte_position = int(et_element.find(
-        "BYTE-POSITION").text) if et_element.find("BYTE-POSITION") is not None else None
-    bit_position = int(et_element.find(
-        "BIT-POSITION").text) if et_element.find("BIT-POSITION") is not None else 0
-
+    byte_position = int(et_element.findtext("BYTE-POSITION", "0"))
+    bit_position = int(et_element.findtext("BIT-POSITION", "0"))
```
kayoub5, line 31: Invalid default value `0` default value MUST be `None`

odxtools/structures.py
```diff
@@ -25,14 +25,14 @@
...
                  short_name,
                  parameters: Iterable[Union[Parameter, "EndOfPduField"]],
                  long_name=None,
                  byte_size=None,
                  description=None):
-        super().__init__(id, short_name, long_name=long_name, description=description)
-        self.parameters : NamedItemList[Union["Parameter", "EndOfPduField"]] = \
+        super().__init__(odx_link_id, short_name, long_name=long_name, description=description)
+        self.parameters : NamedItemList[Union[Parameter, "EndOfPduField"]] = NamedItemList(lambda par: par.short_name, parameters) # type: ignore
             NamedItemList(short_name_as_id, parameters)
```
kayoub5, line 36: Something wrong happened here, line 36 looks out of place

odxtools/audience.py
andlaus: I'm a bit on the edge here: `odx_link_id` is rather clunky, but the ODX spec calls the mechanism "ODXLINK", so it is a bit more explicit to call the attributes `odx_link_id` or `odxlink_id`. that said, I don't mind `link_id` or `odx_id` either. What do you prefer? (BTW: if it wasn't for the fact that the name is already taken, I would still go with `id`.)

odxtools/diaglayer.py
andlaus: good suggestion. Let's keep it in mind. (in order to keep the code churn at an acceptable level, we should probably introduce an `odx_assert()` function for that?)

odxtools/diaglayer.py
andlaus: no, I wasn't. That said the XSD says that DIAG-LAYER-CONTAINERs extend ODX-CATEGORIES which in turn extend ELEMENT-ID for which an iD is mandatory. (tl;dr: now I am sure ;))

odxtools/diaglayer.py
kayoub5: Yes

odxtools/audience.py
kayoub5: We call the `ID-REF,DOCREF` `odx_link` and the `ID` `odx_id` no point putting the word "link" in the id

andlaus:
@kayoub5: I hope to have addressed the issues you've found...

andlaus:
lacking vetos, I intend to merge this PR tomorrow...

odxtools/diaglayer.py
```diff
@@ -751,15 +753,16 @@ def __str__(self) -> str:
...
     long_name = et_element.findtext("LONG-NAME")
 
     # create the current ODX "document fragment" (description of the
     # current document for references and IDs)
     doc_frags = [ OdxDocFragment(short_name, "CONTAINER") ]
 
-    id = OdxLinkId.from_et(et_element, doc_frags)
-
+    odx_id = OdxLinkId.from_et(et_element, doc_frags)
+    assert odx_id is not None
```
kayoub5, line 765: Are you sure ID is required here?

odxtools/multiplexer.py
```diff
@@ -76,9 +76,15 @@ class MultiplexerSwitchKey:
     """This class represents a Switch Key, which is used to select one of the cases defined in the Multiplexer."""
 
     byte_position: int
-    bit_position: int
+    bit_position: Optional[int]
```
kayoub5, line 79: Default bit_position is 0 for multiplexer
```suggestion
    bit_position: int
```

odxtools/multiplexer.py
```diff
@@ -76,9 +76,15 @@ class MultiplexerSwitchKey:
...
     dop_ref: OdxLinkRef
 
+    @property
+    def bit_position_int(self) -> int:
+        if self.bit_position is None:
+            return 0
+        return self.bit_position
```
kayoub5, line 86: You don't need this
```suggestion
```

odxtools/parameters/parameterbase.py
```diff
@@ -18,17 +18,29 @@ def __init__(self,
...
+        if self.byte_position is None:
+            return 0
+        return self.byte_position
+
+    @property
+    def bit_position_int(self) -> int:
+        if self.bit_position is None:
+            return 0
+        return self.bit_position
+
```
kayoub5, line 43: Those are extremely misleading, a byte/bit position None for parameter means "Encode/Decode this parameter just after the previous one"

Any logic that tries to imply 0 from a None will most likely be incorrect.

andlaus: the problem is that a lot of places in the code assumed that `bit_position` and 'byte_position' are integers. I introduced these functions to not having to do the checking manually all the time, but maybe it should be done that way...

andlaus: ok. I have removed them: https://github.com/mercedes-benz/odxtools/pull/86/commits/9fbc7388a6836ea99432edd3c189a9d8b26e8ad5. Hopefully everything still works as before...

odxtools/diaglayer.py
andlaus: after checking with the XSD, I am: DIAG-LAYER-CONTAINER is an extension of ODX-CATEGORY which in turn extends ELEMENT-ID..

odxtools/multiplexer.py
andlaus: according to the XSD, it isn't:

image

maybe I made a mistake, though...

kayoub5: For switch key, a no bit position is equivalent to a bit position of 0, the same logic does not apply to parameters

andlaus: yep, but it should still be possible to distinguish between "BIT-POSITION" not specified and it being 0 for writing the database back into a PDX file. also, it would be pretty confusing to handle it differently than the parameters...

kayoub5:
Got an error, bit_position = None

```
unsupported operand type(s) for +: 'int' and 'NoneType'
odxtools\diagcodedtypes.py", line 110, in _to_bytes
    offset = (8 - ((bit_length + bit_position) % 8)) % 8
```

andlaus:
> unsupported operand type(s) for +: 'int' and 'NoneType'
> odxtools\diagcodedtypes.py", line 110, in _to_bytes
>     offset = (8 - ((bit_length + bit_position) % 8)) % 8

stuff like that was the reason for the `_int` properties ;). can you provide a full traceback?

kayoub5:
Caused by calling `struct.encode(...)`
```
  File "odxtools\diagcodedtypes.py", line 110, in _to_bytes
    offset = (8 - ((bit_length + bit_position) % 8)) % 8
  File "odxtools\diagcodedtypes.py", line 496, in convert_internal_to_bytes
    return self._to_bytes(internal_value,
  File "odxtools\dataobjectproperty.py", line 100, in convert_physical_to_bytes
    return self.diag_coded_type.convert_internal_to_bytes(internal_val, encode_state, bit_position=bit_position)
  File "odxtools\multiplexer.py", line 159, in convert_physical_to_bytes
    key_bytes = self.switch_key._dop.convert_physical_to_bytes(
  File "odxtools\parameters\valueparameter.py", line 54, in get_coded_value_as_bytes
    return self.dop.convert_physical_to_bytes(physical_value,
  File "odxtools\parameters\parameterbase.py", line 108, in encode_into_pdu
    byte_value = self.get_coded_value_as_bytes(encode_state)
  File "odxtools\structures.py", line 146, in convert_physical_to_internal
    coded_rpc = param.encode_into_pdu(encode_state)
  File "odxtools\structures.py", line 266, in encode
    return self.convert_physical_to_internal(params,
```

kayoub5:
See `multiplexer.py`, line 160, pass `self.switch_key.bit_position` to `convert_physical_to_bytes` without None checking

andlaus:
> See multiplexer.py, line 160, self.switch_key.bit_position passed to convert_physical_to_bytes without None checking

ok. should be fixed now...

kayoub5 (approved):

andlaus:
let's get this in...

PR #90, opened 2022-11-10
Use findtext where applicable

### LINKED ISSUE #58: Use findtext() instead of et_element.find("XML-ELEMENT").text
findtext() seems to be a better function. The documentation says

```
findtext(match, default=None, namespaces=None)

Finds text for the first subelement matching match. match may be a tag name or a [path](https://docs.python.org/3/library/xml.etree.elementtree.html#elementtree-xpath). Returns the text content of the first matching element, or default if no element was found. Note that if the matching element has no text content an empty string is returned. namespaces is an optional mapping from namespace prefix to full name. Pass '' as prefix to move all unprefixed tag names in the expression into the given namespace.
```

kayoub5:
Resolves #58

### DIFF
--- a/odxtools/admindata.py
+++ b/odxtools/admindata.py
@@ -69,9 +69,7 @@ def read_admin_data_from_odx(et_element, doc_frags: List[OdxDocFragment]):
     if et_element is None:
         return None
 
-    language = et_element.find("LANGUAGE")
-    if language is not None:
-        language = language.text
+    language = language.findtext("LANGUAGE")
 
     company_doc_infos = et_element.find("COMPANY-DOC-INFOS")
     if company_doc_infos is not None:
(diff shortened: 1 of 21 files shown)


odxtools/admindata.py
```diff
@@ -69,9 +69,7 @@ def read_admin_data_from_odx(et_element, doc_frags: List[OdxDocFragment]):
     if et_element is None:
         return None
 
-    language = et_element.find("LANGUAGE")
-    if language is not None:
-        language = language.text
+    language = language.findtext("LANGUAGE")
```
kayoub5, line 72:
```suggestion
    language = et_element.findtext("LANGUAGE")
```

odxtools/compumethods/readcompumethod.py
```diff
@@ -147,13 +147,13 @@ def read_compu_method_from_odx(et_element,
...
             internal_to_phys.append(CompuScale(
-                short_label=(scale.find("SHORT-LABEL").text
+                short_label=(scale.findtext("SHORT-LABEL")
                              if scale.find("SHORT-LABEL") is not None else None),
                 description=read_description_from_odx(scale.find("DESC")),
                 lower_limit=lower_limit,
                 upper_limit=upper_limit,
                 compu_inverse_value=compu_inverse_value,
-                compu_const=scale.find("COMPU-CONST").find("VT").text
+                compu_const=scale.findtext("COMPU-CONST").find("VT")
```
kayoub5, line 156:
```suggestion
                compu_const=scale.find("COMPU-CONST").findtext("VT")
```

andlaus (approved):
looks good: unit tests still work (whatever that is worth), and my internal files seem to work as before.

PR #92, opened 2022-11-13
Support for COMPARAM-SUBSET

kayoub5:


kayoub5:
@andlaus I can add a test by using the sample PDX from https://github.com/sebastianwilczek/CANoe-Configurations/tree/master/Ethernet/Diagnostics/EthernetDiagnostics/ODX

Creating a full com param configuration using python sounds like an overkill.

odxtools/communicationparameter.py
```diff
@@ -14,41 +38,39 @@ def __init__(self,
...
-        if not isinstance(self.value, list):
-            val = f"'{val}'"
+        return self.__repr__()
 
-        return f"CommunicationParameter('{self.id_ref}', value={val})"
+    def _resolve_references(self, odxlinks: OdxLinkDatabase):
+        # Temporary lenient until tests are updated
+        self.comparam = odxlinks.resolve_lenient(self.id_ref)
+        if not self.comparam:
+            warnings.warn(f"Could not resolve COMPARAM '{self.id_ref}'", OdxWarning)
```
kayoub5, line 61: @andlaus this fails for the `somersaultecu` due to missing com param, but works fine for other PDX files

odxtools/communicationparameter.py
```diff
@@ -1,10 +1,34 @@
...
+
+    if isinstance(comparam, Comparam):
+        return comparam.dop.physical_type.base_data_type.from_string(value)
+
+    if isinstance(comparam, ComplexComparam) and isinstance(value, list):
+        value = value + [None] * (len(comparam.comparams) - len(value))
+        result = dict()
+        for i, cp in enumerate(comparam.comparams):
+            result[cp.short_name] = _get_comparam_value(cp, value[i])
+        return result
```
kayoub5, line 28: @andlaus I am not 100% sure, since I don't have a PDX to check on, but this should remove the need for "candela workarounds"

odxtools/diaglayer.py
```diff
@@ -457,50 +459,29 @@ def get_communication_parameter(self, cp_id: str) \
...
-            # both are different from the one of the COMPARAM
-            # fragment included in the MCD2-D standard
-            try:
-                return int(com_param.value[2])
-            except ValueError:
-                return None
-        else:
-            # assume the parameter order specified by the COMPARAM
-            # fragment of the ASAM MCD2-D standard.
-            return int(com_param.value[1])
```
kayoub5, line 479: @andlaus you may want to double check this, the new logic is name based so `enable_candela_workarounds` should not be needed anymore, but I don't actually have the names that the code was supposed to point to.

andlaus:
> @andlaus I can add a test by using the sample PDX from https://github.com/sebastianwilczek/CANoe-Configurations/tree/master/Ethernet/Diagnostics/EthernetDiagnostics/ODX

if @sebastianwilczek can confirm that the file can be distributed under the MIT license, yes. If we can't get that affirmation, a second repository (`odxtools-additional-tests`?) could be created for that purpose. (this repo must not be affiliated with mercedes-benz, though.)

> Creating a full com param configuration using python sounds like an overkill.

I agree. at least in the short and medium term ;)

odxtools/communicationparameter.py
andlaus: I have a fix for that: https://github.com/kayoub5/odxtools/pull/1 (this is a PR against this PR, if you agree, please merge.)

sebastianwilczek:
Hey everyone, not sure about the context here, but the linked repository is only a modified version of the configuration files used by CANoe, developed and owned by Vector Informatik GmbH: https://www.vector.com/int/en/download/?tx_vecdownload_downloaddetail%5Bdownload%5D=92521

As far as I can tell, they do not list any form of license (in context of MIT or otherwise), so unfortunately I can not tell you whether you may use any of the files in your processes.

odxtools/communicationparameter.py
andlaus: if possible, I'm all for it. When we implemented this, we noticed that the files written by candela use a different field for the can IDs than specified by the `.odx-cs` fragments shipped with the standard. (I think that I also looked up if the `pdx` files were internally and I thought it wasn't, but since I did this manually, I probably screwed up while doing this...)

andlaus:
I tested this PR with our internal files written by candela studio, and it seems that after this PR, the send and receive IDs (`ecu.get_send_id()` and `ecu.get_receive_id()`) are swapped...

andlaus:
@sebastianwilczek: okay, I suppose we should ask vector then. I'll keep you updated.

odxtools/diaglayer.py
andlaus: this PR swaps `ecu.get_send_id()` and `ecu.get_receive_id()`. It might be an issue of CANdela generated files? (does it work correctly with your files?)

odxtools/diaglayer.py
```diff
@@ -457,50 +459,29 @@ def get_communication_parameter(self, cp_id: str) \
...
+        value = com_param.value
+        if not isinstance(value, dict):
+            return None
+            
+        return value.get('CP_CanRespUSDTId', None)
```
kayoub5, line 469:
```suggestion
        return value.get('CP_CanPhysReqId', None)
```

kayoub5:
> I tested this PR with our internal files written by candela studio, and it seems that after this PR, the send and receive IDs (`ecu.get_send_id()` and `ecu.get_receive_id()`) are swapped...

should be correct now

kayoub5:
@andlaus I think you forgot to commit the `ISO_11898_2_DWCAN.odx-cs` file, among others

kayoub5:
@andlaus FYI, you push directly into a fork if the author enabled it (checked "Allow edits by maintainers")

So feel free to push changes directly to my branch.

andlaus:
> @andlaus I think you forgot to commit the ISO_11898_2_DWCAN.odx-cs file, among others

oops. I had some unpacked `somersault.pdx` file lying around in the directory where I tested this, i.e. I loaded these fragments from the incorrect location. should be fixed now: https://github.com/kayoub5/odxtools/pull/2

andlaus:
> So feel free to push changes directly to my branch.

thanks for the offer, but I think it is less confusing if I don't do this (I always get slightly dizzy if some of my PRs change without my intervention...)

andlaus:
btw, I now get the following warning:
```
/home/and/src/odxtools/odxtools/diaglayer.py:453: OdxWarning: Communication parameter `CP_UniqueRespIdTable` specified more than once. Using first occurence.
```

I assume that this is caused by the fact that the ECU can be interacted with using multiple protocols (DoIP and CAN). I guess that fixing this requires a full understanding of the protocol concept and should thus be left for another day?

andlaus:
ok, I've tested this with our internal files, and it seems to work without the `enable_candela_workarounds` hack. can you remove all traces of this blemish?

kayoub5:
> ok, I've tested this with our internal files, and it seems to work without the `enable_candela_workarounds` hack. can you remove all traces of this blemish?

We can do that in a different PR, this one is too loaded as it is

odxtools/units.py
```diff
@@ -3,11 +3,13 @@
 
 from dataclasses import dataclass, field
 from typing import Dict, List, Any, Literal, Optional, Union
+import warnings
```
kayoub5, line 6:
```suggestion
```

odxtools/units.py
```diff
@@ -3,11 +3,13 @@
...
 
 from .utils import short_name_as_id
 from .nameditemlist import NamedItemList
 from .utils import read_description_from_odx
 from .odxlink import OdxLinkRef, OdxLinkId, OdxLinkDatabase, OdxDocFragment
+from .exceptions import OdxWarning
```
kayoub5, line 12:
```suggestion
```

kayoub5:
> btw, I now get the following warning:
> 
> ```
> /home/and/src/odxtools/odxtools/diaglayer.py:453: OdxWarning: Communication parameter `CP_UniqueRespIdTable` specified more than once. Using first occurence.
> ```
> 
> I assume that this is caused by the fact that the ECU can be interacted with using multiple protocols (DoIP and CAN). I guess that fixing this requires a full understanding of the protocol concept and should thus be left for another day?

You can cross check this by checking that the different com params point to different proto stacks

It could even be different CAN buses.

odxtools/diaglayer.py
```diff
@@ -156,7 +155,7 @@ def __init__(self,
         self._services: NamedItemList[Union[DiagService, SingleEcuJob]]\
             = NamedItemList(short_name_as_id, [])
         self._communication_parameters: NamedItemList[CommunicationParameterRef]\
-            = NamedItemList(lambda cp: cp._python_name(), [])
```
kayoub5, line 159: @andlaus I believe with this change, all `NamedItemList` are now using `short_name_as_id`, thus could be simplified.

andlaus: i agree.

kayoub5:
@andlaus anything blocks merging this?

andlaus:
> @andlaus anything blocks merging this?

no. nice work, thanks!

PR #94, opened 2022-11-17
internalize admin data for comparam subsets

andlaus:
This internalizes the `COMPANY-DATAS` and `ADMIN-DATA` tags for comparam subsets. Besides this, the `allow_multiple_values` can now be `None` to be able to determine if the XML attribute was set in the input file.

### DIFF
--- a/odxtools/companydata.py
+++ b/odxtools/companydata.py
@@ -91,9 +91,9 @@ def read_xdoc_from_odx(xdoc):
                 )
 
 def read_company_datas_from_odx(et_element, doc_frags: List[OdxDocFragment]) \
-        -> Optional[NamedItemList[CompanyData]]:
+        -> NamedItemList[CompanyData]:
     if et_element is None:
-        return None
+        return NamedItemList(short_name_as_id)
 
     cdl = NamedItemList(short_name_as_id) # type: ignore
 
--- a/odxtools/companydata.py
+++ b/odxtools/companydata.py
@@ -90,7 +90,8 @@ def read_xdoc_from_odx(xdoc):
                 position=position,
                 )
 
-def read_company_datas_from_odx(et_element, doc_frags: List[OdxDocFragment]):
+def read_company_datas_from_odx(et_element, doc_frags: List[OdxDocFragment]) \
+        -> Optional[NamedItemList[CompanyData]]:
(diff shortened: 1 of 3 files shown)


odxtools/comparam_subset.py
```diff
@@ -58,7 +60,7 @@ def _build_odxlinks(self) -> Dict[OdxLinkId, Any]:
 @dataclass()
 class ComplexComparam(BaseComparam):
     comparams: NamedItemList[BaseComparam]
-    allow_multiple_values: bool = False
+    allow_multiple_values: Optional[bool] = None
```
kayoub5, line 63: xsd is clear about this one, if it's not specified in the xml, then it is automatically false

odxtools/companydata.py
```diff
@@ -90,7 +90,8 @@ def read_xdoc_from_odx(xdoc):
                 position=position,
                 )
 
-def read_company_datas_from_odx(et_element, doc_frags: List[OdxDocFragment]):
+def read_company_datas_from_odx(et_element, doc_frags: List[OdxDocFragment]) \
+        -> Optional[NamedItemList[CompanyData]]:
```
kayoub5, line 94: I think it would be better to return an empty list rather than return none

odxtools/comparam_subset.py
andlaus: yes, but in order to be able to write the files the same way they are read, it must be possible to distinguish "attribute not present" from "attribute is 'false'". that said, `if comparam_subset.allow_multiple_values` works as before because `if None` behaves as `if False`...

odxtools/companydata.py
andlaus: same as above: we must be able to distinguish "not specified" and "empty list" somehow.

andlaus: yes, but the `COMPANY-DATAS` tag is optional:
```xsd
            <xsd:element type="COMPANY-DATAS" name="COMPANY-DATAS" minOccurs="0" maxOccurs="1"/>
```

anyway, I this can be distinguished by checking if the list is empty or not. fixed, thanks for the persistence. (there are still plenty of other occurences where this happens, though...)

kayoub5 (approved):

PR #95, opened 2022-11-17
Allow ignoring non-critical decoding errors

kayoub5:
# Use case
During ECU development some constants gets added to list of coded_values or a previously reserved byte gets assigned a meaning or a struct gets added a new parameter in a backward compatible way.

Those kind of decoding errors could be tolerated if needed.

This PR change those errors to warnings, but still do raise them by default as errors through `warnings.simplefilter('error', DecodeError)`


andlaus:
looks good, thanks!

kayoub5:
@andlaus could you push a version with this change?

andlaus:
done

PR #96, opened 2022-11-18
Comparam subsets: add write support

andlaus:
This PR adds write support for COMPARAM-SUBSET. The main implication of this is that the communication parameters of a database written with odxtools should now always come with the correct definitions of their semantics.

Besides this, this PR fixes some minor issues which prevented me from writing back my internal diagnostics files using the [pdxcopy](https://github.com/mercedes-benz/odxtools/blob/main/examples/pdxcopy.py) example, and it renames all files containing Jinja2 templates to `.xml.jinja2` to make IDEs like VSCode happy.


odxtools/templates/macros/printComparam.xml.jinja2
```diff
@@ -0,0 +1,75 @@
...
+ {%- if cp.description and cp.description.strip() %}
+ <DESC>
+ {{cp.description}}
+ </DESC>
+ {%- endif %}
+ {%- for sub_cp in cp.comparams %}
+ {{- printGenericComparam(sub_cp) | indent(1, first=True) }}
+ {%- endfor %}
+ {%- if cp.physical_default_value is not none %}
+ <COMPLEX-PHYSICAL-DEFAULT-VALUE>{{cp.physical_default_value}}</COMPLEX-PHYSICAL-DEFAULT-VALUE>
```
kayoub5, line 72: not valid

andlaus: hm, I see: this is one of these overcomplicated constructs again. what got me confused is that ComparamSubset `physical_default_value()` folds this into `physical_default_value` instead of `complex_physical_default_value`. I think these two should be separated because the first is for simple parameters while the second is for complex ones?

odxtools/write_pdx_file.py
```diff
@@ -11,61 +11,85 @@
...
+from .comparam_subset import BaseComparam, Comparam, ComplexComparam
+
+odxdatabase = None
+
+def jinja2_odxraise_helper(msg: str) -> None:
     raise Exception(msg)
 
+def comparam_subset_name(x: BaseComparam) -> str:
+    """Find the short name of the comparam subset which contains a
+    given communication parameter"""
```
kayoub5, line 24: why not just use the doc frag of the odx_id?

odxtools/templates/macros/printComparam.xml.jinja2
kayoub5: Using complex_physical_default_value seemed like an overkill, since physical_default_value can support multiple types (based on the dop physical type) and it would force users to do extra type checking for a common property

odxtools/write_pdx_file.py
andlaus: good point

andlaus:
ok, I hopefully fixed the issues you've found. During that exercise I noticed that we should consider raising the minimum python version to 3.10 to be able to use `kw_only=True` for dataclasses. Unfortunately, it is probably a bit too early for that. Opinions, @kayoub5?

kayoub5:
> ok, I hopefully fixed the issues you've found. During that exercise I noticed that we should consider raising the minimum python version to 3.10 to be able to use `kw_only=True` for dataclasses. Unfortunately, it is probably a bit too early for that. Opinions, @kayoub5?

Too early for min version of 3.10

andlaus:
ok, let's defer this then...

odxtools/templates/macros/printComparam.xml.jinja2
andlaus: If I got this right, the typing of this looks something like this:

```
SimpleValue = str
ComplexValue = List[Union[SimpleValue, ComplexValue]]
```

IMO, conflating these into the same attribute would make things quite confusing (read: I was quite confused), and also, the `COMPLEX-COMPARAM` and `COMPARAM` tags also use `COMPLEX-PHYSICAL-DEFAULT-VALUE` vs `PHYSICAL-DEFAULT-VALUE`. I've adapted the parameter extraction code in the [latest commit](https://github.com/mercedes-benz/odxtools/pull/96/commits/c71e1d19cbff01957b5bd0615172ad3e3da3a1c9), and I think the new version is easier to read. (and also takes the default values into account...)

andlaus:
alright. I've just pushed c71e1d19cbff01957b5bd0615172ad3e3da3a1c9, and I think this is now ready to go. can you check if it still works for your files?

andlaus:
(If there are no vetos, I'll merge this on Friday (Nov 25)...)

odxtools/diaglayer.py
```diff
@@ -454,49 +455,115 @@ def get_communication_parameter(self, name: str) \
 
         return cps[0]
 
+    def _get_comparam_value(self,
+                            com_param : CommunicationParameterRef) \
```
kayoub5, line 459: I think it would be better to make this function part of the `CommunicationParameterRef` or Comparam/ComplexComparam
* It's not using "self"
* It's using only `com_param.value` and `com_param.comparam`

odxtools/diaglayer.py
```diff
@@ -454,49 +455,115 @@ def get_communication_parameter(self, name: str) \
...
+        if com_param.value:
+            result = com_param.value
+        else:
+            result = com_param.comparam.physical_default_value
+
+        assert isinstance(result, str)
+
+        return result
+
+    def _get_comparam_subvalue(self,
```
kayoub5, line 479: What happens when a complex param also contains a complex?

odxtools/diaglayer.py
```diff
@@ -454,49 +455,115 @@ def get_communication_parameter(self, name: str) \
...
+
     def get_receive_id(self) -> Optional[int]:
         """CAN ID to which the ECU listens for diagnostic messages"""
         com_param = self.get_communication_parameter("CP_UniqueRespIdTable")
         if com_param is None:
             return None
-        value = com_param.value
-        if not isinstance(value, dict):
+
+        result = self._get_comparam_subvalue(com_param, "CP_CanPhysReqId")
```
kayoub5, line 521: shouldn't `_get_comparam_subvalue` return an int if its DOP says?

odxtools/diaglayer.py
```diff
@@ -508,11 +575,15 @@ def get_tester_present_time(self) -> Optional[float]:
...
-            return 3.0  # default specified by the standard
+            return None
+
+        result = self._get_comparam_value(com_param)
+        if not result:
+            return None
+        assert isinstance(result, str)
 
-        return int(com_param.value) / 1e6
+        return int(float(result)/1e6)
```
kayoub5, line 586: get_tester_present_time typehint says it should return float

odxtools/diaglayer.py
andlaus: thanks for the suggestions. I moved these methods to the `CommunicationParameterRef` class where they belong anyway...

odxtools/diaglayer.py
andlaus: bad luck ;)

(I think for the standardized communication parameters this is never the case?)

odxtools/diaglayer.py
andlaus: the `CommunicationParameterRef` class stores all values as strings ([[simple](https://github.com/mercedes-benz/odxtools/blob/main/odxtools/communicationparameter.py#L85)] [[complex](https://github.com/mercedes-benz/odxtools/blob/18c138ac29085f6dcd680b1a6b6d09e8a2629893/odxtools/comparam_subset.py#L31)])

This is an artifact from the time when the comparam specs where not yet internalized and it should probably be changed. (But not in this PR, because it has become a bit unwieldy already...)

odxtools/diaglayer.py
andlaus: oops, fixed. (who specifies fractional seconds as tester present times anyway? ;))

andlaus:
ok, I pushed an update: 80bc8bf81aa5182745a7949e919129c56917b2eb

odxtools/communicationparameter.py
```diff
@@ -1,44 +1,25 @@
...
-
-
-def _get_comparam_value(comparam: BaseComparam, value):
-    if not comparam:
-        return value
-    if value is None:
-        value = comparam.physical_default_value
-
-    if isinstance(comparam, Comparam):
-        return comparam.dop.physical_type.base_data_type.from_string(value)
```
kayoub5, line 21: @andlaus original code here did take into consideration the dop physical data type

odxtools/communicationparameter.py
```diff
@@ -1,44 +1,25 @@
...
-
-    if isinstance(comparam, ComplexComparam) and isinstance(value, list):
-        value = value + [None] * (len(comparam.comparams) - len(value))
-        result = dict()
-        for i, cp in enumerate(comparam.comparams):
-            result[cp.short_name] = _get_comparam_value(cp, value[i])
```
kayoub5, line 27: @andlaus original code here did take into account nested complex

odxtools/communicationparameter.py
andlaus: yes, but that was a mightily confusing layer above `CommunicationParameterRef.value`. I suppose that in the future, this should be moved to `CommunicationParameterRef`. (which is a bit more work than it first seems, because of the writing stuff...)

odxtools/communicationparameter.py
andlaus: yeah, let's burn that bridge when we come to it (again)

andlaus:
ok. I guess it is time to get this in...

PR #97, opened 2022-11-22
Support parsing GLOBAL-NEG-RESPONSES

smileprem:
The code to parse the responses already exists in our library. This PR simply adds the logic to parse the `GLOBAL-NEG-RESPONSES` using the existing codebase.

### DIFF
--- a/odxtools/table.py
+++ b/odxtools/table.py
@@ -92,6 +92,8 @@ class TableRow:
     def _build_odxlinks(self) -> Dict[OdxLinkId, Any]:
         result = {}
 
+        result[self.odx_id] = self
+
         for sdg in self.sdgs:
             result.update(sdg._build_odxlinks())
 


andlaus:
looks good. could you also add support for writing? (and possibly these response codes should also be checked in the [decode_reponse()](https://github.com/mercedes-benz/odxtools/blob/main/odxtools/diaglayer.py#L418) method?)

smileprem:
yeah sure. I'll add them early next week. Happy Thanksgiving to all of you (not sure if this is the right place to convey these messages)

andlaus:
this has now quite  severe conflicts with main, most likely coming from running black.

also, besides not dealing with the de-/encoding logic, this PR does not seem to deal with the inheritance of global negative responses from the parent diag layers ('NOT-INHERITED-GLOBAL-NEG-RESPONSES'). (Implementing this should be basically  a copy-paste-modify excercise of e.g. the code for services in `DiagLayer`.)

PR #99, opened 2022-12-01
add support for special data groups (SDGs)

andlaus:
Special Data Groups (SDGs) are used to store data which is not yet standardized in a semi-structured manner. Be aware that I might have missed a place or two where SDGs are allowed because the inheritance hierarchy defined by the ODX XSD is not identical to the one of the python code.

Besides support for SDGs, this patch fixes various issues which I've stumbled over while testing it. (Side note: we *need* more comprehensive unittests!)


odxtools/odxlink.py
```diff
@@ -175,6 +176,9 @@ def resolve(self, ref: OdxLinkRef) -> Any:
 
             obj = doc_frag_db.get(odx_id)
             if obj is not None:
+                if odx_id.local_id == "stupid.sdg.caption":
+                    self._stupid_sdg_resolved = True
```
kayoub5, line 180: what?

odxtools/structures.py
```diff
@@ -301,16 +307,29 @@ def parameter_dict(self) -> ParameterDict:
...
+        super()._resolve_references(odxlinks)
+
         for p in self.parameters:
             if isinstance(p, ParameterWithDOP):
                 p.resolve_references(parent_dl, odxlinks)
-            if isinstance(p, TableKeyParameter):
+            elif isinstance(p, TableKeyParameter):
                 p.resolve_references(parent_dl, odxlinks)
+            else:
+                p._resolve_references(odxlinks)
```
kayoub5, line 332: it would be better to make this a common function in base class of parameters

kayoub5:
@andlaus I have limited access to my pc those days due to health issues, don't block merging waiting for my replies if I take too long

odxtools/odxlink.py
andlaus: whoops. that is an artifact of the instrumentation code which I used to develop this: I added an SDG which contained an SDG-caption with the ID "stupid.sdg.caption" to the XML tag I wanted to test and ensured that it was seen during the ID resolution. (using this method I found a few classes which had issues with ID resolution, BTW.)

odxtools/structures.py
andlaus: this is a hack to be able to resolve SN-REFs: because to be able to do so in the limited way that mattered for us, we needed to access the parent diaglayer for those types of parameters where they are commonly used. I intend to be resolve SN-REFs properly at some point (probably by extending the `OdxLinkDatabase` class), but that is another major refactoring and thus left out of this PR for now. (in the mean time, naming the `_resolve_references()` method differently for such classes would significantly reduce the confusion?)

andlaus:
> @andlaus I have limited access to my pc those days due to health issues, don't block merging waiting for my replies if I take too long

get well soon :)

odxtools/structures.py
kayoub5: add resolve_references(parent_dl, odxlinks) to parameter class, default implementation does nothing, and override it in sub classes.

andlaus:
alright, it's time to get this merged...

PR #101, opened 2022-12-12
Fix multi-protocol comparams

dmholtz:
This PR proposes a fix for handling comparams with same names but different protocols.


odxtools/communicationparameter.py
```diff
@@ -1,26 +1,29 @@
...
 from .utils import read_description_from_odx
-from .comparam_subset import BaseComparam, Comparam, ComplexComparam, ComplexValue, read_complex_value_from_odx
 
 
 class CommunicationParameterRef:
 
     def __init__(self,
                  value : Union[str, ComplexValue],
                  id_ref: OdxLinkRef,
+                 is_functional = False,
```
andlaus, line 20: please add type annotations wherever possible (everywhere):
```suggestion
                 is_functional: bool = False,
```
also, where is this parameter comming from? (I cannot find it in the XSD for COMPARAM-REF...)

finally -- if this parameter is optional (i.e., not defining it means "I don't care" -- this should probably be
```suggestion
                 is_functional: Optional[bool] = None,
```

odxtools/communicationparameter.py
```diff
@@ -103,7 +106,7 @@ def short_name(self):
         # This should not happen anyway in a correct PDX
         return self.id_ref.ref_id.replace(".", "__").replace("-", "_")
 
-def read_communication_param_ref_from_odx(et_element, doc_frags: List[OdxDocFragment]):
+def read_communication_param_ref_from_odx(et_element, doc_frags: List[OdxDocFragment], dl_variant_type): # TODO argument type annotation
```
andlaus, line 109: please add the type annotation for `dl_variant_type` ;)

I suppose it is
```suggestion
def read_communication_param_ref_from_odx(et_element, doc_frags: List[OdxDocFragment], dl_variant_type: str):
```

odxtools/communicationparameter.py
```diff
@@ -116,6 +119,8 @@ def read_communication_param_ref_from_odx(et_element, doc_frags: List[OdxDocFrag
     else:
         value = read_complex_value_from_odx(et_element.find("COMPLEX-VALUE"))
 
+    is_functional = dl_variant_type == "FUNCTIONAL-GROUP"
```
andlaus, line 122: brackets make this more easily readable IMO:
```suggestion
    is_functional = (dl_variant_type == "FUNCTIONAL-GROUP")
```

odxtools/diaglayer.py
```diff
@@ -92,8 +93,8 @@ def get_inherited_data_object_properties_by_name(self):
                     if dop.short_name not in self.not_inherited_dops}
             return dops
 
-        def get_inherited_communication_parameters_by_name(self):
-            return {cp.id_ref.ref_id: cp for cp in self.parent_diag_layer._communication_parameters}
+        def get_inherited_communication_parameters(self):
+            return list(self.parent_diag_layer._communication_parameters)
```
andlaus, line 97: what's the reason for this change? possible collision between the keys?

odxtools/diaglayer.py
```diff
@@ -261,7 +262,7 @@ def _resolve_references(self, odxlinks: OdxLinkDatabase) -> None:
 
         self._communication_parameters = NamedItemList[CommunicationParameterRef](
             short_name_as_id,
-             list(self._compute_available_commmunication_parameters_by_name().values())
```
andlaus, line 264: ok, I see: the dict was actively harmful...

odxtools/diaglayer.py
```diff
@@ -462,12 +461,17 @@ def decode_response(
 
     def get_communication_parameter(self,
                                     name: str,
+                                    functional: Optional[bool] = None,
```
andlaus, line 464: please rename:
```suggestion
                                    is_functional: Optional[bool] = None,
```

odxtools/diaglayer.py
```diff
@@ -462,12 +461,17 @@ def decode_response(
...
                                     protocol_name: Optional[str] = None) \
             -> Optional[CommunicationParameterRef]:
 
         cps = [cp for cp in self.communication_parameters if cp.short_name == name]
+
+        if functional is not None:
+            cps = [cp for cp in cps if cp.is_functional == functional]
         if protocol_name:
-            cps = [cp for cp in cps if cp.protocol_sn_ref in (None, protocol_name)]
+            cps = [cp for cp in cps if cp.protocol_sn_ref == protocol_name]
```
andlaus, line 473: the idea of the `cp.protocol_sn_ref in (None, protocol_name)` statement was that if no protocol SNREF is specified by the comparam ref, the param always applies. maybe I misread the spec here, but if not, you probably need to add the same logic for the `is_functional` logic two lines above...

andlaus (commented):
looks fine in principle. I'll merge once my nitpicky questions/suggestions have been resolved...

odxtools/communicationparameter.py
dmholtz: The parameter is **not** coming from the XSD. The pdx files I have worked with so far group normal comparams in a DiagLayer with `variant_type="PROTOCOL"` whereas comparams for functional addressing are in a separate DiagLayer with `variant_type="FUNCTIONAL_GROUP"`. Without this change there is no way to distinguish between functional addressing non-functional addressing (which is IMO the more frequent use case).

odxtools/communicationparameter.py
dmholtz: Good catch! I would even go a step further and make this string an Enum type, since it also occurs at other places in the code.

odxtools/diaglayer.py
```diff
@@ -529,11 +533,12 @@ def get_can_func_req_id(self, protocol_name: Optional[str] = None) \
 
         return int(result)
 
-    def get_doip_logical_gateway_address(self, protocol_name: Optional[str] = None) \
+    def get_doip_logical_gateway_address(self, functional: Optional[bool] = False, protocol_name: Optional[str] = None) \
```
dmholtz, line 536: There is a lot of duplicate code, which I think is hard to maintain. I therefore propose the following change: We write a generic method `def get_doip_address(self, address_type : DoIPAddressType, is_functional : bool, protocol_name: str)`, where DoIPAddressType is a Enum with values `LogicalGatewayAddress, LogicalTesterAddress, LogicalEcuAddress, LogicalFunctionalAddress`.

odxtools/communicationparameter.py
andlaus: ah, okay. I don't really like the fact that after this, objects of the class will contain attributes that are specified by the DiagLayer, but I don't see any other convenient way out either. let's keep this construct, then.

odxtools/communicationparameter.py
andlaus: go ahead. I repeatedly fell on my nose when trying to use `Literal` for such things, but if you can make it work without too much hassle, I'm all for it...

odxtools/diaglayer.py
dmholtz: > ok, I see: the dict was actively harmful...

I believe, that the same problem applies to the `get_inherited_data_object_properties_by_name` and 'get_inherited_services_by_name`. Shouldn't we update these methods as well?

odxtools/diaglayer.py
dmholtz: I think you are right regarding the protoco, I will rollback that change. However, since `is_functional` is truely boolean, we do not need to adapt the logic above

odxtools/diaglayer.py
dmholtz: I absolutely agree that invoking such a function with an extra parameter is a bit too verbose. Using helper functions is probably the way to go. I will however, address this in a future PR where I plan to introduce more `get_comparam_XY` functions 😆

odxtools/diaglayer.py
andlaus: possibly: I'm not 100% about the semantics  specified by the standard, but I consider it likely that DOPs and services specified in derived diag layers override those of their parents. (for comparams  it is probably actually the same, but instead of using a simple name to decide if a comparam ought to be overriden, we need to use `(short_name, protocol, is_functional)`?)

odxtools/communicationparameter.py
dmholtz: The latest version implements a `DIAG_LAYER_TYPE` enum to distinguish the types of diag layers by a concrete type. The `__init__` method of the class `DiagLayer` accepts only enum instances, which ensures correctness-by-construction.

odxtools/communicationparameter.py
dmholtz: Done

odxtools/diaglayer.py
dmholtz: Done

dmholtz:
I'm done with applying your proposals on my PR and the PR is ready to merge from my POV. Please have a final look at the changes and let me know if something needs to be adapted.

odxtools/diaglayer.py
dmholtz: I'll keep this in mind in case I observe some strange behavior. Since everything has worked so far with the pdx files I have used, this topic is a bit too hot for me to address this in this PR right away.

odxtools/diaglayer.py
```diff
@@ -461,16 +462,16 @@ def decode_response(
...
+                                    is_functional: Optional[bool] = None,
                                     protocol_name: Optional[str] = None) \
             -> Optional[CommunicationParameterRef]:
 
         cps = [cp for cp in self.communication_parameters if cp.short_name == name]
 
-        if functional is not None:
-            cps = [cp for cp in cps if cp.is_functional == functional]
+        if is_functional is not None:
+            cps = [cp for cp in cps if cp.is_functional == is_functional]
```
andlaus, line 472:
```suggestion
            cps = [cp for cp in cps if cp.is_functional in (None, is_functional)]
```

(at least if the comparam does not need to specify the "FUNCTIONAL" attribute and if its absence means "don't care"...)

dmholtz: The way the `is_functional` attribute is set for a instance of type `CommunicationParameterRef`, it is currently not possible to have a `None` value at this attribute:

`is_functional = (dl_type == DIAG_LAYER_TYPE.FUNCTIONAL_GROUP)`

... except someone initializes these `CommunicationParameterRef` objects in the future in a different way.

andlaus: fair enough. thanks for the explaination

andlaus:
thanks for the contribution!

PR #102, opened 2022-12-16
Add method to query the CP_DoIPLogicalEcuAddress

dmholtz:
This is a small feature PR, which adds a method to query the Comparam for the DoIP logical ecu address.

```python
get_doip_logical_ecu_address(self, ...):
```

This comparam is crucial for establishing a DoIP connection with the vehicle. The desired data item is nested in a complex Comparam called `UniqueRespIdTable`.

### DIFF
--- a/odxtools/diaglayer.py
+++ b/odxtools/diaglayer.py
@@ -534,6 +534,25 @@ class DiagLayer:
 
         return int(result)
 
+    def get_doip_logical_ecu_address(self, doip_layer_name: str) -> Optional[int]:
+        """Return the CP_DoIPLogicalEcuAddress.
+        The parameter doip_layer_name is required to distinguish between different interfaces, 
+        e.g., offboard and onboard DoIP Ethernet.
+        """
+
+        com_param = self.get_communication_parameter("CP_UniqueRespIdTable", 
+                                                    protocol_name=doip_layer_name, 
+                                                    is_functional=False)
+
+        if com_param is None:
+            return None
+        
+        # The CP_DoIPLogicalEcuAddress is a subvalue of a complex Comparam called CP_UniqueRespIdTable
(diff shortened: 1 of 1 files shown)


kayoub5:
@andlaus I am pretty sure the list of standard communication parameters in ASAM standard is not short, maybe create class with properties instead of a function for each property?

andlaus:
@kayoub5: I suppose that would be a possibility. That said, as I see it currently, the main point of these methods is to make accessing and discovering the most commonly used of those parameters more convenient. Having a class which features all standardized parameters would IMO be counter-productive in this regard, because people would be flooded by exotic parameters. (at least if there was no way to highlight the more important ones.)

andlaus:
Maybe I misunderstood you: If you meant to say that these methods should be decorated with `@property`, this is unfortunately not (easily) possible AFAICS because they require the `protocol_name` and `is_functional` parameters...

dmholtz:
As far as I have understood, @kayoub5's point is to create a new class (beyond the scope of the PDX spec), which bundles all the comparams as properties. It is rather a technicality whether comparams are accessible as properties or with methods. The key idea is to to shrink the `DiagLayer` class.

From my POV, such a change only transfers the complexity of dealing with various comparams to a different place and even adds a tiny piece of extra overhead. I would rather appreciate a design which bundles comparams of the same protocol (e.g. CAN / DoIP). Typically, one requires the comparams for one protocol but not for multiple at the same time. Thereby, one could also avoid to repeatedly pass the same kwarg to the methods.

Admittedly, the public interface of the `DiagLayer` class is huge at the moment but its usability is still fine. Such a change would be a refactoring with low to medium priority IMO.

andlaus:
ok. I guess this is ready, in particular because it is fairly simple. Thanks for the contribution.

PR #103, opened 2022-12-19
Comparam fixes

andlaus:
Seems like the comparams are becoming a never-ending story. They got broken by a2c8e0ea: The problem with the simplified approach is that comparameters with the same name that are specified by low-priority parents are not properly overwritten by the higher-priority ones. This could be fixed by simply always taking the last instead of the first of the filtered comparams, but this approach would IMO not very explicit and it would also not get rid of the "multiple definition of comparam Foo" warnings. So let's revert back to building a dictionary as before a2c8e0ea, but we (a) now use a `(comparam.short_name, protocol_short_name)` tuple as the key and (b) treat the dictionary as an internal implementation detail of the function (i.e., we return a plain list).

Besides this, the protocol handling code is improved: It now returns a `NamedItemList` of diag layers instead of a list of short name strings like `Database.protocol_names` did, and we also have a `protocols` property for each individual diag layer.

### DIFF
--- a/odxtools/diaglayer.py
+++ b/odxtools/diaglayer.py
@@ -514,7 +514,7 @@ class DiagLayer:
             cps = [cp for cp in cps if cp.is_functional == is_functional]
         if protocol_name:
             cps = [cp for cp in cps if cp.protocol_sn_ref in (None, protocol_name)]
-        
+
         if len(cps) > 1:
             warnings.warn(f"Communication parameter `{name}` specified more "
                           f"than once. Using first occurence.", OdxWarning)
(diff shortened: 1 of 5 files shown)


odxtools/diaglayer.py
```diff
@@ -489,14 +531,16 @@ def get_can_receive_id(self, protocol_name: Optional[str] = None) \
         if com_param is None:
             return None
 
-        result = com_param.get_subvalue("CP_CanPhysReqId")
+        with warnings.catch_warnings():
+            warnings.simplefilter("ignore")
```
kayoub5, line 535: class to ignore?

kayoub5 (approved):

andlaus:
alright, let's merge. thanks for going over this, @kayoub5!

PR #104, opened 2022-12-20
use named constructors for constructing objects from XML where it makes sense

andlaus:
This PR is pretty large but quite boring (that's the best combination, isn't it?). These constructors mostly replace the `read_<foo>_from_odx()` functions and follow the convention `<Foo>.from_et(et_elem, doc_frags)`. The new approach has the advantages that much more of the code related to a given class is contained within the class' body and that only the class name (`Foo`) needs to be imported by the consumers.

Not all `read_<foo>_from_odx()` match well to be replaced by this pattern, though:

- Factory functions which create an object of a derived class but return it as an instance of the base class. These have been renamed to `create_any_<foo>_from_et()`, e.g., `create_any_data_object_property_from_et()`.
- Functions which do not return an object of a special-purpose class but e.g. a string.  Most prominently: `read_description_from_et()`
- Functions which return a list of a given object type, e.g., `create_sdgs_from_et()`

### DIFF
--- a/odxtools/admindata.py
+++ b/odxtools/admindata.py
@@ -24,6 +24,22 @@ class CompanyDocInfo:
     def team_member(self) -> Optional[TeamMember]:
         return self._team_member
 
+    @staticmethod
+    def from_et(et_element: ElementTree.Element ,
+                doc_frags: List[OdxDocFragment]) \
+            -> "CompanyDocInfo" :
+        # the company data reference is mandatory
+        company_data_ref = OdxLinkRef.from_et(et_element.find("COMPANY-DATA-REF"), doc_frags)
+        assert company_data_ref is not None
+        team_member_ref = OdxLinkRef.from_et(et_element.find("TEAM-MEMBER-REF"), doc_frags)
+        doc_label = et_element.findtext("DOC-LABEL")
+        sdgs = read_sdgs_from_odx(et_element.find("SDGS"), doc_frags)
+
+        return CompanyDocInfo(company_data_ref=company_data_ref,
+                              team_member_ref=team_member_ref,
+                              doc_label=doc_label,
+                              sdgs=sdgs)
+
     def _build_odxlinks(self) -> Dict[OdxLinkId, Any]:
         result = { }
 
@@ -46,6 +62,15 @@ class Modification:
     change: Optional[str] = None
     reason: Optional[str] = None
 
+    @staticmethod
+    def from_et(et_element: ElementTree.Element,
+                doc_frags: List[OdxDocFragment]) \
+            -> "Modification":
+        change = et_element.findtext("CHANGE")
+        reason = et_element.findtext("REASON")
+
+        return Modification(change=change,
+                            reason=reason)
 @dataclass()
 class DocRevision:
     """
@@ -59,6 +84,30 @@ class DocRevision:
     tool: Optional[str] = None
     modifications: List[Modification] = field(default_factory=list)
 
+    @staticmethod
+    def from_et(et_element: ElementTree.Element,
+                doc_frags: List[OdxDocFragment]) \
+            -> "DocRevision":
+
+        team_member_ref = OdxLinkRef.from_et(et_element.find("TEAM-MEMBER-REF"), doc_frags)
+        revision_label = et_element.findtext("REVISION-LABEL")
+        state = et_element.findtext("STATE")
+        date = et_element.findtext("DATE")
+        assert date is not None
+        tool = et_element.findtext("TOOL")
+
+        modlist = [
+            Modification.from_et(mod_elem, doc_frags)
+            for mod_elem in et_element.iterfind("MODIFICATIONS/MODIFICATION")
+        ]
+
+        return DocRevision(team_member_ref=team_member_ref,
+                           revision_label=revision_label,
+                           state=state,
+                           date=date,
+                           tool=tool,
+                           modifications=modlist)
+
     def _resolve_references(self, odxlinks: OdxLinkDatabase):
         if self.team_member_ref is not None:
             self.team_member = odxlinks.resolve(self.team_member_ref)
@@ -67,7 +116,31 @@ class DocRevision:
 class AdminData:
     language: Optional[str] = None
     company_doc_infos: List[CompanyDocInfo] = field(default_factory=list)
-    doc_revisions: Optional[List[DocRevision]] = None
+    doc_revisions: List[DocRevision] = field(default_factory=list)
+
+    @staticmethod
+    def from_et(et_element: Optional[ElementTree.Element],
+                doc_frags: List[OdxDocFragment]) \
+            -> Optional["AdminData"]:
+
+        if et_element is None:
+            return None
+
+        language = et_element.findtext("LANGUAGE")
+
+        company_doc_infos = [
+            CompanyDocInfo.from_et(cdi_elem, doc_frags)
+            for cdi_elem in et_element.iterfind("COMPANY-DOC-INFOS/COMPANY-DOC-INFO")
+        ]
+
+        doc_revisions = [
+            DocRevision.from_et(dr_elem, doc_frags)
+            for dr_elem in et_element.iterfind("DOC-REVISIONS/DOC-REVISION")
+        ]
+
+        return AdminData(language=language,
+                         company_doc_infos=company_doc_infos,
+                         doc_revisions=doc_revisions)
 
     def _build_odxlinks(self) -> Dict[OdxLinkId, Any]:
         result: Dict[OdxLinkId, Any] = {}
@@ -78,69 +151,8 @@ class AdminData:
         return result
 
     def _resolve_references(self, odxlinks: OdxLinkDatabase):
-        if self.company_doc_infos is not None:
-            for cdi in self.company_doc_infos:
-                cdi._resolve_references(odxlinks)
-
-        if self.doc_revisions is not None:
-            for dr in self.doc_revisions:
-                dr._resolve_references(odxlinks)
-
         for cdi in self.company_doc_infos:
             cdi._resolve_references(odxlinks)
 
-def read_admin_data_from_odx(et_element: Optional[ElementTree.Element],
-                             doc_frags: List[OdxDocFragment]) \
-        -> Optional[AdminData]:
-
-    if et_element is None:
-        return None
-
-    language = et_element.findtext("LANGUAGE")
-
-    company_doc_infos = list()
-    if cdis_elem := et_element.find("COMPANY-DOC-INFOS"):
-        for cdi in cdis_elem.iterfind("COMPANY-DOC-INFO"):
-            # the company data reference is mandatory
-            company_data_ref = OdxLinkRef.from_et(cdi.find("COMPANY-DATA-REF"), doc_frags)
-            assert company_data_ref is not None
-            team_member_ref = OdxLinkRef.from_et(cdi.find("TEAM-MEMBER-REF"), doc_frags)
-            doc_label = cdi.findtext("DOC-LABEL")
-            sdgs = read_sdgs_from_odx(cdi.find("SDGS"), doc_frags)
-
-            company_doc_infos.append(CompanyDocInfo(company_data_ref=company_data_ref,
-                                                    team_member_ref=team_member_ref,
-                                                    doc_label=doc_label,
-                                                    sdgs=sdgs))
-
-    doc_revisions = []
-    if drs_elem := et_element.find("DOC-REVISIONS"):
-        for dr in drs_elem.iterfind("DOC-REVISION"):
-            team_member_ref = OdxLinkRef.from_et(dr.find("TEAM-MEMBER-REF"), doc_frags)
-            revision_label = dr.findtext("REVISION-LABEL")
-            state = dr.findtext("STATE")
-            date = dr.findtext("DATE")
-            assert date is not None
-            tool = dr.findtext("TOOL")
-
-            modlist = None
-            mods = dr.find("MODIFICATIONS")
-            if mods is not None:
-                modlist = list()
-                for mod in mods.iterfind("MODIFICATION"):
-                    m_change = mod.findtext("CHANGE")
-                    m_reason = mod.findtext("REASON")
-
-                    modlist.append(Modification(change=m_change,
-                                                reason=m_reason))
-
-            doc_revisions.append(DocRevision(team_member_ref=team_member_ref,
-                                      revision_label=revision_label,
-                                      state=state,
-                                      date=date,
-                                      tool=tool,
-                                      modifications=modlist))
-
-    return AdminData(language=language,
-                     company_doc_infos=company_doc_infos,
-                     doc_revisions=doc_revisions)
+        for dr in self.doc_revisions:
+            dr._resolve_references(odxlinks)
--- a/odxtools/audience.py
+++ b/odxtools/audience.py
@@ -19,6 +19,31 @@ class Audience:
     _enabled_audiences: Optional[List["Audience"]] = None
     _disabled_audiences: Optional[List["Audience"]] = None
 
+    @staticmethod
+    def from_et(et_element, doc_frags: List[OdxDocFragment]) \
+            -> "Audience":
+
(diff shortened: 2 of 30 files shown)


odxtools/admindata.py
```diff
@@ -46,28 +66,133 @@ class Modification:
...
+@dataclass()
+class CompanyRevisionInfo:
+    company_data_ref: OdxLinkRef
+    revision_label: Optional[str] = None
+    state: Optional[str] = None
+
+    _company_data: Optional[CompanyData] = None
+    @property
+    def company_data(self) -> CompanyData:
+        assert self._company_data is not None
```
kayoub5, line 88: Avoid throwing in getters, user of API always assume that getters have no logic

See also https://stackoverflow.com/questions/48778945/by-design-should-a-property-getter-ever-throw-an-exception-in-python

odxtools/admindata.py
```diff
@@ -46,28 +66,133 @@ class Modification:
...
+        return self._company_data
+
+    @staticmethod
+    def from_et(et_element: ElementTree.Element,
+                doc_frags: List[OdxDocFragment]) \
+            -> "CompanyRevisionInfo":
+
+        company_data_ref = OdxLinkRef.from_et(et_element.find("COMPANY-DATA-REF"),
+                                              doc_frags)
+        assert company_data_ref is not None
```
kayoub5, line 98: required in XSD?

odxtools/audience.py
```diff
@@ -19,6 +19,31 @@ class Audience:
...
+            for ref in et_element.iterfind("ENABLED-AUDIENCE-REFS/"
+                                           "ENABLED-AUDIENCE-REF")]
+        disabled_audience_refs = [OdxLinkRef.from_et(ref, doc_frags)
+            for ref in et_element.iterfind("DISABLED-AUDIENCE-REFS/"
+                                           "DISABLED-AUDIENCE-REF")]
+        is_supplier = et_element.get("IS-SUPPLIER", "true") == 'true'
+        is_development = et_element.get("IS-DEVELOPMENT", "true") == 'true'
+        is_manufacturing = et_element.get("IS-MANUFACTURING", "true") == 'true'
+        is_aftersales = et_element.get("IS-AFTERSALES", "true") == 'true'
+        is_aftermarket = et_element.get("IS-AFTERMARKET", "true") == 'true'
```
kayoub5, line 36: Sure default value for all of those is true?

odxtools/compumethods/makeanycompumethod.py
```diff
@@ -79,37 +77,10 @@ def _parse_compu_scale_to_linear_compu_method(scale_element,
...
-                limit = Limit(internal_type.from_string(et_element.text),
-                              interval_type)
-    return limit
-
-
-def read_compu_method_from_odx(et_element,
-                               doc_frags: List[OdxDocFragment],
-                               internal_type: DataType,
-                               physical_type: DataType) -> CompuMethod:
+def make_any_compu_method_from_et(et_element,
```
kayoub5, line 80: "any" sounds redundant, even without it the function role is pretty clear.
```suggestion
def make_compu_method_from_et(et_element,
```

odxtools/singleecujob.py
```diff
@@ -60,6 +84,32 @@ class OutputParam:
...
+    @staticmethod
+    def from_et(et_element, doc_frags: List[OdxDocFragment]) \
+            -> "OutputParam":
+
+        odx_id = OdxLinkId.from_et(et_element, doc_frags)
+        assert odx_id is not None
+        short_name = et_element.findtext("SHORT-NAME")
+        assert short_name is not None
+        long_name = et_element.findtext("LONG-NAME")
+        description = read_description_from_et(et_element.find("DESC"))
```
kayoub5, line 96: description is always in a DESC tag, maybe move that part of the code to the common function?
```suggestion
        description = read_description_from_et(et_element)
```

odxtools/specialdata.py
```diff
@@ -122,7 +123,7 @@ def _resolve_references(self, odxlinks: OdxLinkDatabase) -> None:
         for val in self.values:
             val._resolve_references(odxlinks)
 
-def read_sdgs_from_odx(et_element: Optional[ElementTree.Element],
+def create_sdgs_from_et(et_element: Optional[ElementTree.Element],
```
kayoub5, line 126: sometimes you use prefix "create", sometimes "read", and others "make", it would be better to normalize all to use same prefix.

odxtools/table.py
```diff
@@ -56,6 +56,38 @@ def __post_init__(self) -> None:
         self._structure: Optional[DopBase] = None
         self._dop: Optional[DopBase] = None
 
+    @staticmethod
+    def from_et(et_element, doc_frags: List[OdxDocFragment]) \
+            -> "TableRow":
+        """Reads a TABLE-ROW."""
+        odx_id=OdxLinkId.from_et(et_element, doc_frags)
+        assert odx_id is not None
+        short_name=et_element.findtext("SHORT-NAME")
```
kayoub5, line 65: You need a code style linter
```suggestion
        short_name = et_element.findtext("SHORT-NAME")
```

odxtools/utils.py
```diff
@@ -6,7 +6,7 @@
 
 from .odxlink import OdxDocFragment
 
-def read_description_from_odx(et_element: Optional[ElementTree.Element]) \
+def read_description_from_et(et_element: Optional[ElementTree.Element]) \
```
kayoub5, line 9: Can you split automatic simple updates like changing a function name into their own PR?

That way reviewing the rest of the logic won't be so difficult

odxtools/admindata.py
andlaus: hm, the problem with this is that this would make the type of the return value `Optional[T]` thus making mypy happy would become pretty awkward for mandatory items that are defined by reference. At this particular location, the `company_data` is supposed to be optional anyway because the reference can be `None` as well. That said, I still think that raising an `AssertationError` is okay to do even if it's not expected by the user because this represents an error in the program logic to call such properties before the references are resolved. (In e.g. C++, this usually leads to an `abort()` or a segfault...)

odxtools/admindata.py
andlaus: no. thanks for the catch!

odxtools/admindata.py
andlaus: ok, scrap my last comment: for COMPANY-REVISION-INFO the COMPANY-DATA-REF tag *is* indeed mandatory. what to do about it?

odxtools/admindata.py
andlaus: ok, I misread the class which this was about. for COMPANY-REVISION-INFO the COMPANY-DATA-REF tag is indeed mandatory...

odxtools/admindata.py
kayoub5: if you must throw, at least throw something that could be caught (one of the exception classes defined in odxtools) and add an error message

odxtools/audience.py
```diff
@@ -19,6 +19,31 @@ class Audience:
...
+    def from_et(et_element, doc_frags: List[OdxDocFragment]) \
+            -> "Audience":
+
+        enabled_audience_refs = [OdxLinkRef.from_et(ref, doc_frags)
+            for ref in et_element.iterfind("ENABLED-AUDIENCE-REFS/"
+                                           "ENABLED-AUDIENCE-REF")]
+        disabled_audience_refs = [OdxLinkRef.from_et(ref, doc_frags)
+            for ref in et_element.iterfind("DISABLED-AUDIENCE-REFS/"
+                                           "DISABLED-AUDIENCE-REF")]
+        is_supplier = et_element.get("IS-SUPPLIER", "true") == 'true'
```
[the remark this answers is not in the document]
andlaus: good point. I'll introduce an `str_to_bool()' utility function. this is a bit elaborate, I hope to have get it done by tomorrow. stay tuned...

odxtools/audience.py
andlaus: yes:
Screenshot_20221220_215656

the problem is that these attributes are also optional, so if we want to be able to distinguish "not specified", from "default" we have to interpret `None` as `True`. I've got a patch in the works for this...

odxtools/singleecujob.py
```diff
@@ -163,6 +258,72 @@ def __post_init__(self) -> None:
         if not self.neg_output_params:
             self.neg_output_params = NamedItemList(short_name_as_id, [])
 
+    @staticmethod
+    def from_et(et_element, doc_frags: List[OdxDocFragment]) \
```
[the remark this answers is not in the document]
andlaus: for the first argument? yes, but if this ought to be done for all `_from_et` functions, making mypy happy again will lead to another giant patch. This will be one of my next projects...

odxtools/specialdata.py
andlaus: I agree. what do you prefer?

odxtools/table.py
```diff
@@ -56,6 +56,38 @@ def __post_init__(self) -> None:
         self._structure: Optional[DopBase] = None
         self._dop: Optional[DopBase] = None
 
+    @staticmethod
+    def from_et(et_element, doc_frags: List[OdxDocFragment]) \
```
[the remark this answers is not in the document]
andlaus: yes, see above: I wanted to spare you the changes required to make mypy happy. In this PR, the new constructors ought to be as close to the old ones as possible...

odxtools/table.py
andlaus: I don't mind. but let's not mingle these changes with functional ones or this PR will become even more unreviewable than it already is, i.e., let's do a separate giant PR which just puts the whole code base through one of the linters (black?)

odxtools/templates/macros/printAdminData.xml.jinja2
```diff
@@ -44,6 +44,21 @@
    {%- if doc_revision.tool is not none %}
    <TOOL>{{doc_revision.tool|e}}</TOOL>
    {%- endif %}
+   {%- if doc_revision.company_revision_infos %}
+   <COMPANY-REVISION-INFOS>
```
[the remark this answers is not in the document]
andlaus: I'll try. This is quite a bit of effort though, because these "drive by" fixes like here can be non-trivial and hard to "unentangle" from the other changes...

odxtools/utils.py
andlaus: that can be done

odxtools/compumethods/makeanycompumethod.py
andlaus: hm, I want to somehow signify that this is a factory method...

odxtools/singleecujob.py
andlaus: I also thought about that, but decided against it because everything else takes the element for the tag which describes the object...

odxtools/admindata.py
andlaus: ok. separate PR?!

andlaus: BTW: I think that it is possible to catch `AssertationError` and I also think that this exception is appropriate at this location because it signifies a programming error by the user. (For most of the asserts this is not really the case, in particular not those who's primary purpose is to satisfy mypy...)

odxtools/specialdata.py
kayoub5: All seem fine, my opinion on naming:
* "read_" prefix makes it obvious about intent, so could possibly remove the "_from_et" suffix in that case
* The "create_" prefix is a more popular choice than "make_" in Python https://docs.python.org/3/search.html?q=create_

andlaus: let's go for `create_<foo>_from_et` then: I'm not really fond of `read_` because these functions do not read in anything from storage or the network (`parse_` would probably be more appropriate IMO) and I like to keep the `_from_et` suffix because python does not really support or encourage C++-style function overloading...

odxtools/table.py
kayoub5: sounds good

odxtools/specialdata.py
dmholtz: You could simplify the suffix to `_from`. This makes the the entire signature `create_<foo>_from(et_element)` well readable.

odxtools/table.py
dmholtz: Black is a nice formatter, which is widely used in OS projects. The major drawback is, that it strictly limits lines to roughly 90 characters. Especially with the verbose typing annotations, this will result in some cumbersome method signatures or variable declarations.

I agree, this should be part of a different PR, since black would probably reformat the entire codebase and the functional changes would be untraceable.

odxtools/specialdata.py
andlaus: IMO this is a *very* bad idea in python because python is not typed and such a method name could make the impression that the object to be created can be made from anything. In c++ this is more feasible because the compiler will complain about stuff that does not work...

odxtools/admindata.py
```diff
@@ -14,16 +14,36 @@
...
+    @property
+    def company_data(self) -> Optional[CompanyData]:
+        return self._company_data
+
     _team_member: Optional[TeamMember] = None
     @property
     def team_member(self) -> Optional[TeamMember]:
         return self._team_member
 
+    @staticmethod
```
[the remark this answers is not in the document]
andlaus: > This is a very general comment about Python style: As far as I know, it's more common to use @classmethod instead of a @static_method for factory methods (read more: [When to use the class or static method?](https://www.geeksforgeeks.org/class-method-vs-static-method-python/))

as far as I can see, there is an issue with classmethods that applies here: the constructors for the  "concrete" derived classes usually need a different set of arguments. Also, the main issue which the `create_foo_from_et()` methods aim to solve is that it is not clear which of the derived classes ought to be instantiated without looking at the ElementTree, so it is not easily doable to call `DerivedClass.from_et()` for the right `DerivedClass`. It's very well possible that I'm missing something, though...

If you are only talking about the "named constructor" methods for concrete objects: You're correct, but what is the point if these methods will not work for derived classes anyway?

odxtools/specialdata.py
dmholtz: Fair, I see that point. No errors at compile time requires additional attention by the programmers. Nonetheless, I don't believe that `_et` contains a lot information about the actual expected type if you are not 100% familiar with all the acronyms used in the project. In doubt, one has to read the documentation in either case. 

All of this is less critical in practice, since good IDEs (such as VS Code) show previews of method signatures and thereby indicate the parameter type, which is known from the typing annotation.

In the end, this is a discussion about three characters ;-) I definitely like the overall naming convention `create_<foo>_from[_et]` and am fine with both ways.

odxtools/admindata.py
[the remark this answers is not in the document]
dmholtz: > If you are only talking about the "named constructor" methods for concrete objects: You're correct, but what is the point if these methods will not work for derived classes anyway?

You are right, there is no issue if the methods do not work for derived classed. I didn't mean to exploit the inheritance property of these methods, since the set of parameters might differ between the base and the derived classes. The comment was solely refering to the heuristic how factory methods could be annotated. The caller should not notice a difference.

andlaus:
okay. it was quite a bit of work, but the latest version of this PR is purely concerned with `ClassName.from_et()`  things. (modulo stuff which I missed.) the remaining stuff will require three or four additional PRs...

odxtools/admindata.py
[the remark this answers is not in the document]
andlaus: let's stick with `@staticmethod` for the mean time: I like the fact that yo have to mention the name of the class which is to be instantiated explicitly, a la
```python
@staticmethod
def make_foo(arg1, arg2):
  return Foo(arg1, arg2)
```

instead of 

```python
@classmethod
def make_foo(cls, arg1, arg2):
  return cls(arg1, arg2)
```

odxtools/specialdata.py
andlaus: > [...] which is known from the typing annotation.

not currently ( in the case of the first parameter for most the constructor functions). PR is in the pipeline, though...

kayoub5 (approved):

andlaus:
let's merge this. thanks for having a look @kayoub5!

PR #107, opened 2022-12-23
fix handling of boolean attributes

andlaus:
In ODX, a true value can be specified by either the strings "true" or "1", and a false one by the strings "false" or "0". Also, leading and trailing white space is to be ignored. All of these complications are now dealt with by the new and shiny `str_to_bool()` function. (Suggestions for a better name are appreciated!)

Further, if an attribute of the XML is optional but specified to default to true, `None` must be interpreted as true, which is done by assigning the raw value to a protected attribute `._<foo>` while also defining a property `foo` that returns a boolean...

### DIFF
--- a/odxtools/audience.py
+++ b/odxtools/audience.py
@@ -3,7 +3,8 @@
 
 from dataclasses import dataclass, field
 from typing import Optional, List, Dict, Any
-from .utils import create_description_from_et, str_to_bool
+from .utils import create_description_from_et
+from .odxtypes import odxstr_to_bool
 from .odxlink import OdxLinkRef, OdxLinkId, OdxDocFragment, OdxLinkDatabase
 
 @dataclass()
@@ -11,30 +12,30 @@ class Audience:
     enabled_audience_refs: list = field(default_factory=list)
     disabled_audience_refs: list = field(default_factory=list)
 
-    _is_supplier: Optional[bool] = None
+    is_supplier_raw: Optional[bool] = None
     @property
     def is_supplier(self) -> bool:
-        return self._is_supplier in [None, True]
+        return self.is_supplier_raw in [None, True]
 
-    _is_development: Optional[bool] = None
+    is_development_raw: Optional[bool] = None
     @property
     def is_development(self) -> bool:
-        return self._is_development in [None, True]
+        return self.is_development_raw in [None, True]
 
-    _is_manufacturing: Optional[bool] = None
+    is_manufacturing_raw: Optional[bool] = None
     @property
     def is_manufacturing(self) -> bool:
-        return self._is_manufacturing in [None, True]
+        return self.is_manufacturing_raw in [None, True]
 
-    _is_aftersales: Optional[bool] = None
+    is_aftersales_raw: Optional[bool] = None
     @property
     def is_aftersales(self) -> bool:
-        return self._is_aftersales in [None, True]
+        return self.is_aftersales_raw in [None, True]
 
-    _is_aftermarket: Optional[bool] = None
+    is_aftermarket_raw: Optional[bool] = None
     @property
     def is_aftermarket(self) -> bool:
-        return self._is_aftermarket in [None, True]
+        return self.is_aftermarket_raw in [None, True]
 
     _enabled_audiences: Optional[List["AdditionalAudience"]] = None
     @property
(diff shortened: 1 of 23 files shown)


odxtools/audience.py
```diff
@@ -29,28 +58,20 @@ def from_et(et_element, doc_frags: List[OdxDocFragment]) \
...
 
         return Audience(enabled_audience_refs=enabled_audience_refs,
                         disabled_audience_refs=disabled_audience_refs,
-                        is_supplier=is_supplier,
-                        is_development=is_development,
-                        is_manufacturing=is_manufacturing,
-                        is_aftersales=is_aftersales,
-                        is_aftermarket=is_aftermarket)
-
+                        _is_supplier=is_supplier,
```
kayoub5, line 69: don't add underscore to function arguments

odxtools/utils.py
```diff
@@ -1,11 +1,29 @@
...
 
 from .odxlink import OdxDocFragment
 
+@overload
+def str_to_bool(str_val: None) -> None: ...
+
+@overload
+def str_to_bool(str_val: str) -> bool: ...
+
+def str_to_bool(str_val: Union[None, str]) -> Union[None, bool]:
```
kayoub5, line 15: simply use Optional instead of union?

odxtools/utils.py
```diff
@@ -1,11 +1,29 @@
...
+    if str_val is None:
+        return None
+
+    str_val = str_val.strip()
+    assert str_val in ["0", "1", "false", "true"]
```
kayoub5, line 20: add message to explain

kayoub5: or even better, use exception from the ones already defined

odxtools/audience.py
andlaus: yeah, I also don't like it, but I did not have a better idea: the problem is that I want to keep the `.is_supplier` attribute for users, but we cannot directly use the value specified by the XML because if the attribute does not get specified, it ought to be interpreted as `True` whilst python interprets `None` as false. maybe use another prefix for the "raw" values for such cases? which one?

odxtools/templates/macros/printAudience.xml.jinja2
```diff
@@ -19,32 +19,22 @@
 {%- endmacro -%}
 
 {%- macro printAudience(audience) -%}
-<AUDIENCE 
- {%- filter odxtools_collapse_xml_attribute %}
-  {%- if not audience.is_supplier %}
-                 IS-SUPPLIER="false"
+<AUDIENCE
+  {%- if audience._is_supplier is not none %}
+                 IS-SUPPLIER="{{bool_to_str(audience._is_supplier)}}"
```
kayoub5, line 24: just a suggestion, maybe make a function that also handles None, and only print attribute if not None
```
print_bool_attr("IS-SUPPLIER", audience._is_supplier) 
```

odxtools/audience.py
kayoub5: I am saying don't add underscore to argument name in constructor, use underscore only for the private attribute. the rest of the logic is fine

odxtools/utils.py
andlaus: yep, thanks for the catch. that is an development artifact because I had mypy complaining here at some point, so I thought that it had an issue with `Optional`. (It turned out that the issue was something completely different...)

odxtools/templates/macros/printAudience.xml.jinja2
andlaus: done.

odxtools/audience.py
andlaus: yeah, the issue is that I consider these attributes to be private (i.e., not meant to be used by "normal" users), but for dataclasses you need use the same name to pass them on construction time...

kayoub5: you added logic to the class creation, they are no longer "data classes", thus needs to be converted to normal classes.

andlaus: while that's true in the strict sense, one can also look at these classes as primarily for storing data. (also, it is much more convenient and less error prone having to type these attributes only once...)

kayoub5: having a function argument starts with underscore is unheard of in python, it makes it look like a private argument that's not supposed to be used publicly.

andlaus: alright. I changed this to using `_raw` suffixes. I don't really like this either, but I think it is not worse than using an underscore prefix...

kayoub5 (approved):

PR #108, opened 2023-01-02
Add missing kwargs to several parameter classes

phorkyas-tg:
In createanyparameter.py the LengthKeyParameter, SystemParameter, DynamicParameter and TableEntryParameter can't be initialised because of a missing keyword argument (sdgs). This can be fixed by adding the kwargs argument.

PR #109, opened 2023-01-03
Fixes

andlaus:
this PR contains a bunch of small but relevant fixes for things which I recently stumbled over.


odxtools/comparam_subset.py
```diff
@@ -115,11 +115,10 @@ def _build_odxlinks(self) -> Dict[OdxLinkId, Any]:
 class Comparam(BaseComparam):
     dop_ref: OdxLinkRef
     physical_default_value: Optional[str] = field(default=None, init=False)
-    _dop: Optional[DataObjectProperty] = field(default=None, init=False)
 
     @staticmethod
-    def from_et(et_element, doc_frags: List[OdxDocFragment]) -> ComplexComparam:
-        result = Comparam.__new__(ComplexComparam)
+    def from_et(et_element, doc_frags: List[OdxDocFragment]) -> "Comparam":
+        result = Comparam.__new__(Comparam)
```
kayoub5, line 121: why not just `result = Comparam()`?

odxtools/odxlink.py
```diff
@@ -60,6 +60,9 @@ def __eq__(self, other: Any) -> bool:
         # i.e. the same OdxId object can be put into all of them.
         return self.local_id == other.local_id
 
+    def __str__(self):
+        return f"OdxLinkId('{self.local_id}')"
```
kayoub5, line 64: include doc_name if present

odxtools/odxlink.py
```diff
@@ -130,6 +133,9 @@ def from_id(odxid: OdxLinkId) -> "OdxLinkRef":
         """
         return OdxLinkRef(odxid.local_id, odxid.doc_fragments)
 
+    def __str__(self):
+        return f"OdxLinkRef('{self.ref_id}')"
```
kayoub5, line 137: include doc name if present

odxtools/comparam_subset.py
andlaus: because then the "offical" constructor would need to be called and for this I need all attributes stored by the `Comparam` as well as the `BaseComparam` classes. (the problem with the `from_et()` idiom is that it does not work very well with inheritance, thus the split of the instantiation of such classes into object creation (`XYZ.__new__()`) and their actual initialization (`__init_from_et__()`.) If you know a better solution, I'd be happy to give it a shot...

odxtools/odxlink.py
andlaus: we only have a list of document fragments which do not necessarily correspond to the file name of the XML document in the zip file. The idea here is to give people enough to `grep` for the ID in an unpacked PDX while getting a unique result for ~99% of all cases and at the same time not cluttering up the output too much...

odxtools/comparam_subset.py
kayoub5: since it's existing code, I think it would suffice to explain the magic behind __new__ usage in a comment

andlaus: It's still pretty new code (#104 from last December). I suppose I should have put this into a separate PR instead of burying it in #104. (When I separated the original monster PR for that, I forgot about that detail. Sorry for that...)

andlaus: BTW: In the medium term, I intend to refactor the class hierarchy a bit so that it resembles the XSD more closely. this means that this pattern will see quite a bit more use as the XSD makes quite a bit of use of inheritance...

andlaus: ok, I've added some comments. I hope that I don't forget to add them to future occurrences of this pattern...

andlaus:
alright. I suppose this should go in even if you did not approve it (yet?). I think we should do a release (mainly because of #108), and I would like these fixes to be part of it...

PR #110, opened 2023-01-05
Refactor GitHub Action test-and-lint workflow

dmholtz:
The refactoring introduces a matrix workflow with 2 DOF: OS and Python version. We avoid setting environment variables to make the workflow run on both Windows and Ubuntu in the same manner.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -16,6 +16,12 @@ jobs:
       matrix:
         os: [ubuntu-latest, windows-latest]
         python-version: ["3.8", "3.10"]
+        
+        # due to the slow windows runners, we refrain from testing every python
+        # version on windows-latest
+        exclude:
+          - os: windows-latest
+            python-version: "3.8"
 
     steps:
       - uses: actions/checkout@v3
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -1,28 +1,57 @@
 name: Run Tests
 
-on: [push, pull_request, workflow_dispatch]
+on: 
+  push:
+    branches: [main]
+  pull_request:
+    branches: [main]
+  workflow_dispatch:
 
 jobs:
   unittest-and-lint:
-    runs-on: ubuntu-latest
+
+    runs-on: ${{matrix.os}}
+
+    strategy:
+      matrix:
+        os: [ubuntu-latest, windows-latest]
+        python-version: ["3.8", "3.10"]
+
     steps:
-    - uses: actions/checkout@v1
-    - name: Install dependencies
-      run: |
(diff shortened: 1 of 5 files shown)


.github/workflows/tests.yml
```diff
@@ -1,28 +1,57 @@
...
+      - name: Install odxtools dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install -r requirements.txt
+      
+      # we need additional dependencies for testing and the type checker
+      # but we avoided to specify as hard dependencies within 'requirements.txt'
+      - name: Install test dependencies
+        run:  |
+          pip install types-Deprecated coveralls mypy pytest pytest-cov
```
kayoub5, line 37: put them in a test-requirements.txt file

.github/workflows/tests.yml
```diff
@@ -1,28 +1,57 @@
...
+      - name: Create somersault.pdx
+        run: |
+          python -m examples.mksomersaultpdx examples/somersault.pdx
+
+      - name: Run unit tests and report line coverage
+        # use `python -m pytest ...` instead of `pytest ...` to avoid setting the
+        # PYTHONPATH environment variable. This works independently of the OS.
+
+        # For some reason, pytest-cov reports invalid line coverage. As a remedy,
+        # we utilize the `coverage` module instead of `pytest-cov`. Line coverage
```
kayoub5, line 48: * then why is pytest-cov in the dependency list if not used?
* sure this issue happens in latest version? pytest-cov uses the coverage module internally

.github/workflows/tests.yml
```diff
@@ -1,28 +1,57 @@
...
 jobs:
   unittest-and-lint:
-    runs-on: ubuntu-latest
+
+    runs-on: ${{matrix.os}}
+
+    strategy:
+      matrix:
+        os: [ubuntu-latest, windows-latest]
+        python-version: ["3.8", "3.10"]
```
kayoub5, line 18: python 3.11 has been released

.github/workflows/tests.yml
dmholtz: Unfortunately, yes. `pytest-cov` reports only 36 % coverage in contrast to 65 % reported by `coverage`. For some reason, `pytest-cov` marks every statement on module level (i.e. imports, method signatures, constants) as uncovered, which seems wrong from my point of view.

.github/workflows/tests.yml
dmholtz: Fair, I would suggest testing Python 3.8 and the latest version is then sufficient.

andlaus: I agree

andlaus:
merging. thanks for the review @kayoub5.

PR #111, opened 2023-01-05
use keyword=value to pass constuctor arguments when creating most types of objects

andlaus:
in this context, "most types of objects" basically means everything which is essentially a dataclass. this makes constructing them more explicit and less fragile.

note that for `@dataclass`es themselves, we currently cannot enforce this policy strictly because this would require to mark these classes with `kw_only=True` which is only available for python >= 3.10...

### DIFF
--- a/odxtools/admindata.py
+++ b/odxtools/admindata.py
@@ -11,7 +11,7 @@ from xml.etree import ElementTree
 from dataclasses import dataclass, field
 from typing import Optional, Any, Dict, List
 
-@dataclass()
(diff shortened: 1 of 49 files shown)


kayoub5:
@andlaus changes look ok, but you may want to increment major version number, some may consider this a breaking change.

andlaus:
ok, will do. I'd like to get a few other breaking changes in before I'll do the next release, though. (no defaults for optional arguments in the python code anymore.)

PR #113, opened 2023-01-09
Feature/ECU-VARIANT-PATTERNS

dmholtz:
This PR implements `ECU-VARIANT-PATTERNS` according to ISO 22901-1. These patterns define expected identification data of ECU variants, which enables dynamic discovery of ECUs.

The PR also implements the corresponding ecu variant matching algorithm.


odxtools/ecu_variant_patterns.py
```diff
@@ -0,0 +1,72 @@
...
+def create_ecu_variant_patterns_from_et(et_element: Optional[ElementTree.Element], doc_frags: List[OdxDocFragment]) -> List[EcuVariantPattern]:
+
+    if not et_element:
+        return []
+
+    result = []
+    for evp_elem in et_element.iterfind("ECU-VARIANT-PATTERN"):
+        result.append(EcuVariantPattern.from_et(evp_elem, doc_frags))
+
+    return result
```
andlaus, line 72: list comprehension makes this slightly shorter:
```suggestion
    return [
        EcuVariantPattern.from_et(evp_elem, doc_frags)
        for evp_elem in et_element.iterfind("ECU-VARIANT-PATTERN")
    ]
```

the same applies for the other instances where you create lists (basically everywhere where you use `iterfind()`) in this PR...

tests/test_ecu_variant_matching.py
```diff
@@ -0,0 +1,285 @@
...
+from odxtools.odxlink import OdxDocFragment, OdxLinkDatabase, OdxLinkId
+from odxtools.service import DiagService
+from odxtools.structures import Request, Response
+
+doc_frags = [OdxDocFragment("pytest", "WinneThePoh")]
+
+odxlinks = OdxLinkDatabase()
+
+
+@pytest.fixture
```
andlaus, line 22: Note to myself: Start replacing `unittest` by `pytest`...

tests/test_ecu_variant_matching.py
```diff
@@ -0,0 +1,285 @@
...
+        msg_str = message.decode(encoding="utf-8")
+        msg_dict = json.loads(msg_str)
+        return msg_dict
+
+    monkeypatch.setattr(resp, "decode", decode)
+    return resp
+
+
+@pytest.fixture()
+def identificationService(monkeypatch, dummy_response):
```
andlaus, line 41: python normally uses snake_case for function names:
```suggestion
def identification_service(monkeypatch, dummy_response):
```

(The python convention of using CamelCase for class names is a bit inconsistent IMO, but.... Also, I left out the other instances of camelCase as an exercise for the reader.)

Further, type annotations would be great here...

tests/test_ecu_variant_matching.py
```diff
@@ -0,0 +1,285 @@
...
+        resp = req_resp_mapping[req]
+        matcher.evaluate(resp)
+    assert not matcher.has_match()
+    with pytest.raises(AssertionError):
+        matcher.get_active_ecu_variant()
+
+
+@pytest.mark.parametrize("use_cache", [True, False])
+# test if pending matchers reject the has_match() or active variant query
+def test_no_request_loop(ecu_variant_1, ecu_variant_2, ecu_variant_3, use_cache):
```
andlaus, line 251:
```suggestion
def test_no_request_loop(ecu_variant_candidates : List[DiagLayer], use_cache: bool) -> None:
```

(same for the other methods with three hardcoded variant arguments.)

andlaus (changes requested):
looks pretty good!

odxtools/ecu_variant_matching.py
```diff
@@ -0,0 +1,141 @@
...
+        for ecu in self.ecus:
+            assert ecu.variant_type == DIAG_LAYER_TYPE.ECU_VARIANT
+
+        self.use_cache = use_cache
+        self.req_resp_cache: Dict[ByteString, Union[bytes, bytearray]] = {}
+        self._recent_ident_response: Optional[Union[bytes, bytearray]] = None
+
+        self._state = EcuVariantMatcher.State.PENDING
+
+    def request_loop(self):
```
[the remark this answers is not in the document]
dmholtz: Since the `request_loop()` is a Python generator (due to `yield`), the annotation is a bit more complicated, but I will give it a try.

odxtools/ecu_variant_matching.py
```diff
@@ -0,0 +1,141 @@
...
+        """
+        return self._state == EcuVariantMatcher.State.PENDING
+
+    def has_match(self) -> bool:
+        """Returns true iff the non-pending matcher found a matching ecu variant.
+        
+        Raises a runtime error if the matcher is pending.
+        """
+        if self.is_pending():
+            raise RuntimeError("EcuVariantMatcher is pending. Run the request_loop to determine the active ecu variant.")
```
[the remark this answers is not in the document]
dmholtz: I don't want the user to call any other method than `evaluate` within the loop body. The `request_loop` is a generator and implements the matching algorithm. At any time within the loop, the matching state is not determined and I want the user to know that. In particular, I prefer boolean semantics of a `has_XY` method.

The use case you are referring to can be realized with the `is_pending()`, which is again truely boolean.

odxtools/ecu_variant_patterns.py
```diff
@@ -0,0 +1,72 @@
...
+        for mp_el in mp_collection_el.iterfind("MATCHING-PARAMETER"):
+            matching_params.append(MatchingParameter.from_et(mp_el, doc_frags))
+
+        assert len(matching_params) > 0 # required by ISO 22901-1 Figure 141
+        return EcuVariantPattern(matching_params)
+
+
+def create_ecu_variant_patterns_from_et(et_element: Optional[ElementTree.Element], doc_frags: List[OdxDocFragment]) -> List[EcuVariantPattern]:
+
+    if not et_element:
```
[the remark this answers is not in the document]
dmholtz: Thanks, that's a very good hint!

odxtools/ecu_variant_patterns.py
dmholtz: agree, comprehensions are nice :)

odxtools/ecu_variant_matching.py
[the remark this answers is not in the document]
andlaus: `typing.Generator` should do the trick!?

https://docs.python.org/3/library/typing.html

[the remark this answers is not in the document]
andlaus: also, above snippet should also work:
```
Alternatively, annotate your generator as having a return type of either Iterable[YieldType] or Iterator[YieldType]

[...]
```

odxtools/ecu_variant_matching.py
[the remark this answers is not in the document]
andlaus: fair enough

odxtools/ecu_variant_matching.py
```diff
@@ -39,40 +39,48 @@ class State(Enum):
...
         self.use_cache = use_cache
-        self.req_resp_cache: Dict[ByteString, Union[bytes, bytearray]] = {}
-        self._recent_ident_response: Optional[Union[bytes, bytearray]] = None
+        self.req_resp_cache: Dict[ByteString, bytes] = {}
+        self._recent_ident_response: Optional[bytes] = None
 
         self._state = EcuVariantMatcher.State.PENDING
 
-    def request_loop(self):
+    def request_loop(self) -> Generator[bytes, None, None]:
```
andlaus, line 83: do really you want to stick to this name for this generator? IMO it is not a loop, so `request_proceedure()` or `request_sequence()` would be more fitting, but if you insist, I'm okay with it..

dmholtz: The term loop is inspired from control theory rather than computer science. I like it because it contains the notion of feedback, which is happening inside.

odxtools/ecu_variant_matching.py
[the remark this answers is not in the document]
dmholtz: > `typing.Generator[ByteString, None, None]` should do the trick!?
> 
> https://docs.python.org/3/library/typing.html

Thanks, I went for this solution.

andlaus (approved):

odxtools/ecu_variant_matching.py
```diff
@@ -0,0 +1,155 @@
...
+    @staticmethod
+    def decode_ident_response(
+        diag_layer: DiagLayer,
+        matching_param: MatchingParameter,
+        response: Union[bytes, bytearray],
+    ) -> str:
+        service = EcuVariantMatcher.get_ident_service(diag_layer, matching_param)
+        assert service.positive_responses is not None
+        resp_decoded = service.positive_responses[0].decode(response)
+        assert matching_param.out_param_if_snref in resp_decoded
```
kayoub5, line 68: raise an OdxError instead of AssertionError

odxtools/ecu_variant_patterns.py
```diff
@@ -0,0 +1,84 @@
...
+        assert diag_comm_snref is not None
+
+        out_param_snref_el = et_element.find("OUT-PARAM-IF-SNREF")
+        assert out_param_snref_el is not None
+        out_param_if_snref = out_param_snref_el.get("SHORT-NAME")
+        assert out_param_if_snref is not None
+
+        return cls(expected_value, diag_comm_snref, out_param_if_snref)
+
+    def is_match(self, ident_value: Union[int, str]) -> bool:
```
kayoub5, line 45: Not possible for ident_value to be `bytes` ?

odxtools/ecu_variant_matching.py
```diff
@@ -0,0 +1,155 @@
...
+
+    @staticmethod
+    def decode_ident_response(
+        diag_layer: DiagLayer,
+        matching_param: MatchingParameter,
+        response: Union[bytes, bytearray],
+    ) -> str:
+        service = EcuVariantMatcher.get_ident_service(diag_layer, matching_param)
+        assert service.positive_responses is not None
+        resp_decoded = service.positive_responses[0].decode(response)
```
kayoub5, line 67: What happens if positive_responses is empty?

odxtools/ecu_variant_matching.py
```diff
@@ -0,0 +1,155 @@
...
+        assert matching_param.out_param_if_snref in resp_decoded
+        return resp_decoded[matching_param.out_param_if_snref]
```
kayoub5, line 69: Why not just return None if not found?

odxtools/ecu_variant_patterns.py
dmholtz: Hmm, difficult... According to the ISO, the expected value is of type string. Thus, the `ident_value` should also be of type `str` in order to compare the values.

During the tests, I realized that odxtools convert numeric (integer) values from responses of a `DiagService` to `int`. As a workaround, I converted these values to `str`, which is straightforward. 

If the `ident_value` was of type `bytes`, we would have to know its encoding, which we do not know here. That does not imply, that the response's DOP cannot be a HexDump or something similar. We only need to make sure, that we can derive a string representation from whatever DOP it is. What Python datatype do odxtools use to represent a HexDump? @kayoub5 @andlaus
I guess we have similar issues with non-integer numeric values, such as float. Casting could be dangerous, since the decimal separator could be `.` or `,` in the string representation...

Having now thought more about this in greater detail, I guess it is the best if we strictly demand a `str` here. The conversion is deferred to the caller, who has likely more context about the DOP than we have here.

odxtools/ecu_variant_matching.py
dmholtz: `OdxError` is possibly better here

odxtools/ecu_variant_patterns.py
```diff
@@ -0,0 +1,84 @@
...
+        expected_value = et_element.findtext("EXPECTED-VALUE")
+        assert expected_value is not None
+
+        diag_com_snref_el = et_element.find("DIAG-COMM-SNREF")
+        assert diag_com_snref_el is not None
+        diag_comm_snref = diag_com_snref_el.get("SHORT-NAME")
+        assert diag_comm_snref is not None
+
+        out_param_snref_el = et_element.find("OUT-PARAM-IF-SNREF")
+        assert out_param_snref_el is not None
```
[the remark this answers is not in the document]
dmholtz: Thanks, good catch!

odxtools/ecu_variant_matching.py
dmholtz: Thanks, good remark! The ISO wants the `snref` or `snpathref` be resolvable in "at least one" positive response. I will loop over all the response and raise another `OdxError` if it is not resolvable at all. Unfortunately, it is not stated what happens if it is resolvable more than once. So lets choose the first occurence.

odxtools/ecu_variant_matching.py
dmholtz: Because I don't want to suggest, that finding the value is optional. The ISO wants the snref or snpathref be resolvable in "at least one" positive response. Let's rather raise an `OdxError` in case no value is found.

odxtools/ecu_variant_matching.py
```diff
@@ -0,0 +1,155 @@
...
+    ```python
+
+    # initialize the matcher with a list of ecu variants,
+    # i.e., DiagLayer instances of variant_type == DIAG_LAYER_TYPE.ECU-VARIANT
+    matcher = EcuVariantMatcher(ecu_variant_candidates=[...], use_cache=use_cache)
+
+    # run the request loop to obtain responses for every request
+    for req in matcher.request_loop():
+        resp = ... # make a UDS request
+        matcher.evaluate(resp)
```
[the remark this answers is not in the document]
dmholtz: I cannot think of any use-case where functional addressing is appropriate for ECU variant identification. Before the request, we do know the ECU and send a ECU-specific request and expect a ECU-specific response. This is different from broadcasting a fault-read, where we may be interested interested in the responder. But maybe I am mistaken, do you know any uses-cases?

The EcuVariantMatcher is designed to find the variant of exactly one ECU at a time and can be considered a optional, supplementary service. Unless there is no severe flaw in the algorithm, let's stick to this implementation, which is well tested for our use-cases. Would you agree @andlaus?

dmholtz:
Thanks a lot for the feedback @kayoub5 

I will integrate / commit the proposals ASAP.

odxtools/ecu_variant_patterns.py
kayoub5: float is not likely to be used since float equality is always a problematic topic.

for bytes we could assume an ascii encoding, not likely anyone will use non latin characters in an ecu identification

andlaus: > During the tests, I realized that odxtools convert numeric (integer) values from responses of a DiagService to int. As a workaround, I converted these values to str, which is straightforward.

I suppose that what it is basically depends on what your PDX defines as physical type. I guess it would be better either to define this argument as `Any` and demand that it can be compared, or pass the corresponding `Parameter` object along with it.

> What Python datatype do odxtools use to represent a HexDump?

Maybe I am missing something here (ODX is even more twisted than usual in that regard), but I think this works via the parameter's `dop`: conversion of the raw values from the wire to the internal type is done via the DOPs [`convert_physical_to_internal()`](https://github.com/mercedes-benz/odxtools/blob/main/odxtools/dataobjectproperty.py#L167) method, the reverse is basically done using [`convert_physical_to_bytes()`](https://github.com/mercedes-benz/odxtools/blob/main/odxtools/dataobjectproperty.py#L63) which might be what you want...

odxtools/ecu_variant_matching.py
andlaus: I guess the assert on line 66 should ensure this then?!

odxtools/ecu_variant_matching.py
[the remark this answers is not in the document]
andlaus: I'm fine with anything that works ;)

odxtools/ecu_variant_matching.py
andlaus: (i.e.
```python
assert service.positive_responses is not None
assert len(service.positive_responses) > 0
```

odxtools/ecu_variant_matching.py
```diff
@@ -0,0 +1,155 @@
+# SPDX-License-Identifier: MIT
```
andlaus, line 1: I noticed this only now (sorry): if possible, the file name should correspond to the class name it contains (i.e., `ecu_variant_matcher.py`)

dmholtz (commented):
All discussions have been resolved and the PR is ready to merge.

The implementation of the matching algorithm is now closer to the ISO:
 - Add support for snpathref
 - Try to resolve snref / snpathref in any response

I added further tests to check the implementation.

dmholtz:
Thanks for the review @andlaus @kayoub5

odxtools/utils.py
```diff
@@ -43,3 +46,21 @@ def short_name_as_id(obj: Any) -> str:
...
 
     return sn
+
+
+def is_short_name(test_val: str) -> bool:
+    """Returns true iff the test_val string is a ODX short name.
+
+    See also: ISO 22901 section 7.1.1
+    """
+    short_name_pattern = re.compile("[a-zA-Z0-9_]+")
```
kayoub5, line 56: compiling regex is an expensive operation, you may want to move it out of the function so it only gets done once

kayoub5: other than this, LGTM

kayoub5 (approved):

PR #117, opened 2023-01-12
No parameter defaults

andlaus:
This PR firmly belongs into the "long and boring" category. It removes the default specifiers for class constructor arguments and dataclass attributes. The reasons are that with defaults specified, it is all too easy to miss internalizing optional XML attributes even if resulting object will feature it and that for dataclasses it is impossible to create class hierarchies with defaulted attributes without `kw_only=True` (which is only available for python 3.10 or newer).


odxtools/endofpdufield.py
```diff
@@ -148,9 +141,14 @@ def _resolve_references(self, # type: ignore[override]
...
         if self.structure_ref is not None:
             self._structure = odxlinks.resolve(self.structure_ref)
-        else:
+        elif self.structure_snref is not None:
             self._structure = parent_dl.data_object_properties[self.structure_snref]
 
+        if self.env_data_desc_ref is not None:
+            self._env_data_desc = odxlinks.resolve(self.env_data_desc_ref)
+        elif self.env_data_desc_snref is not None:
+            self._env_data_desc = parent_dl.data_object_properties[self.env_data_desc_snref]
```
kayoub5, line 150: Does not looks related to PR topic, what this about?

odxtools/envdatadesc.py
```diff
@@ -19,15 +19,13 @@ class EnvironmentDataDescription(DopBase):
...
                  env_data_refs: List[OdxLinkRef],
-                 param_snref: Optional[str] = None,
-                 param_snpathref: Optional[str] = None,
+                 param_snref: Optional[str],
+                 param_snpathref: Optional[str],
                  **kwargs):
         super().__init__(**kwargs)
 
         self.bit_length = None
-        self.env_datas = env_datas
```
kayoub5, line 30: What happened here?

odxtools/envdatadesc.py
```diff
@@ -52,31 +52,21 @@ def from_et(et_element, doc_frags: List[OdxDocFragment]) \
             assert ref is not None
             env_data_refs.append(ref)
 
-        # ODX 2.0.0 says ENV-DATA-DESC could contain a list of ENV-DATAS
-        env_datas = []
-        for env_data_elem in et_element.iterfind("ENV-DATAS/ENV-DATA"):
-            env_data = EnvironmentData.from_et(env_data_elem, doc_frags)
-            env_datas.append(env_data)
```
kayoub5, line 59: Not parsing this anymore?

odxtools/structures.py
```diff
@@ -548,20 +514,11 @@ def __str__(self) -> str:
...
-                 description=None,
-                 sdgs=[]):
-        super().__init__(odx_id=odx_id,
-                         short_name=short_name,
-                         parameters=parameters,
-                         long_name=long_name,
-                         description=description,
-                         sdgs=sdgs)
-        self.response_type = "POS-RESPONSE" if response_type == "POS-RESPONSE" else "NEG-RESPONSE"
+                 response_type: str, # "POS-RESPONSE" or "NEG-RESPONSE"
```
kayoub5, line 517: Literal type hint instead of comment?

odxtools/table.py
```diff
@@ -139,13 +137,12 @@ def __init__(self,
...
                  short_name: str,
                  table_rows: List[TableRow],
-                 table_row_refs: Optional[List[OdxLinkRef]] = None,
-                 long_name: Optional[str] = None,
-                 key_dop_ref: Optional[OdxLinkRef] = None,
-                 description: Optional[str] = None,
-                 semantic: Optional[str] = None,
-                 sdgs: List[SpecialDataGroup] = [],
-                 ):
+                 table_row_refs: Optional[List[OdxLinkRef]],
```
kayoub5, line 140: Since all fields are now mandatory, lists should be initialized with empty lists instead of None

```
table_row_refs: List[OdxLinkRef],
```

odxtools/audience.py
```diff
@@ -6,46 +6,74 @@
...
+        return AdditionalAudience(odx_id=odx_id,
+                                  short_name=short_name,
+                                  long_name=long_name,
+                                  description=description)
+
+    def _build_odxlinks(self) -> Dict[OdxLinkId, Any]:
+        return { self.odx_id: self }
+
+    def _resolve_references(self, odxlinks: OdxLinkDatabase) -> None:
+        pass
```
kayoub5, line 39: Why have an empty _resolve_references ?

odxtools/diagdatadictionaryspec.py
```diff
@@ -30,32 +30,16 @@ def _construct_named_item_list(iterable):
...
+    dtc_dops: NamedItemList[DtcDop]
+    data_object_props: NamedItemList[DataObjectProperty]
+    structures: NamedItemList[BasicStructure]
+    end_of_pdu_fields: NamedItemList[EndOfPduField]
+    tables: NamedItemList[Table]
+    env_data_descs: NamedItemList[EnvironmentDataDescription]
+    env_datas: NamedItemList[EnvironmentData]
+    muxs: NamedItemList[Multiplexer]
+    unit_spec: Optional[UnitSpec]
+    sdgs: List[SpecialDataGroup]
```
kayoub5, line 42: Still needed?
````
# The attributes are already initialized as (most likely normal) lists.
# Convert them to NamedItemLists if necessary
```

odxtools/envdata.py
```diff
@@ -17,6 +17,7 @@
 @dataclass
 class EnvironmentData(BasicStructure):
     """This class represents Environment Data that describes the circumstances in which the error occurred."""
+    dtc_values: List[int]
```
kayoub5, line 20: Where did this come from?

tests/test_unit_spec.py
```diff
@@ -142,7 +166,18 @@ def test_resolve_references(self):
             single_ecu_jobs=[],
             diag_comm_refs=[],
             diag_data_dictionary_spec=DiagDataDictionarySpec(data_object_props=[dop],
-                                                             unit_spec=UnitSpec(units=[unit])
+                                                             unit_spec=UnitSpec(units=[unit],
+                                                                                physical_dimensions=None,
+                                                                                unit_groups=None,
+                                                                                sdgs=[],),
+                                                             dtc_dops = NamedItemList(short_name_as_id, []),
```
kayoub5, line 173: Do we still need `short_name_as_id` ?

odxtools/ecu_variant_matcher.py
```diff
@@ -97,7 +97,7 @@ def decode_ident_response(
                 resolved for any positive or negative response."
         )
 
-    def __init__(self, ecu_variant_candidates: List[DiagLayer], use_cache: bool = True):
```
kayoub5, line 100: Not an XML model class, I think the default can stay

odxtools/endofpdufield.py
andlaus: this was probably a "drive-by" change most likely motivated because I got some errors in the end-of-pdu-field class. I'll split it out.

odxtools/envdatadesc.py
andlaus: the XSD allows only to specify ENV-DATA-REFS, so we should not take a list of full  environment data objects here. I can factor this out, if wanted...

odxtools/envdatadesc.py
andlaus: oops. ok, odx 2.0. I didn't find a ENV-DATAS sub-element in the XSD. I'll revert.

odxtools/table.py
andlaus: oops, true...

andlaus:
ok, i've factored out the unrelated changes which you found, I hope that this was everything. They will come in a separate (much smaller) PR. thanks for the review!

odxtools/envdata.py
andlaus: this is new. I noticed that the XSD features a "DTC-VALUES" tag, but the python class didn't. I'll factor it out...

andlaus: actually, the above is inaccurate: currently, `EnvironmentData` is a dataclass that defines `__init__()` and `dtc_values` was only created by the latter. I suppose the manual constructor ought to be removed...

andlaus: ... but that is currently not so easy because the base class (`BasicStructure`) is currently not yet a dataclass. let's just keep that kludge for now...

tests/test_unit_spec.py
andlaus: yes, because short names may start with a digit, which is not allowed in python. maybe that case should be made easier, e.g. by introducing a `ShortNameItemList` derived from `NamedItemList`? (in this particular instance, one could also just pass through the short names, because the unit test does not (yet) feature one which starts with a digit...)

odxtools/ecu_variant_matcher.py
andlaus: ok

odxtools/companydata.py
```diff
@@ -66,7 +65,7 @@ def from_et(et_element) -> "RelatedDoc":
 @dataclass
 class CompanySpecificInfo:
     related_docs: Optional[List[RelatedDoc]]
```
kayoub5, line 67:
```suggestion
    related_docs: List[RelatedDoc]
```

odxtools/companydata.py
```diff
@@ -99,16 +98,16 @@ def _resolve_references(self, odxlinks: OdxLinkDatabase):
...
-    department: Optional[str] = None
-    address: Optional[str] = None
-    zip: Optional[str] = None
-    city: Optional[str] = None
-    phone: Optional[str] = None
-    fax: Optional[str] = None
-    email: Optional[str] = None
+    long_name: Optional[str]
+    description: Optional[str]
+    roles: Optional[List[str]]
```
kayoub5, line 103:
```suggestion
    roles: List[str]
```

odxtools/companydata.py
```diff
@@ -153,11 +152,11 @@ def from_et(et_element: ElementTree.Element,
...
     odx_id: OdxLinkId
     short_name: str
-    long_name: Optional[str] = None
-    description: Optional[str] = None
-    roles: Optional[List[str]] = None
-    team_members: Optional[NamedItemList[TeamMember]] = None
-    company_specific_info: Optional[CompanySpecificInfo] = None
+    long_name: Optional[str]
+    description: Optional[str]
+    roles: Optional[List[str]]
```
kayoub5, line 157:
```suggestion
    roles: List[str]
```

odxtools/compumethods/compurationalcoeffs.py
```diff
@@ -1,10 +1,7 @@
...
 # Copyright (c) 2022 MBition GmbH
-
-
 from typing import List, NamedTuple, Optional
 
-
 class CompuRationalCoeffs(NamedTuple):
     numerators: List[float]
-    denominators: Optional[List[float]] = None
+    denominators: Optional[List[float]]
```
kayoub5, line 7:
```suggestion
    denominators: List[float]
```

odxtools/multiplexer.py
```diff
@@ -168,8 +166,8 @@ class Multiplexer(DopBase):
     long_name: str
     byte_position: int
     switch_key: MultiplexerSwitchKey
-    default_case: Optional[MultiplexerDefaultCase] = None
-    cases: Optional[List[MultiplexerCase]] = None
+    default_case: Optional[MultiplexerDefaultCase]
+    cases: Optional[List[MultiplexerCase]]
```
kayoub5, line 170:
```suggestion
    cases: List[MultiplexerCase]
```

tests/test_ecu_variant_matching.py
```diff
@@ -174,15 +238,31 @@ def ecu_variant_3(
...
+        import_refs=[],
+        communication_parameters=[],
+        diag_comm_refs=[],
+        diag_data_dictionary_spec=None,
         services=[ident_service, supplier_service],
+        requests=[],
+        positive_responses=[],
+        negative_responses=[],
+        single_ecu_jobs=[],
+        states=[],
```
dmholtz, line 259: Testers do not like this change ;-)

But seriously, I think it is acceptable since the constructor of DiagLayer is usually not part of the public interface.

dmholtz (approved):
LGTM.
The increased realiability of XML internalization should outweight the more verbose interface for testing. None of these methods are part of the public interface.

tests/test_unit_spec.py
dmholtz: I think they can start with a digit, I have just looked in the spec yesterday and implemented a checker function according to the ISO: https://github.com/mercedes-benz/odxtools/blob/b14262fd4d9b51ad79e40148e9ce663c36ab2387/odxtools/utils.py#L57

andlaus: > Short names can't start with digit, it's the IDs that can

wasn't it you who had a PDX file that committed that atrocity ;)? I'm 100% sure that the original code used the short_name directly.

odxtools/multiplexer.py
andlaus: Good suggestion. I've already implemented these changes but I'd like to keep them out of this PR due to their off-topic nature. (will do a separate PR once this one is merged to avoid conflicts and confusion.)

tests/test_unit_spec.py
kayoub5: yes, it was an odx from me, and it was in the comparam id, now comparam uses short names for keys

andlaus: the issue with the IDs is, although they have to start with a letter, they may contain hyphens. these are not allowed in short names, but -- in order to settle the balance -- they may start with a digit. IIRC you reported both of these issues...

kayoub5: I will recheck this later, either way it's a suggestion that's outside the scope of the PR

odxtools/diagdatadictionaryspec.py
andlaus: actually no: now NamedItemLists need to be provided to the constructor, i.e., no need to convert them in post anymore. Since this is IMO a bit out of this PR's scope, I'll fix that in the next one...

odxtools/table.py
andlaus: (will be fixed in the next PR)

kayoub5 (approved):

PR #119, opened 2023-01-20
add a shell skript to reformat the whole code base

andlaus:
this script is simply a call to [`YAPF`](https://github.com/google/yapf) for code formatting followed by a run of `isort` to correctly sort the imports. The main purpose of this script is to document how to bring the coding style in line with expectations, but it also makes it easier to reformat code before submitting a PR...

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -7,6 +7,14 @@ requires = [
 
 build-backend = "setuptools.build_meta"
 
+[tool.yapf]
+based_on_style = "yapf"
+column_limit = 100
+indent_width = 4
+split_before_dict_set_generator = 1
+split_before_first_argument = 0
+split_penalty_after_opening_bracket=1000
+
 [tool.pytest.ini_options]
 filterwarnings = [
     "error::odxtools.exceptions.DecodeError",
--- a/reformat-source.sh
+++ b/reformat-source.sh
@@ -1,0 +1,19 @@
+#! /bin/sh
+#
+# This script brings the style of all python code within the odxtools
+# director into conformance the style mandated for the odxtool
+# project. Before you open a pull request, run it on your working
+# copy. (You need to have yapf with toml support and isort installed:
+#
+# pip3 install yapf isort toml
+#
+if ! test -f odxtools/version.py || \
+   ! test -d tests || \
+   ! test -d examples; then
+
+    echo "You need to run this script in the topmost directory" >&2
+    echo "of a working copy of odxtools." >&2
+    exit 1
+fi
+yapf -r -i odxtools/ tests/ examples/
+isort -l 100 -m VERTICAL odxtools tests examples


andlaus:
if anybody has anything in the making, please get it merged until Wednesday next week (25th January) to avoid conflicts. If you can't make it until then, please leave a message here so that the Big Style Update can be delayed...

andlaus:
also: If you're not happy with the result, please discuss the options for `black` and `isort` here...

reformat-source.sh
```diff
@@ -0,0 +1,4 @@
+#! /bin/sh
+
+black -l 100 odxtools examples tests
```
kayoub5, line 3: keep black config in pyproject.toml instead of cli arguments, so that IDE takes it into account.

bogdzn:
You probably should take a look at a project like [shellcheck](https://www.shellcheck.net/). Shell/Bash scripts can be trick and it's easy to fall into a trap, it's NBD for a small script like this, but it might be good to lint when doing new features on it.

kayoub5 (approved):

reformat-source.sh
```diff
@@ -0,0 +1,19 @@
...
+#
+if ! test -f odxtools/version.py || \
+   ! test -d tests || \
+   ! test -d examples; then
+
+    echo "You need to run this script in the topmost directory" >&2
+    echo "of a working copy of odxtools." >&2
+    exit 1
+fi
+yapf -r -i odxtools/ tests/ examples/
```
dmholtz, line 18: As far as I can see, `yapf` is now used rather than `black`. You should reflect this change in the description of PR #119 and #120

pyproject.toml
```diff
@@ -7,6 +7,14 @@ requires = [
 
 build-backend = "setuptools.build_meta"
 
+[tool.yapf]
```
dmholtz, line 10: IMO, black with default configuration would have been sufficient. But of course I'm fine with everything that improves readability.

reformat-source.sh
andlaus: ok. I switched the code formatter to `YAPF` because I like its result better than those of `black`. Besides having a *huge* number of knobs, yapf seems to prefer to be configured  via `pyproject.toml`...

reformat-source.sh
andlaus: done.

pyproject.toml
andlaus: I like the results of the proposed setup better than those produced by `black` because it is closer to the style used for most of the existing odxtools codebase. That said, discussing what the best style is is the main point of this PR. (and also, that we can avoid having to do such discussions again in the future ;))

dmholtz (approved):

andlaus:
#120 has been merged, and #120 includes this one. closing.

PR #120, opened 2023-01-20
Reformat

andlaus:
This PR contains #119, the results of running that script and fixes the minor mypy hiccups that were caused by that. (mainly these were due to the fact that various `type: ignore` comments ended up on the wrong line.) IOW, this PR contains no functional changes.
    
after this is merged, we can run the reformat script as part of the CI pipeline and make the CI fail if the reformat script results in any changes.


andlaus:
@kayoub5: I'd like your opinion on that matter: Do you agree with the style enforced by `black` and should we run the reformatter as part of the CI pipeline?

kayoub5 (approved):
All automated formatting, so nothing really to review

kayoub5:
> @kayoub5: I'd like your opinion on that matter: Do you agree with the style enforced by `black` and should we run the reformatter as part of the CI pipeline?

* Does not really matter what style to use as long as I can integrate it in my IDE
* I worked in projects that fail the CI before even running the tests if the code format check fails
* Yes, you want this enforced format in CI to avoid merge conflict/diff caused by auto formatting

andlaus:
> -  Does not really matter what style to use as long as I can integrate it in my IDE

I agree: I also do not care too much about the coding style which people use. That said, I consider enforcing consistency to be somewhat important: If a contributor's IDE decides to automatically "fix" the style of the files which are modified, reviewing the actual changes becomes pretty challenging and the PR becomes very prone to merge conflicts which in turn tend to demotivate contributors (example: #97)

> - I worked in projects that fail the CI before even running the tests if the code format check fails

I think we shouldn't be that strict. IMO we should let the tests run in the CI besides the formatter and let the run fail if either fails of these. This should reduce the number of necessary round trips...

> - Yes, you want this enforced format in CI to avoid merge conflict/diff caused by auto formatting

+1

andlaus:
just a heads-up: If nothing happens, I'll merge this tomorrow...

dmholtz:
Two more issues to consider before merging:

- yapf and isort influence each other since they define different styles
- isort is not properly configured to be supported by the IDE

Currently, the only deterministic way to achieve the desired style is to run the shell script. This is not feasible, if only a single file is touched. The commands defined in the shells script runs over a minute on my Windows machine.

andlaus:
okay. it turns out that there were some issues with IDEs due to the fact that `isort` changes the formatting of some imports slightly compared to YAPF. We thus decided that `isort` will not be included in the CI pipeline (but be still part  of the reformatting script) and that the import statement indentation style of yapf is the cannonical. (i.e., isort is run before  yapf in the reformatting script)

andlaus:
Let's get over this!

PR #121, opened 2023-01-23
Decompose action into CI pipeline jobs

dmholtz:
The goal of this refactoring is to make the CI pipeline more transparent and enhance the traceability of failures.

The new pipeline has
- a lint job (using yapf, see also #119 and #120), 
- a static-type-checking job and
- a test job.

Each job runs independently from each other. To improve efficiency, pip dependencies are cached.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -1,4 +1,7 @@
-name: Run Tests
+# SPDX-License-Identifier: MIT
+# Copyright (c) 2023 MBition GmbH
+
+name: odxtools CI pipeline
 
 on:
   push:
@@ -8,20 +11,32 @@ on:
   workflow_dispatch:
 
 jobs:
-  unittest-and-lint:
-    runs-on: ${{matrix.os}}
+  lint:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v3
+
+      - name: Set up yapf linter
+        uses: actions/setup-python@v4
+        with:
+          python-version: "3.11"
+        run: |
+          pip install toml yapf
+
+      - name: Lint code formatting
+        # make yapf fail if any file needs to be formatted
+        run: |
+          yapf -r --diff examples/ odxtools/ tests/
+
+  static-type-checking:
+    runs-on: ubuntu-latest
 
     strategy:
       matrix:
-        os: [ubuntu-latest, windows-latest]
-        # restrict matrix to the latest Python version and one older version
+        # restrict the matrix to the oldest and the latest Python
+        # version being supported by odxtools
         python-version: ["3.8", "3.11"]
-
-        # due to the slow windows runners, we refrain from testing every python
-        # version on windows-latest
-        exclude:
-          - os: windows-latest
-            python-version: "3.8"
 
     steps:
       - uses: actions/checkout@v3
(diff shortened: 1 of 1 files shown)


dmholtz:
This PR shall be merged after #119 and #120 and I will do a rebase beforehand.

.github/workflows/tests.yml
```diff
@@ -8,13 +11,59 @@ on:
...
+          yapf -r --diff examples/ odxtools/ tests/
+
+  static-type-checking:
+    runs-on: ubuntu-latest
 
+    strategy:
+      matrix:
+        # restrict the matrix to the oldest and the latest Python
+        # version being supported by odxtools
+        python-version: ["3.8", "3.11"]
```
kayoub5, line 41: Maybe use `--python-version` argument of mypy instead?

https://mypy.readthedocs.io/en/stable/command_line.html#cmdoption-mypy-python-version

.github/workflows/tests.yml
```diff
@@ -8,13 +11,59 @@ on:
...
+      - name: Set up Python ${{matrix.python-version}}
+        uses: actions/setup-python@v4
+        with:
+          python-version: ${{matrix.python-version}}
+          cache: "pip"
+
+      - name: Install odxtools development and testing dependencies
+        run: |
+          pip install -r requirements.txt
+          pip install -r test-requirements.txt
```
kayoub5, line 55: Are those requirements really needed to just run mypy ?

.github/workflows/tests.yml
```diff
@@ -8,13 +11,59 @@ on:
...
+      - uses: actions/checkout@v3
+
+      - name: Set up Python
+        uses: actions/setup-python@v4
+        with:
+          python-version: "3.11"
+
+      - name: Setup up Linter
+        run: |
+          pip install toml yapf
```
kayoub5, line 27: You may want to specify the version of those dependencies

.github/workflows/tests.yml
dmholtz: Except for `coverage`, which is contained in `test-requirements.txt`, mypy needs at least stubs with type annotations of every imported package.

.github/workflows/tests.yml
dmholtz: I prefer the matrix notation and thus multiple job instances, because one can precisely tell, which runner (i.e. which version) failed. There is no need to have a look at the logs of Github Actions. It happens quite often, that certain typing annotations are supported with the latest version of Python but not with previous versions.

.github/workflows/tests.yml
dmholtz: Both toml and yapf seem to be pretty stable, and thus latest versions are fine from my point of view. Anyway, we explicitly specify the style in the pyproject.toml as proposed in #119

dmholtz:
Thanks for the review @kayoub5 !

.github/workflows/tests.yml
andlaus: I tend to agree with David: if tests fail due to updated tools, we should scramble to fix our code anyway. if this happens to be impossible, (I have trouble comming up with such a case), we can pin the versions then...

.github/workflows/tests.yml
```diff
@@ -8,13 +11,59 @@ on:
...
+
+      - name: Lint code formatting
+        # make yapf fail if any file needs to be formatted
+        run: |
+          yapf -r --diff examples/ odxtools/ tests/
```
andlaus, line 32: in order to be consistent with #119, you need to run `isort` after that: YAPF does not sort the imports according to PEP8 and it also changes the indentation of multi-line imports (I like the way isort does it better...)

```suggestion
          yapf -r --diff examples/ odxtools/ tests/
          isort -l 100 -m VERTICAL odxtools/ tests/ examples/
```

.github/workflows/tests.yml
andlaus, line 27: `isort` is also required (see below)

.github/workflows/tests.yml
andlaus: BTW: shouldn't we let the step fail if the source changes as a result of the reformatting action?

andlaus: (this can be done using

```suggestion
          yapf -r -i odxtools/ tests/ examples/
          isort -l 100 -m VERTICAL odxtools/ tests/ examples/
          git diff --exit-code 2>&1 > /dev/null
```
)

dmholtz: The `--diff` option of `yapf`is simpler: It returns a non-zero return code in case something is reformatted and is supposed to be used in CIs according to the docs.

But you are right, isort is missing. I guess there is also a check option, otherwise I will go for the `git diff`

dmholtz:
@andlaus I guess this can be merged. The pipeline is stable now and all checks are passed.

andlaus:
ok, let's do it! (TM)

PR #126, opened 2023-05-15
implement state charts

andlaus:
so far, we only had a global list of states and state transitions to make the Services and Single-ECU jobs work (in most cases). Also, due to missing information, write support was obviously not implemented.


odxtools/statechart.py
```diff
@@ -0,0 +1,93 @@
...
+
+    @property
+    def start_state(self) -> State:
+        return self._start_state
+
+    @staticmethod
+    def from_et(et_element, doc_frags: List[OdxDocFragment]) -> "StateChart":
+        short_name = et_element.findtext("SHORT-NAME")
+        odx_id = OdxLinkId.from_et(et_element, doc_frags)
+        assert odx_id is not None
```
kayoub5, line 34: throw an odx exception with a message instead of an asset.

odxtools/templates/macros/printState.xml.jinja2
```diff
@@ -0,0 +1,18 @@
+{#- -*- mode: sgml; tab-width: 1; indent-tabs-mode: nil -*-
```
kayoub5, line 1: what is this?

odxtools/statechart.py
```diff
@@ -0,0 +1,93 @@
...
+
+        long_name = et_element.findtext("LONG-NAME")
+        description = create_description_from_et(et_element.find("DESC"))
+
+        semantic = et_element.findtext("SEMANTIC")
+        assert semantic is not None
```
kayoub5, line 40: the code does not depend on the value of this field, maybe a warning is enough?

odxtools/statechart.py
andlaus: It is on my list: I'd like to introduce an `odxassert(condition, "error message")` method which does this. but it is IMO easier to keep using plain old `assert` for now and switch everything in one go...

odxtools/statechart.py
andlaus: the XSD says that the `SEMANTIC` tag is mandatory. you're right that this should not be an plain `assert`, tough...

kayoub5 (approved):

examples/isotp_send.py
```diff
@@ -38,7 +38,7 @@ def isotp_error_handler(error):
 # bs: maximum block size. (Must be smaller than 4096?)
 isotp_socket.set_fc_opts(stmin=5, bs=100)
 
-can_bus = can.Bus(channel=channel, bustype="socketcan")
+can_bus = can.Bus(channel=channel, interface="socketcan")  # type: ignore[abstract]
```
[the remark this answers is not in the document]
andlaus: true. I forgot about that when pushing, do you prefer to have this factored out now that it slipped in anyway?

kayoub5:
the changes look good, the only thing is that formatting and linting changes not related to original topic are better off moved to their own pr.

andlaus:
sure. the problem with this is that the CI system always uses the latest YaPF version, i.e., perfectly fine PRs can fail by just letting some time pass if YaPF changes (as happened here). On the other hand, I'd like to continue to be strict when it comes to enforcement of the style, so we now have a dilemma...

Anyway, let's get this merged. Thanks, @kayoub5 !

PR #128, opened 2023-05-24
- resolved unused import in write_pdx_file.py

louwersj:
removed two import lines from write_pdx_file.py as they are not used. making the overall codebase less memory consuming and the code more clean.

DO NOTE: Contributors License Agreement is in progress and I will send it in a bit. Will notify as part of this pull request when that is done.

### DIFF
--- a/odxtools/write_pdx_file.py
+++ b/odxtools/write_pdx_file.py
@@ -5,14 +5,9 @@ import inspect
 import os
 import time
 import zipfile
-from typing import Any, Dict, List, Optional, Tuple
-
 import jinja2
-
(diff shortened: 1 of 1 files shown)


louwersj:
DO NOTE: Contributors License Agreement has been signed and send.

andlaus:
I have my doubts about whether this improves anything, but I also cannot see how it could make things worse, so let's merge. Welcome to the odxtools contributors @louwersj!

louwersj:
@andlaus I think I just very much like clean code... and in theory it should save a tiny bit of memory allocation.

PR #129, opened 2023-05-24
code cleanup

louwersj:
- Removed a number of unused import statements
- Removed a number of unused parameters / variables

### DIFF
--- a/odxtools/structures.py
+++ b/odxtools/structures.py
@@ -2,8 +2,7 @@
 # Copyright (c) 2022 MBition GmbH
 import math
 import warnings
-from typing import (TYPE_CHECKING, ByteString, Dict, Iterable, List, Optional, OrderedDict,
-                    Tuple, Union)
+from typing import (TYPE_CHECKING, ByteString, Dict, Iterable, List, Optional, OrderedDict, Tuple, Union)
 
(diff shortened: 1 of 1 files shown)


andlaus (commented):
once the style sheriff is happy, I don't have any objections to this. (to make it happy, use
```
python3 -m pip install yapf isort
cd $ODXTOOLS_SOURCE_DIR
./reformat-source.sh
```
)

louwersj:
@andlaus I will have a chat with the "sheriff" and come back with some changes :)

andlaus (changes requested):
I agree the object initialization is currently not perfect, but IMO these changes go into the wrong direction. For all classes which currently have a `_resolve_references()` I envision to change the object resolution scheme to something like this:
- `_resolve_references()` should probably be renamed to `_resolve_odxlinks()` and only resolve ODXLINK references
- a new method named `_resolve_snrefs(diag_layer)` should be introduces which resolves the short name references. (note tat this is not trivial, because SNREFS can point to objects inherited by some parent layer.)
- possibly introduce a `finalize_init()` method to do something with the object where all references have been completely resolved...

andlaus:
this PR is now obsolete: we now run `ruff` as part of the CI, ensuring that all imported modules are used. closing.

PR #130, opened 2023-05-27
fix/wrong denominator linearcompumethod

JoSglch:
Fixes a bug in _parse_compu_scale_to_linear_compu_method() to set the denominator correctly. 

Currently the denominator is always set to 1, even if it was parsed correctly before.


odxtools/compumethods/createanycompumethod.py
```diff
@@ -52,9 +52,10 @@ def _parse_compu_scale_to_linear_compu_method(
     offset = computation_python_type(next(nums).text)
     factor_el = next(nums, None)
     factor = computation_python_type(factor_el.text if factor_el is not None else "0")
+    denominator = 1.0
     if coeffs.find("COMPU-DENOMINATOR/V") is not None:
-        kwargs["denominator"] = float(coeffs.findtext("COMPU-DENOMINATOR/V"))
-        assert kwargs["denominator"] > 0
+        denominator = float(coeffs.findtext("COMPU-DENOMINATOR/V"))
+        assert denominator > 0
```
kayoub5, line 58: denominator may be negative

andlaus: and since python 3.8 has been required for a while, this can be done more efficiently like this:
```suggestion
    if (denom_str := coeffs.findtext("COMPU-DENOMINATOR/V")) is not None:
        denominator = float(denom_str)
```

(not a big deal, though)

further, it would be nice if you verified that the writing code for this was correct: https://github.com/mercedes-benz/odxtools/blob/main/odxtools/templates/macros/printDOP.xml.jinja2#L102

odxtools/compumethods/createanycompumethod.py
```diff
@@ -52,9 +52,10 @@ def _parse_compu_scale_to_linear_compu_method(
     offset = computation_python_type(next(nums).text)
     factor_el = next(nums, None)
     factor = computation_python_type(factor_el.text if factor_el is not None else "0")
+    denominator = 1.0
     if coeffs.find("COMPU-DENOMINATOR/V") is not None:
-        kwargs["denominator"] = float(coeffs.findtext("COMPU-DENOMINATOR/V"))
-        assert kwargs["denominator"] > 0
+        denominator = float(coeffs.findtext("COMPU-DENOMINATOR/V"))
```
andlaus, line 57: and since python 3.8 has been required for a while, this can be done more efficiently like this:
```suggestion
    if (denom_str := coeffs.findtext("COMPU-DENOMINATOR/V")) is not None:
        denominator = float(denom_str)
```

(not a big deal, though)

further, it would be nice if you verified that the writing code for this is correct: https://github.com/mercedes-benz/odxtools/blob/main/odxtools/templates/macros/printDOP.xml.jinja2#L102

JoSglch: I looked at the generation, and there are 2 discrepancies. First I think in [line 98](https://github.com/mercedes-benz/odxtools/blob/main/odxtools/templates/macros/printDOP.xml.jinja2#L98) it should only print the factor.
Second is the line you mentioned, I have not seen a `<COMPU-DENOMINATOR>` with 2 `<V>` elements before, and the current parser is also not handling this case. 
The template probably needs to look like [this](https://github.com/mercedes-benz/odxtools/pull/130/commits/d904126ed98ae0a69618575334934dc9744f177c), but I do not have access to the odx specification, so I can not verify this.

andlaus: the XSD specification for COMPU-DENOMINATOR looks like this:
image
with "V" being any double precision floating point number. the accompanying PDF says that the raw-to-scaled transfer function is defined as
```
f(x) = (n_0 + n_1*x + n_2*x^2 ...)/(d_0 + d_1*x + d_2*x^2 ...)
```
where `n_i` is the `i`-th COMPU-NUMERATOR and `d_i` is the `i`-th COMPU-DENOMINATOR. I suppose this means that for a linear function every `n_i` except `n_0` and `n_1` must be 0 and `d_0` is always 1 while everything else is 0. (not the most intuitive way to define things IMO...)

Looking at this now, I think that the writing code is correct except for the `| int` filters?

odxtools/templates/macros/printDOP.xml.jinja2
```diff
@@ -95,11 +95,10 @@
 				<COMPU-RATIONAL-COEFFS>
 					<COMPU-NUMERATOR>
 						<V>{{cm.offset | int}}</V>
-						<V>{{(cm.factor*cm.denominator) | int}}</V>
+						<V>{{(cm.factor) | int}}</V>
```
andlaus, line 98: looking at the parsing code, this should probably be something like
```
						<V>{{cm.offset*cm.denominator}}</V>
						<V>{{cm.factor*cm.denominator}}</V>
```
?

odxtools/templates/macros/printDOP.xml.jinja2
```diff
@@ -95,11 +95,10 @@
...
 					</COMPU-NUMERATOR>
    {%- if cm.denominator != 1 %}
 					<COMPU-DENOMINATOR>
-							<V>1</V>{# TODO: currently we always assume the offset to be an integer #}
 							<V>{{cm.denominator | int}}</V>
```
andlaus, line 102: this must be
```suggestion
							<V>{{cm.denominator}}</V>
```

andlaus:
@JoSglch: what is the status here?

andlaus:
#148 is based on this PR, includes a few fixes and has just been merged. closing...

PR #131, opened 2023-05-27
Fix/max length parsing

JoSglch:
Fixes a bug in create_any_diag_coded_type_from_et(). The check to see if a ``<MAX-LENGTH>`` element exists fails, because Element in xml.etree.ElementTree overwrites \_\_bool\_\_() and ``<MAX-LENGTH>`` never has any children ([see implementation](https://github.com/python/cpython/blob/9f89c471fb152fd031791f3c598ceba93d91dcf0/Lib/xml/etree/ElementTree.py#L207)).

Changed the check to the recommended explicit is not None comparison.

### DIFF
--- a/odxtools/diagcodedtypes.py
+++ b/odxtools/diagcodedtypes.py
@@ -640,7 +640,8 @@ def create_any_diag_coded_type_from_et(et_element, doc_frags: List[OdxDocFragmen
     elif dct_type == "MIN-MAX-LENGTH-TYPE":
         min_length = int(et_element.findtext("MIN-LENGTH"))
         max_length = None
-        if et_element.find("MAX-LENGTH"):
+        # comparison has to be 'is not None' as Element overwrites __bool__(), and always returns false for MAX-LENGTH elements
+        if et_element.find("MAX-LENGTH") is not None:
             max_length = int(et_element.findtext("MAX-LENGTH"))
         termination = et_element.get("TERMINATION")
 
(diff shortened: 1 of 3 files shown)


odxtools/diagcodedtypes.py
```diff
@@ -640,7 +640,8 @@ def create_any_diag_coded_type_from_et(et_element, doc_frags: List[OdxDocFragmen
     elif dct_type == "MIN-MAX-LENGTH-TYPE":
         min_length = int(et_element.findtext("MIN-LENGTH"))
         max_length = None
-        if et_element.find("MAX-LENGTH"):
+        # comparison has to be 'is not None' as Element overwrites __bool__(), and always returns false for MAX-LENGTH elements
```
kayoub5, line 643: nice catch

odxtools/write_pdx_file.py
```diff
@@ -5,9 +5,12 @@
 import os
 import time
 import zipfile
+from typing import Any, Dict, List, Optional, Tuple
+
 import jinja2
+
 import odxtools
-from typing import Any, Dict, List, Optional, Tuple
+
```
kayoub5, line 13: changes unrelated to PR scope

andlaus (approved):
looks good to me. before I can merge this, you have to sign the [CLA](https://github.com/mercedes-benz/odxtools/blob/main/CONTRIBUTORS_LICENSE_AGREEMENT.md) (if you did not do this already...)

JoSglch:
I sent the signed CLA on saturday, so should be ok.

andlaus:
> I sent the signed CLA on saturday, so should be ok.

okay, thanks!

PR #132, opened 2023-05-30
Untangle Diaglayer (1 of X)

andlaus:
This PR marks the start a small quest of mine to untangle the code spaghetti that is currently located in `diaglayer.py`. This PR moves the `ParentRef` and `DiagLayerContainer` classes into their separate files, renames the `DIAG_LAYER_TYPE` enum to `DiagLayerType` and fixes a small copy-and-pasto in the writing code for communication parameters.

Since all patches are self-contained and simple but some are fairly big, I recommend reviewing this PR by looking at the individual commits instead of the aggregate changes of the whole PR.

### DIFF
--- a/examples/somersaultecu.py
+++ b/examples/somersaultecu.py
@@ -21,7 +21,7 @@ from odxtools.dataobjectproperty import DataObjectProperty
 from odxtools.diagcodedtypes import StandardLengthType
 from odxtools.diagdatadictionaryspec import DiagDataDictionarySpec
 from odxtools.diaglayer import DiagLayer, DiagLayerContainer
-from odxtools.diaglayertype import DIAG_LAYER_TYPE
+from odxtools.diaglayertype import DiagLayerType
 from odxtools.envdata import EnvironmentData
 from odxtools.envdatadesc import EnvironmentDataDescription
 from odxtools.functionalclass import FunctionalClass
@@ -1719,7 +1719,7 @@ somersault_diag_data_dictionary_spec = DiagDataDictionarySpec(
 
 # diagnostics layer
 somersault_diaglayer = DiagLayer(
-    variant_type=DIAG_LAYER_TYPE.BASE_VARIANT,
+    variant_type=DiagLayerType.BASE_VARIANT,
     odx_id=OdxLinkId("somersault", doc_frags),
     short_name="somersault",
     long_name="Somersault base variant",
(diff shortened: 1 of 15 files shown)


odxtools/diaglayertype.py
```diff
@@ -3,7 +3,7 @@
 from enum import Enum
 
 
-class DIAG_LAYER_TYPE(Enum):
+class DiagLayerType(Enum):
```
kayoub5, line 6: Breaking change?

kayoub5: Maybe make just a small PR only for rename

examples/somersaultecu.py
```diff
@@ -1745,13 +1747,13 @@ def extract_constant_bytes(params):
...
 somersault_lazy_diaglayer = DiagLayer(
-    variant_type=DIAG_LAYER_TYPE.ECU_VARIANT,
+    variant_type=DiagLayerType.ECU_VARIANT,
     odx_id=OdxLinkId("somersault_lazy", doc_frags),
     short_name="somersault_lazy",
     long_name="Somersault lazy ECU",
     description="<p>Sloppy variant of the somersault ECU (lazy &lt; assiduous)</p>",
     parent_refs=[
-        DiagLayer.ParentRef(  # <- TODO: this is a bit sketchy IMO
+        ParentRef(  # <- TODO: this is a bit sketchy IMO
```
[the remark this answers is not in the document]
andlaus: not really. I removed it, thanks.

odxtools/diaglayertype.py
andlaus: the problem with this is that in its current form, the patch needs the `ParentRef` class in a separate file (which is done in the first commit of this PR). This which implies quite a bit of additional work for IMO very limited benefits. do you want to have it done that way anyway?

odxtools/parentref.py
```diff
@@ -0,0 +1,127 @@
...
+            parent=parent_ref,
+            ref_type=ref_type,
+            not_inherited_diag_comms=not_inherited_diag_comms,
+            not_inherited_dops=not_inherited_dops,
+        )
+
+    def _resolve_references(self, odxlinks: OdxLinkDatabase):
+        self.parent_diag_layer = odxlinks.resolve(self.parent_ref)
+
+    def get_inheritance_priority(self):
```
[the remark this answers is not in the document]
andlaus: I think it makes it clearer what this method does: there could be other priorities, so `pref.get_priority()` is IMO slightly less descriptive than `pref.get_inheritance_priority()`? besides this, I intend to remove most (all?) of the methods which contain `_inherited_` in their names as part of the untangling PR series. (I've already implemented the intended result, but the resulting PR would be next to unreviewable.)

kayoub5 (approved):

PR #133, opened 2023-05-31
Untangle DiagLayer (2 of X)

andlaus:
This is the next PR in the quest to untangle the `DiagLayer` bowl of spaghetti. This time, the `DiagLayer` class is separated into `DiagLayerRaw` to represent the information of the XML and `DiagLayer` which provides the "logical" view on the diagnostic layer (i.e., it deals with inheritance et al) plus a few convenience functions.


odxtools/diaglayer.py
```diff
@@ -238,11 +152,23 @@ def data_object_properties(self) -> NamedItemList[DopBase]:
...
+    # </stuff subject to value inheritance>
+    #######
+
+    #######
+    # <communication parameters>
+    #######
+
     @property
     def communication_parameters(self) -> NamedItemList[CommunicationParameterRef]:
         """All communication parameters including inherited ones."""
```
kayoub5, line 165: If this class is a wrapper for "Diagnostic Layer" including "inherited Diagnostic Layers", then this class is no longer a "Layer", it would be more appropriate to call it a "Diagnostic Stack" (since it includes multiple layers)

odxtools/communicationparameter.py
```diff
@@ -76,6 +76,9 @@ def __repr__(self) -> str:
     def __str__(self) -> str:
         return self.__repr__()
 
+    def _build_odxlinks(self) -> Dict[OdxLinkId, Any]:
+        return {}
+
```
kayoub5, line 81: This PR is already too big, throwing in small random changes like this one is not helping

odxtools/diaglayer.py
```diff
@@ -97,132 +64,79 @@ def __init__(
...
-            diag_data_dictionary_spec = None
+        # Create DiagLayer
+        return DiagLayer(diag_layer_raw=diag_layer_raw)
+
+    #####
+    # <properties forwarded to the "raw" diag layer>
+    #####
+    @property
+    def variant_type(self) -> DiagLayerType:
+        return self.diag_layer_raw.variant_type
```
kayoub5, line 79: Maybe use class inheritance instead of those property wrappers ?

odxtools/diaglayer.py
```diff
@@ -273,74 +199,28 @@ def finalize_init(self, odxlinks: Optional[OdxLinkDatabase] = None):
...
-        if self.local_diag_data_dictionary_spec:
-            odxlinks.update(self.local_diag_data_dictionary_spec._build_odxlinks())
+        # we want to get the full diag layer, not just the raw layer
+        # when referencing...
+        result[self.odx_id] = self
 
-        return odxlinks
+        return result
 
     def _resolve_references(self, odxlinks: OdxLinkDatabase) -> None:
```
kayoub5, line 210: Looking at _resolve_references and _build_odxlinks, I am more convinced that you should just use class inheritance

odxtools/diaglayerraw.py
```diff
@@ -0,0 +1,283 @@
...
+    def _build_odxlinks(self) -> Dict[OdxLinkId, Any]:
+        """Construct a mapping from IDs to all objects that are contained in this diagnostic layer."""
+        odxlinks = {self.odx_id: self}
+
+        for obj in chain(
+            [self.admin_data],
+                self.company_datas,
+                self.functional_classes,
+            [self.diag_data_dictionary_spec],
+                self.diag_comms,
```
kayoub5, line 218: Indent does not look right here

odxtools/diaglayerraw.py
```diff
@@ -0,0 +1,283 @@
...
+                self.additional_audiences,
+                self.sdgs,
+                self.parent_refs,
+                self.communication_parameters,
+        ):
+            if obj is None or isinstance(obj, OdxLinkRef):
+                continue
+
+            if hasattr(obj, "odx_id"):
+                odxlinks[obj.odx_id] = obj
```
kayoub5, line 233: I don't get this part, what are you doing here?

odxtools/diaglayerraw.py
```diff
@@ -0,0 +1,283 @@
...
+                self.negative_responses,
+                self.global_negative_responses,
+                self.state_charts,
+                self.additional_audiences,
+                self.sdgs,
+                self.parent_refs,
+                self.communication_parameters,
+        ):
+            if obj is None or isinstance(obj, OdxLinkRef):
+                continue
```
kayoub5, line 230: Why would obj be an OdxLinkRef ?

tests/test_singleecujob.py
```diff
@@ -340,29 +340,43 @@ def test_default_lists(self):
...
+            admin_data=None,
+            company_datas=NamedItemList(short_name_as_id),
+            functional_classes=NamedItemList(short_name_as_id, [self.context.extensiveTask]),
             diag_data_dictionary_spec=None,
-            additional_audiences=[],
-            functional_classes=[],
-            state_charts=[],
+            diag_comms=[self.singleecujob_object],
+            requests=NamedItemList(short_name_as_id, []),
+            positive_responses=NamedItemList(short_name_as_id),
```
kayoub5, line 355: Is it `NamedItemList(short_name_as_id, [])` or `NamedItemList(short_name_as_id)` ?

odxtools/diaglayer.py
andlaus: while I don't disagree, the ODX spec calls the logical view on such an object -- i.e., including inheritance -- a "layer" (cf. section 7.3.2). I think we should only deviate from the spec if there is a *very* good reason for it...

odxtools/communicationparameter.py
andlaus: right. the issue with things like this is that you stumble over them while working on the actual subject of the PR, you fix it (maybe because it simplifies things elsewhere, in this case, this avoids needing to treat special treatment of the comparams in `DiagLayer._build_odxlinks()`) and then I promptly forget. I'll factor this out into a separate commit and open a PR for it. (NB: that the PR here is already an effort of getting a substantially larger patch into a more digestible size...)

andlaus: okay, the outsourced PR is #134. (i had to rebase this one on #134 to avoid a trivial cleanup PR after both have merged.)

odxtools/diaglayer.py
andlaus: I'm not so sure if this is a good idea. do it in a later PR? Once this saga is finished, this should be pretty easy,,,

odxtools/diaglayerraw.py
andlaus: I also noticed that. The problem is that this seems to be a bug in our style police (YaPF), and CI fails if YaPF reformats anything. (I would like to keep it this way to avoid larger formatting issues...)

I concluded that we have to live with this issue for now until it is fixed in YaPF...

odxtools/diaglayerraw.py
andlaus: some objects here (erhm, `diag_data_dictionary_spec`) do not have an `odx_id`,  but contain objects which have it, so their `_build_odxlinks()` method needs to be called. This issue can be avoided if it is guaranteed that all objects add themselfs in `_build_odxlinks()` if they exhibit `.odx_id`. I think almost all do, but I did not have the nerves to go hunting for the ones which don't. I'll have a look...

odxtools/diaglayerraw.py
andlaus: because `.diag_comms` can now contain `OdxLinkRefs`. (The specification mandates this. We currently have avoided to deal with this because the contents of the `DIAG-COMMS` tag where separated into services, single-ECU jobs and  references...)

tests/test_singleecujob.py
andlaus: these are synonyms. I'll change this to `NamedItemList(short_name_as_id)`. stay tuned.

andlaus: okay, done: ee14065b56a8815b51e76745340bf75af8a1a6b0 . (I know, this should go into a separate PR, but this would mean twice the work: Once for fixing `main`, once for fixing this PR...)

odxtools/diaglayerraw.py
kayoub5: so this is a hack that needs to be fixed? maybe add TODO?

andlaus: > I'll have a look...

Okay, done. the result is pretty large and I don't think this should be part of the present PR. I'll submit that cleanup once this PR is in...

kayoub5 (approved):

PR #135, opened 2023-06-02
give all objects full sovereignty over their `OdxLinkId`s

andlaus:
i.e., all `_build_odxlinks()` methods now add all OdxLinkIds including the id of the object for which it is called. This means that from the outside, all objects are now treated as opaque w.r.t. ODXLINK-able sub-objects. This PR emerged from the discussion of https://github.com/mercedes-benz/odxtools/pull/133 . Thanks to @kayoub5 for pointing this issue out!


examples/somersaultecu.py
```diff
@@ -523,6 +523,9 @@ class SomersaultSID(IntEnum):
             odx_id=OdxLinkId("somersault.multiplexer.flip_preference", doc_frags),
             short_name="flip_preference",
             long_name="Flip Preference",
+            description=None,
+            is_visible_raw=None,
+            sdgs=[],
```
andlaus, line 528: before you asked: The `DopBase` constructor had a default for the `sdgs` parameter, which lead to quite a few objects not being amendable using SDGs. Once more, I decided to do this as a "drive-by" fix.

The other two parameters are now necessary because this patches makes `DopBase` a dataclass which in turn became necessary because deriving a dataclass from a non-dataclass (`Multiplexer` was guilty of that attrocity) messes up the constructor arguments.

odxtools/companydata.py
```diff
@@ -179,15 +187,10 @@ def from_et(et_element, doc_frags: List[OdxDocFragment]) -> "CompanyData":
...
-
-            for tm in team_members.iterfind("TEAM-MEMBER"):
-                tml.append(TeamMember.from_et(tm, doc_frags))
-
-            team_members = tml
-
+        team_members = [
+            TeamMember.from_et(tm, doc_frags)
+            for tm in et_element.iterfind("TEAM-MEMBERS/TEAM-MEMBER")
+        ]
```
andlaus, line 193: this is also an unrelated cleanup, but IMO it is trivial enough not to warrant a separate PR...

odxtools/multiplexer.py
```diff
@@ -277,16 +283,18 @@ def convert_bytes_to_physical(self, decode_state: DecodeState, bit_position: int
         return mux_value, mux_next_byte
 
     def _build_odxlinks(self) -> Dict[OdxLinkId, Any]:
-        odxlinks = {}
-        odxlinks.update({self.odx_id: self})
+        odxlinks = super()._build_odxlinks()
+
         return odxlinks
```
kayoub5, line 288: Useless overloading

odxtools/envdata.py
```diff
@@ -49,18 +51,15 @@ def from_et(et_element, doc_frags: List[OdxDocFragment]) -> "EnvironmentData":
...
             dtc_values=dtc_values,
         )
 
     def _build_odxlinks(self) -> Dict[OdxLinkId, Any]:
-        odxlinks = super()._build_odxlinks()
-
-        odxlinks[self.odx_id] = self
-
-        return odxlinks
+        return super()._build_odxlinks()
```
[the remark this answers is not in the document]
andlaus: mainly because the class has a `_resolve_references()` method, so it's probably convenient to have it around. that said, in this case `_resolve_references()` also does basically nothing, so let's remove both. Thanks for the catch!

kayoub5 (approved):

kayoub5:
Squash or or merge commit?

andlaus:
the patches are pretty quite on-topic, so I'd go for a merge commit...

PR #136, opened 2023-06-05
clean up the `ParentRef` class

andlaus:
this cleans up the internals of the `ParentRef` class to make it consistent with the information provided by the XML, use (IMO) slightly more descriptive names, changes the inheritance priority of ECU-SHARED-DATA from the highest to lowest (i.e., everbody can derive from it instead of nobody) and converts the class to be a dataclass to make it clearer which attributes are supposed to be available and slightly reduce the amount of boiler-plate code.

In this context, note that the inheritance scheme which we implement is not 100% conformant with the specification: We allow a more specific (e.g., ECU-VARIANT) layer to inherit from any more generic layer (e.g., PROTOCOL), but the spec only allows to inherit from more generic layers that are once removed (i.e., BASE-VARIANT for ECU-VARIANT) plus from ECU-SHARED-DATA. Since we allow all inheritance possibilities of the spec (plus some more), I doubt that this will be a problem anywere except -- maybe -- for ODX editors build on top of odxtools...

### DIFF
--- a/examples/somersaultecu.py
+++ b/examples/somersaultecu.py
@@ -1792,14 +1792,18 @@ somersault_lazy_diaglayer_raw = DiagLayerRaw(
     sdgs=[],
     parent_refs=[
         ParentRef(
-            parent=OdxLinkRef.from_id(somersault_diaglayer.odx_id),
-            ref_type=DiagLayerType.BASE_VARIANT,
+            parent_layer_ref=OdxLinkRef.from_id(somersault_diaglayer.odx_id),
+            self_layer_type=DiagLayerType.ECU_VARIANT,
+            parent_layer_type=DiagLayerType.BASE_VARIANT,
             # this variant does not do backflips
             not_inherited_diag_comms=[
                 somersault_requests["backward_flips"].short_name,
                 somersault_requests["set_operation_params"].short_name,
             ],
             not_inherited_dops=[],
+            not_inherited_variables=[],
+            not_inherited_tables=[],
+            not_inherited_global_neg_responses=[],
         )
     ],
     communication_parameters=somersault_communication_parameters,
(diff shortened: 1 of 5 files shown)


odxtools/parentref.py
```diff
@@ -15,101 +14,106 @@
...
+    not_inherited_dops: List[str]  # short_name references
+    not_inherited_tables: List[str]  # short_name references
+    not_inherited_global_neg_responses: List[str]  # short_name references
+
+    def __post_init__(self) -> None:
+        # make sure that the layer from which we inherit are of lower
+        # priority than us.
+        self_prio = PRIORITY_OF_DIAG_LAYER_TYPE[self.self_layer_type]
+        parent_prio = PRIORITY_OF_DIAG_LAYER_TYPE[self.parent_layer_type]
+        assert self_prio > parent_prio, "diagnostic layers can only inherit from layers of lower priority"
```
kayoub5, line 41: Passing `self.layer_type` just for this assert?

kayoub5: It would be better to move the check to the DiagLayer

odxtools/parentref.py
```diff
@@ -15,101 +14,106 @@
...
+                                                               "NOT-INHERITED-GLOBAL-NEG-RESPONSE/"
+                                                               "GLOBAL-NEG-RESPONSE-SNREF")
         ]
-        ref_type = et_element.get(f"{xsi}type")
-        ref_type = ref_type[:-len("-REF")]
+
+        # determine the type of the referenced diag layer. for
+        # this, we need to strip the '-REF' suffix of the
+        # element's {xsi}type attribute
+        parent_layer_type_str = et_element.get(f"{xsi}type")
```
kayoub5, line 84: Sure all have `xsi:type` attribute?

andlaus: if I interpret the XSD correctly, these tags could be plain PARENT-REF, but then the XSD types would be useless. (I guess that's a minor bug in the spec.) anyway, I changed this to make it work unconditionally: 1a204e38faf90f3082b3eb5fb4354797681a5fbf

odxtools/parentref.py
```diff
@@ -27,6 +26,10 @@ class ParentRef:
     def parent_layer(self) -> "DiagLayer":
         return self._parent_layer
 
+    @property
+    def parent_layer_type(self) -> DiagLayerType:
```
kayoub5, line 30: is this property really needed? parent_layer is available after all

examples/somersaultecu.py
```diff
@@ -2012,7 +2011,6 @@ def extract_constant_bytes(params):
     parent_refs=[
         ParentRef(
             parent_layer_ref=OdxLinkRef.from_id(somersault_diaglayer.odx_id),
-            self_layer_type=DiagLayerType.ECU_VARIANT,
             parent_layer_type=DiagLayerType.BASE_VARIANT,
```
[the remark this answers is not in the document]
andlaus: not really, so I removed it: 730518f3be143aa549714561c3611b0cd8d20b5a . (Note that I think it also does not improve things...)

odxtools/parentref.py
andlaus: correct. removed...

kayoub5 (approved):

andlaus:
ok, let's get over this hump. thanks @kayoub5 !

PR #137, opened 2023-06-07
refactor the initialization and reference resolution mechanism

andlaus:
unfortunately, this PR firmly belongs into the "long and boring" category:

It cleans up the object initialization code in the sense that the API is now much more consistent; Now, all non-primitive objects which are part of a diagnostic layer provide a `._build_odxlinks()` method to create a dictionary of all `OdxLinkId` mappings contained by this object. These dictionaries are then collected for the whole database and used to resolve ODXLINK references via the `._resolve_odxlinks(odxlinks)` methods. Once this is done, the diagnostic layers deal with inheritance of objects, whereafter the short name references (SNREFs) are resolved via the `._resolve_snrefs(diag_layer)` methods.
    
The initialization of objects which are not part of any `DiagLayer` is slightly different: The `Database` object has a `refresh()` method which deals with the full initializtion logic of all objects it contains, the `DiagLayerContainer` and `DiagLayer` classes have `_build_odxlinks()` and `_finalize_init()` methods. `_finalize_init()` deals with inheritance and SNREF resolution for all objects contained.

### DIFF
--- a/tests/test_decoding.py
+++ b/tests/test_decoding.py
@@ -452,7 +452,6 @@ class TestDecoding(unittest.TestCase):
 
     def test_decode_request_structure(self):
         """Test the decoding for a structure."""
-        odxlinks = OdxLinkDatabase()
         diag_coded_type = StandardLengthType(
             base_data_type="A_UINT32",
             base_type_encoding=None,
@@ -483,7 +482,6 @@ class TestDecoding(unittest.TestCase):
             unit_ref=None,
             sdgs=[],
         )
-        odxlinks.update({dop.odx_id: dop})
 
         req_param1 = CodedConstParameter(
             short_name="SID",
@@ -530,7 +528,6 @@ class TestDecoding(unittest.TestCase):
             parameters=[struct_param1, struct_param2],
             byte_size=None,
         )
-        odxlinks.update({struct.odx_id: struct})
         req_param2 = ValueParameter(
             short_name="structured_param",
             long_name=None,
@@ -554,7 +551,6 @@ class TestDecoding(unittest.TestCase):
             parameters=[req_param1, req_param2],
             byte_size=None,
         )
-        odxlinks.update({req.odx_id: req})
         service = DiagService(
             odx_id=OdxLinkId("service_id", doc_frags),
             short_name="service_sn",
@@ -580,7 +576,17 @@ class TestDecoding(unittest.TestCase):
             admin_data=None,
             company_datas=NamedItemList(short_name_as_id),
             functional_classes=NamedItemList(short_name_as_id),
-            diag_data_dictionary_spec=None,
+            diag_data_dictionary_spec=DiagDataDictionarySpec(
+                dtc_dops=NamedItemList(short_name_as_id),
+                data_object_props=NamedItemList(short_name_as_id, [dop]),
+                structures=NamedItemList(short_name_as_id, [struct]),
+                end_of_pdu_fields=NamedItemList(short_name_as_id),
+                tables=NamedItemList(short_name_as_id),
+                env_data_descs=NamedItemList(short_name_as_id),
+                env_datas=NamedItemList(short_name_as_id),
+                muxs=NamedItemList(short_name_as_id),
+                unit_spec=None,
+                sdgs=[]),
             diag_comms=[service],
             requests=NamedItemList(short_name_as_id, [req]),
             positive_responses=NamedItemList(short_name_as_id),
@@ -595,13 +601,10 @@ class TestDecoding(unittest.TestCase):
             ecu_variant_patterns=[],
         )
         diag_layer = DiagLayer(diag_layer_raw=diag_layer_raw)
+        odxlinks = OdxLinkDatabase()
+        odxlinks.update(diag_layer._build_odxlinks())
         diag_layer._resolve_odxlinks(odxlinks)
         diag_layer._finalize_init(odxlinks)
-
-        req_param1._resolve_odxlinks(odxlinks)
-        req_param2._resolve_odxlinks(odxlinks)
-        struct_param1._resolve_odxlinks(odxlinks)
-        struct_param2._resolve_odxlinks(odxlinks)
 
         coded_message = bytes([0x12, 0x34])
         expected_message = Message(
@@ -624,7 +627,6 @@ class TestDecoding(unittest.TestCase):
 
     def test_decode_request_end_of_pdu_field(self):
         """Test the decoding for a structure."""
-        odxlinks = OdxLinkDatabase()
         diag_coded_type = StandardLengthType(
             base_data_type="A_UINT32",
             base_type_encoding=None,
@@ -644,7 +646,7 @@ class TestDecoding(unittest.TestCase):
 
         compu_method = IdenticalCompuMethod(internal_type="A_INT32", physical_type="A_INT32")
         dop = DataObjectProperty(
-            odx_id=OdxLinkId("dop.odx_id", doc_frags),
+            odx_id=OdxLinkId("dop.id", doc_frags),
             short_name="dop_sn",
             long_name=None,
             description=None,
@@ -655,7 +657,6 @@ class TestDecoding(unittest.TestCase):
             unit_ref=None,
             sdgs=[],
         )
-        odxlinks.update({dop.odx_id: dop})
 
         req_param1 = CodedConstParameter(
             short_name="SID",
@@ -702,7 +703,6 @@ class TestDecoding(unittest.TestCase):
             parameters=[struct_param1, struct_param2],
             byte_size=None,
         )
-        odxlinks.update({struct.odx_id: struct})
         eopf = EndOfPduField(
             odx_id=OdxLinkId("eopf_id", doc_frags),
             short_name="eopf_sn",
@@ -717,8 +717,6 @@ class TestDecoding(unittest.TestCase):
             max_number_of_items=None,
             is_visible_raw=True,
         )
-        odxlinks.update({eopf.odx_id: eopf})
-
         req_param2 = ValueParameter(
             short_name="eopf_param",
             long_name=None,
@@ -742,7 +740,6 @@ class TestDecoding(unittest.TestCase):
             parameters=[req_param1, req_param2],
             byte_size=None,
         )
-        odxlinks.update({req.odx_id: req})
         service = DiagService(
             odx_id=OdxLinkId("service_id", doc_frags),
             short_name="service_sn",
@@ -768,7 +765,17 @@ class TestDecoding(unittest.TestCase):
             admin_data=None,
             company_datas=NamedItemList(short_name_as_id),
             functional_classes=NamedItemList(short_name_as_id),
-            diag_data_dictionary_spec=None,
+            diag_data_dictionary_spec=DiagDataDictionarySpec(
+                dtc_dops=NamedItemList(short_name_as_id),
+                data_object_props=NamedItemList(short_name_as_id, [dop]),
+                structures=NamedItemList(short_name_as_id, [struct]),
+                end_of_pdu_fields=NamedItemList(short_name_as_id, [eopf]),
+                tables=NamedItemList(short_name_as_id),
+                env_data_descs=NamedItemList(short_name_as_id),
+                env_datas=NamedItemList(short_name_as_id),
+                muxs=NamedItemList(short_name_as_id),
+                unit_spec=None,
+                sdgs=[]),
             diag_comms=[service],
             requests=NamedItemList(short_name_as_id, [req]),
             positive_responses=NamedItemList(short_name_as_id),
@@ -783,15 +790,10 @@ class TestDecoding(unittest.TestCase):
             ecu_variant_patterns=[],
         )
         diag_layer = DiagLayer(diag_layer_raw=diag_layer_raw)
+        odxlinks = OdxLinkDatabase()
+        odxlinks.update(diag_layer._build_odxlinks())
         diag_layer._resolve_odxlinks(odxlinks)
         diag_layer._finalize_init(odxlinks)
-
-        eopf._resolve_odxlinks(odxlinks)
-        struct_param2._resolve_odxlinks(odxlinks)
-        req_param2._resolve_odxlinks(odxlinks)
-        req._resolve_odxlinks(odxlinks)
-        service._resolve_odxlinks(odxlinks)
-        diag_layer._resolve_odxlinks(odxlinks)
 
         coded_message = bytes([0x12, 0x34, 0x34])
         expected_message = Message(
@@ -1065,12 +1067,6 @@ class TestDecoding(unittest.TestCase):
             byte_size=None,
         )
 
-        odxlinks = OdxLinkDatabase()
-        odxlinks.update({
-            req.odx_id: req,
-            pos_response.odx_id: pos_response,
-            neg_response.odx_id: neg_response
-        })
         service = DiagService(
             odx_id=OdxLinkId("service_id", doc_frags),
             short_name="service_sn",
@@ -1111,6 +1107,8 @@ class TestDecoding(unittest.TestCase):
             ecu_variant_patterns=[],
         )
         diag_layer = DiagLayer(diag_layer_raw=diag_layer_raw)
+        odxlinks = OdxLinkDatabase()
+        odxlinks.update(diag_layer._build_odxlinks())
         diag_layer._resolve_odxlinks(odxlinks)
         diag_layer._finalize_init(odxlinks)
 
@@ -1164,8 +1162,6 @@ class TestDecoding(unittest.TestCase):
             is_temporary_raw=None,
             sdgs=[],
         )
-        dtcs = [dtc1, dtc2]
-        odxlinks.update({dtc1.odx_id: dtc1, dtc2.odx_id: dtc2})
         dop = DtcDop(
             odx_id=OdxLinkId("dtc.dop.odx_id", doc_frags),
             short_name="dtc_dop_sn",
@@ -1176,11 +1172,11 @@ class TestDecoding(unittest.TestCase):
             physical_type=PhysicalType(DataType.A_UINT32, display_radix=None, precision=None),
             compu_method=compu_method,
             unit_ref=None,
-            dtcs_raw=dtcs,
+            dtcs_raw=[dtc1, dtc2],
             is_visible_raw=True,
             sdgs=[],
         )
-        odxlinks.update({dop.odx_id: dop})
+        odxlinks.update(dop._build_odxlinks())
         resp_param1 = CodedConstParameter(
             short_name="SID",
             long_name=None,
@@ -1215,10 +1211,9 @@ class TestDecoding(unittest.TestCase):
             byte_size=None,
             response_type="POS-RESPONSE",
         )
+        odxlinks.update(pos_response._build_odxlinks())
 
         dop._resolve_odxlinks(odxlinks)
-        resp_param1._resolve_odxlinks(odxlinks)
-        resp_param2._resolve_odxlinks(odxlinks)
         pos_response._resolve_odxlinks(odxlinks)
 
         coded_message = bytes([0x12, 0x34])
@@ -1252,7 +1247,7 @@ class TestDecodingAndEncoding(unittest.TestCase):
             sdgs=[],
         )
         dop = self.dop_bytes_termination_end_of_pdu
-        odxlinks.update({dop.odx_id: dop})
+        odxlinks.update(dop._build_odxlinks())
         self.parameter_termination_end_of_pdu = ValueParameter(
             short_name="min_max_parameter",
             long_name=None,
@@ -1325,7 +1320,7 @@ class TestDecodingAndEncoding(unittest.TestCase):
             parameters=[struct_param],
             byte_size=None,
         )
-        odxlinks.update({structure.odx_id: structure})
+        odxlinks.update(structure._build_odxlinks())
 
         req_param1 = self.parameter_sid
         req_param2 = ValueParameter(
@@ -1400,7 +1395,7 @@ class TestDecodingAndEncoding(unittest.TestCase):
             unit_ref=None,
             sdgs=[],
         )
-        odxlinks.update({dop.odx_id: dop})
+        odxlinks.update(dop._build_odxlinks())
         req_param1 = CodedConstParameter(
             short_name="SID",
             long_name=None,
--- a/tests/test_encoding.py
+++ b/tests/test_encoding.py
@@ -123,9 +123,9 @@ class TestEncodeRequest(unittest.TestCase):
             internal_upper_limit=None,
         )
         dop = DataObjectProperty(
-            odx_id=OdxLinkId("dop-odx_id", doc_frags),
-            short_name="example dop",
-            long_name=None,
+            odx_id=OdxLinkId("dop.id", doc_frags),
+            short_name="dop_sn",
+            long_name="example dop",
             description=None,
             is_visible_raw=None,
             diag_coded_type=diag_coded_type,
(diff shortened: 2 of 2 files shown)


odxtools/admindata.py
```diff
@@ -51,15 +54,19 @@ def _build_odxlinks(self) -> Dict[OdxLinkId, Any]:
...
 
         self._team_member: Optional[TeamMember] = None
         if self.team_member_ref is not None:
             self._team_member = odxlinks.resolve(self.team_member_ref, TeamMember)
 
         for sdg in self.sdgs:
-            sdg._resolve_references(odxlinks)
+            sdg._resolve_odxlinks(odxlinks)
+
+    def _resolve_snrefs(self, diag_layer: "DiagLayer"):
```
kayoub5, line 67: Why not just one function that takes a DiagLayer and OdxLinkDatabase as arguments?

odxtools/admindata.py
```diff
@@ -51,15 +54,19 @@ def _build_odxlinks(self) -> Dict[OdxLinkId, Any]:
 
         return result
 
-    def _resolve_references(self, odxlinks: OdxLinkDatabase):
+    def _resolve_odxlinks(self, odxlinks: OdxLinkDatabase):
```
kayoub5, line 57: You really make it hard for me to review with rename/refactor in same PR.

odxtools/admindata.py
```diff
@@ -74,6 +81,15 @@ def from_et(et_element: ElementTree.Element, doc_frags: List[OdxDocFragment]) ->
...
         return Modification(change=change, reason=reason)
 
+    def _build_odxlinks(self) -> Dict[OdxLinkId, Any]:
+        return {}
+
+    def _resolve_odxlinks(self, odxlinks: OdxLinkDatabase):
+        pass
+
+    def _resolve_snrefs(self, diag_layer: "DiagLayer"):
+        pass
```
kayoub5, line 91: Any reason why would this class have references in the future?

kayoub5: Even _build_odxlinks is emtpy

odxtools/audience.py
```diff
@@ -35,7 +38,10 @@ def from_et(et_element, doc_frags: List[OdxDocFragment]) -> "AdditionalAudience"
     def _build_odxlinks(self) -> Dict[OdxLinkId, Any]:
         return {self.odx_id: self}
 
-    def _resolve_references(self, odxlinks: OdxLinkDatabase) -> None:
+    def _resolve_odxlinks(self, odxlinks: OdxLinkDatabase) -> None:
+        pass
+
+    def _resolve_snrefs(self, diag_layer: "DiagLayer") -> None:
         pass
```
kayoub5, line 45: This block seems to repeat everywhere, maybe move it to a common parent class ?

```
        return {self.odx_id: self}

    def _resolve_references(self, odxlinks: OdxLinkDatabase) -> None:
    def _resolve_odxlinks(self, odxlinks: OdxLinkDatabase) -> None:
        pass

        pass
```

odxtools/dataobjectproperty.py
```diff
@@ -386,16 +410,15 @@ def convert_physical_to_bytes(self, physical_value, encode_state, bit_position):
     def _build_odxlinks(self) -> Dict[OdxLinkId, Any]:
         odxlinks = super()._build_odxlinks()
         odxlinks[self.odx_id] = self
```
kayoub5, line 412: `self.odx_id` can be None (according to typehint)

odxtools/diagdatadictionaryspec.py
```diff
@@ -152,29 +152,24 @@ def _build_odxlinks(self) -> Dict[OdxLinkId, Any]:
...
+
+        for obj in chain(self.data_object_props, self.dtc_dops, self.end_of_pdu_fields,
+                         self.env_data_descs, self.env_datas, self.muxs, self.sdgs, self.structures,
+                         self.tables):
+            obj._resolve_odxlinks(odxlinks)
+
+        if self.unit_spec:
+            self.unit_spec._resolve_odxlinks(odxlinks)
+
+    def _resolve_snrefs(self, parent_dl: "DiagLayer") -> None:
```
kayoub5, line 165: Keep argument name consistent across implementations

```suggestion
    def _resolve_snrefs(self, diag_layer: "DiagLayer") -> None:
```

odxtools/diaglayer.py
```diff
@@ -207,10 +193,26 @@ def _build_odxlinks(self) -> Dict[OdxLinkId, Any]:
...
         return result
 
-    def _resolve_references(self, odxlinks: OdxLinkDatabase) -> None:
+    def _resolve_odxlinks(self, odxlinks: OdxLinkDatabase) -> None:
         """Recursively resolve all references."""
 
-        self.diag_layer_raw._resolve_references(self, odxlinks)
+        self.diag_layer_raw._resolve_odxlinks(odxlinks)
+
+    def _finalize_init(self, odxlinks: OdxLinkDatabase) -> None:
```
kayoub5, line 201: Not `_resolve_snrefs` like the others?

odxtools/multiplexer.py
```diff
@@ -120,14 +135,20 @@ def from_et(et_element, doc_frags: List[OdxDocFragment]) -> "MultiplexerSwitchKe
             dop_ref=dop_ref,
         )
 
-    def _resolve_references(self, odxlinks: OdxLinkDatabase) -> None:
+    def _build_odxlinks(self) -> Dict[OdxLinkId, Any]:
+        return {}
```
kayoub5, line 139: Why have the function if not going to implement it?

odxtools/nameditemlist.py
```diff
@@ -63,6 +63,9 @@ def append(self, item: T):
     def sort(self, key=None, reverse=False):
         return self._list.sort(key=key, reverse=reverse)
 
+    def __contains__(self, x) -> bool:
+        return x in self._list
```
kayoub5, line 67: Looks a bit unrelated to PR

odxtools/parameters/parameterwithdop.py
```diff
@@ -28,31 +28,31 @@ def __init__(
...
-                logger.info(f"Param {self.short_name} could not resolve DOP-REF {self.dop_ref}")
-            else:
-                self._dop = dop
-        else:
-            logger.warn(f"Param {self.short_name} without DOP-(SN)REF should not exist!")
+            self._dop = parent_dl.data_object_properties[self.dop_snref]
+
+    @property
+    def dop(self) -> Optional[DopBase]:
+        """may be a DataObjectProperty, a Structure or None"""
```
kayoub5, line 53: So `Union[DataObjectProperty, Structure, None]` would be more accurate?

odxtools/parameters/tablekeyparameter.py
```diff
@@ -19,6 +20,41 @@ def __init__(self, *, odx_id, table_ref, table_snref, table_row_snref, table_row
...
+
+        return result
+
+    def _resolve_odxlinks(self, odxlinks: OdxLinkDatabase) -> None:
+        super()._resolve_odxlinks(odxlinks)
+
+        if self.table_ref:
+            self._table = odxlinks.resolve(self.table_ref)
+        if self.table_row_ref:
+            self._table_row = odxlinks.resolve(self.table_row_ref)
```
kayoub5, line 40: resolve does not accept None and return None?

odxtools/parameters/tablekeyparameter.py
```diff
@@ -19,6 +20,41 @@ def __init__(self, *, odx_id, table_ref, table_snref, table_row_snref, table_row
...
+
+    def _resolve_snrefs(self, parent_dl: "DiagLayer") -> None:
+        super()._resolve_snrefs(parent_dl)
+
+        if self.table_snref:
+            self._table = parent_dl.local_diag_data_dictionary_spec.tables[self.table_snref]
```
kayoub5, line 46: Only the local ones?

odxtools/service.py
```diff
@@ -55,6 +58,7 @@ def __init__(
         self.long_name: Optional[str] = long_name
         self.description: Optional[str] = description
         self.semantic: Optional[str] = semantic
+        self.admin_data: Optional[AdminData] = admin_data
```
kayoub5, line 61: Existing bug unrelated to PR?

odxtools/singleecujob.py
```diff
@@ -59,9 +63,15 @@ def from_et(et_element, doc_frags: List[OdxDocFragment]) -> "InputParam":
             oid=oid,
         )
 
-    def _resolve_references(self, odxlinks: OdxLinkDatabase) -> None:
+    def _build_odxlinks(self) -> Dict[OdxLinkId, Any]:
+        return {}
```
kayoub5, line 67: Sure this guy have no odx_id?

odxtools/singleecujob.py
```diff
@@ -106,9 +116,15 @@ def from_et(et_element, doc_frags: List[OdxDocFragment]) -> "OutputParam":
...
+
+    def _resolve_odxlinks(self, odxlinks: OdxLinkDatabase) -> None:
         self._dop = odxlinks.resolve(self.dop_base_ref)
 
+    def _resolve_snrefs(self, diag_layer: "DiagLayer") -> None:
+        pass
```
kayoub5, line 126: Duplicate with lines 66-73

odxtools/singleecujob.py
```diff
@@ -142,9 +158,15 @@ def from_et(et_element, doc_frags: List[OdxDocFragment]) -> "NegOutputParam":
...
```
kayoub5, line 168: Duplicate with lines 66-73

odxtools/statechart.py
```diff
@@ -91,3 +91,14 @@ def _resolve_references(self, odxlinks: OdxLinkDatabase) -> None:
...
+    def _resolve_snrefs(self, diag_layer: "DiagLayer") -> None:
+        for st in self.states:
+            st._resolve_snrefs(diag_layer)
+
+        for strans in self.state_transitions:
+            # note that the signature of the state transition's
+            # _resolve_snrefs() method is non-standard as the
+            # namespace of these SNREFs is the state chart, not the
+            # whole diag layer...
+            strans._resolve_snrefs(diag_layer, states=self.states)
```
kayoub5, line 104: Why the extra states argument?

tests/test_decoding.py
```diff
@@ -287,7 +288,10 @@ def test_decode_request_coded_const(self):
             ecu_variant_patterns=[],
         )
         diag_layer = DiagLayer(diag_layer_raw=diag_layer_raw)
-        diag_layer.finalize_init(odxlinks=odxlinks)
+        odxlinks = OdxLinkDatabase()
+        odxlinks.update(diag_layer._build_odxlinks())
+        diag_layer._resolve_odxlinks(odxlinks)
+        diag_layer._finalize_init(odxlinks)
```
kayoub5, line 294: Not possible to merge _resolve_odxlinks and _finalize_init if both will always be called together?

tests/test_decoding.py
```diff
@@ -590,12 +595,13 @@ def test_decode_request_structure(self):
...
 
-        req_param1._resolve_references(diag_layer, odxlinks)
-        req_param2._resolve_references(diag_layer, odxlinks)
-        struct_param1._resolve_references(diag_layer, odxlinks)
-        struct_param2._resolve_references(diag_layer, odxlinks)
+        req_param1._resolve_odxlinks(odxlinks)
+        req_param2._resolve_odxlinks(odxlinks)
+        struct_param1._resolve_odxlinks(odxlinks)
+        struct_param2._resolve_odxlinks(odxlinks)
```
kayoub5, line 604: This is still required even after `diag_layer._resolve_odxlinks(odxlinks)` ?

tests/test_decoding.py
```diff
@@ -833,8 +838,8 @@ def test_decode_request_linear_compu_method(self):
             is_highlow_byte_order_raw=None,
         )
         dop = DataObjectProperty(
-            odx_id=OdxLinkId("linear.dop.odx_id", doc_frags),
-            short_name="linear.dop.sn",
+            odx_id=OdxLinkId("linear.dop.id", doc_frags),
+            short_name="linear_dop_sn",
```
kayoub5, line 842: Why the change of short_name?

kayoub5:
Still don't see the reason why you split _resolve_odxlinks and _resolve_snrefs, I believe it was better as one function

andlaus:
> Still don't see the reason why you split _resolve_odxlinks and _resolve_snrefs, I believe it was better as one function

that's because the SNREFS can only be resolved once the respective namespace for the short names is complete, i.e. once the diaglayer has sorted out inheritance of the objects that are subject to value inheritance and that's only really possible after the ODXLINKs have been resolved. Maybe the spec can be interpreted differently, (i.e., SNREFS could be only allowed for locally defined objects), but then stuff like the `NOT-INHERITED-*` tags of `PARENT-REF` would not make sense...

Besides this the current code is pretty hacky and fragile, because the `_resolve_references()` signature is different for objects that need to resolve short name references compared to the ones which don't. this becomes even more fun  once class inheritance gets involved: sometimes the base class uses the variant without the `DiagLayer` parameter and the derived class the one with it. This happens e.g., for `EndOfPduField`. tl;dr: object initialization currently is a mess!

tests/test_decoding.py
andlaus: my plan is to get rid of the `odxlinks` argument for `finalize_init` in a follow-up PR, but that would have led to even more extensive changes here...

odxtools/singleecujob.py
andlaus: for (probably) historic reasons, this file contains several classes. Line 66 is `InputParam`, this here is for `NegOutputParam`. This should be fixed in a future PR. (all files should be made to contain at most a single class; I also get annoyed by this quite frequently.)

odxtools/singleecujob.py
andlaus: I just looked it up again: `INPUT-PARAM` tags come with an optional `OID` attribute, but for some strange reason without an `ID`. (strangely enough, for `OUTPUT-PARAM` the `ID` attribute is mandatory. looks like an oversight in the spec to me...)

odxtools/service.py
andlaus: yes, there might be a few more such minor fixes lurking. I can make these separate PRs or commits if you want (and we can find them)...

odxtools/audience.py
andlaus: I'm not really sure if this is a good idea: having these methods around should reduce copy-and-paste mistakes because these functions are more visible, and also the `_build_odxlinks()` sample implementation of above snipped is only valid for classes which do not have any complex sub-objects but which an ID.

odxtools/companydata.py
```diff
@@ -23,16 +26,16 @@ class XDoc:
...
-        short_name = xdoc.findtext("SHORT-NAME")
-        long_name = xdoc.findtext("LONG-NAME")
-        description = create_description_from_et(xdoc.find("DESC"))
-        number = xdoc.findtext("NUMBER")
-        state = xdoc.findtext("STATE")
-        date = xdoc.findtext("DATE")
-        publisher = xdoc.findtext("PUBLISHER")
-        url = xdoc.findtext("URL")
-        position = xdoc.findtext("POSITION")
+    def from_et(et_element) -> "XDoc":
```
andlaus, line 29: BTW: another "drive-by" fix...

odxtools/admindata.py
kayoub5: That would be correct if any ODX Link resolution would cause a change in SNREF resolution, but I don't see how that could happen.

odxtools/admindata.py
kayoub5: The idea is that this looks like a "simple class", that don't need to participate in the resolution logic

odxtools/service.py
kayoub5: That would be better, since it would make for a cleaner git history, and easier to make changelog

odxtools/diagdatadictionaryspec.py
andlaus: good point. fixed: 38159785e3265f46dc77c02c6bfa4717776f5b5c

odxtools/diaglayer.py
andlaus: this is the diaglayer. it calls `_resolve_snrefs()` for all its sub-objects...

odxtools/multiplexer.py
andlaus: uniform API: you do not need to handle any sub-objects differently, regardless of whether they have an ID or not, need to resolve ODXLINKs or not or need to resolve SNREFs or not...

odxtools/nameditemlist.py
andlaus: true. drive-by fix. I think this one was caused by changing an ordinary list to a `NamedItemList` somewhere and the `in` operator was used on that list (a la `foo in bar`)

odxtools/parameters/tablekeyparameter.py
andlaus: well, the purpose of these methods is to resolve the internally contained references of the object, i.e. it *modifies* the object (akin `list.sort()`)

odxtools/parameters/tablekeyparameter.py
andlaus: I hope this corresponds to what's currently in `main`. doing this properly requires refactoring the value inheritance mechanism of diaglayers. (PR for this is in the pipeline, but doing this too here would be way too much IMO.)

tests/test_decoding.py
andlaus: I forgot to actual answer: `_finalize_init()` conceptually has not much to do with the ODXLINK mechanism (the current need for the `odxlinks` argument for `finalize_init()` is unfortunate, see comment above).

tl;dr: it's possible but its IMO not a terribly good idea.

odxtools/admindata.py
andlaus: for example by not inheriting a DIAG-COMM that is only referenced by the base layer...

odxtools/parameters/tablekeyparameter.py
kayoub5: I am asking if the "if self.table_row_ref:" is really necessary

kayoub5:
> > Still don't see the reason why you split _resolve_odxlinks and _resolve_snrefs, I believe it was better as one function
> 
> that's because the SNREFS can only be resolved once the respective namespace for the short names is complete, i.e. once the diaglayer has sorted out inheritance of the objects that are subject to value inheritance and that's only really possible after the ODXLINKs have been resolved. Maybe the spec can be interpreted differently, (i.e., SNREFS could be only allowed for locally defined objects), but then stuff like the `NOT-INHERITED-*` tags of `PARENT-REF` would not make sense...
> 
> Besides this the current code is pretty hacky and fragile, because the `_resolve_references()` signature is different for objects that need to resolve short name references compared to the ones which don't. this becomes even more fun once class inheritance gets involved: sometimes the base class uses the variant without the `DiagLayer` parameter and the derived class the one with it. This happens e.g., for `EndOfPduField`. tl;dr: object initialization currently is a mess!

this means _resolve_references should receive a dict(short_name => object) instead of directly receiving diag layer, or class specific extra arguments

odxtools/singleecujob.py
kayoub5: so maybe those lines need to go to a parent Param class?

odxtools/audience.py
kayoub5: to reduce copy-paste mistakes simply make a unit test that:
* checks that every ID attribute in sample odx XML end up registered in the odx link database.
* checks resolve gets called for each ID-REF in the odx xml (ny providing a mocked link database, and checking the call count)

odxtools/parameters/tablekeyparameter.py
andlaus: ah okay, my bad. the signature of this method currently is [`def resolve(self, ref: OdxLinkRef, expected_type: Optional[Type[T]] = None) -> Any`](https://github.com/mercedes-benz/odxtools/blob/main/odxtools/odxlink.py#L180). So it currently does not accept `None`, but you're right: it might avoid a few case distinctions if it did. On the flipside, it might come with a few more implicit assumptions which one must consider while reading code that uses it. I suppose I have to meditate a bit about it before I can form an opinion...

andlaus:
> this means _resolve_references should receive a dict(short_name => object) instead of directly receiving diag layer, or class specific extra arguments

the problem here is that the namespace for these short name references depends on the type of object which specifies the reference. e.g. the for the `diag_comms`, the list of relevant objects is `diag_layer.diag_comms`, for the `dops` it is `diag_layer.data_object_props` and so on. Since there can -- and likely will -- be naming conflicts between the lists, such a simple mechanism is unfortunately not possible. (I don't know yet how SNPATHREFS can be implemented, but it fortunately seems like these are extremely uncommon...)

odxtools/singleecujob.py
andlaus: that might be a possibility for a future PR. I'm not really sure if this a terribly good idea though, because it makes things quite a bit less explicit, so I'd like to have a separate PR to assess that...

odxtools/audience.py
andlaus: that's exactly the problem: while I would *love* to have such extensive test coverage, we currently still have a very long way to go to get there ...

kayoub5:
> > this means _resolve_references should receive a dict(short_name => object) instead of directly receiving diag layer, or class specific extra arguments
> 
> the problem here is that the namespace for these short name references depends on the type of object which specifies the reference. e.g. the for the `diag_comms`, the list of relevant objects is `diag_layer.diag_comms`, for the `dops` it is `diag_layer.data_object_props` and so on. Since there can -- and likely will -- be naming conflicts between the lists, such a simple mechanism is unfortunately not possible. (I don't know yet how SNPATHREFS can be implemented, but it fortunately seems like these are extremely uncommon...)

That means a SNREF database similar to the OdxLinkDatabase, with a `resolve(self, short_name: str, expected_type: Optional[Type[T]] = None)` and the internal storage with a dict key = (name, type)
This would be similar to how the OdxLinkDatabase have a storage with dict key = (doc_frag, odx_id)

I am not asking for this change to be done in this PR (it's already big as it is), I am asking for the concept to be considered for future PRs

odxtools/admindata.py
kayoub5: ODX Link uses doc REF to decided what to actually use (by default they always use same document unless there is a DOC-REF), I don't think they are meant to be affected by diag layer inheritance.

andlaus:
> I am not asking for this change to be done in this PR (it's already big as it is), I am asking for the concept to be considered for future PRs

I'd be very open for such considerations, I just have some trouble imagining how this would work. (thus, do not expect too much from my side in this direction ;))

odxtools/admindata.py
andlaus: yes, but if you have e.g., an ECU-INSTANCE layer `ECU` which does not inherit the diag-comm `FOO` from its parent BASE-VARIANT `BASE` and `BASE` references `FOO` from somewhere else instead of defining it, you must have resolved `FOO` in `BASE` before being able to determine which diag comms `ECU` ought to inherit... (if you ask me, the spec is way too complicated and I doubt that it is implemented 100% correctly by *any* software package...)

andlaus:
okay: I moved most unrelated stuff into #141. Hopefully makes this PR more focused (since not doing so would create quite a few merge conflicts, this PR is based on top of #141, so until #141 is merged, the diff here will stay unchanged...)

tests/test_decoding.py
andlaus: good catch. fixed: f8f1ab758c576d9319fe015aa8e4839a1d1ecf8d

andlaus:
@kayoub5: do you see any major issues with this?

kayoub5:
> @kayoub5: do you see any major issues with this?

Looks good, but you will need to bump the major version once you are done with all the refactoring

A lot of API changes that could break things for users.

andlaus:
>  Looks good, but you will need to bump the major version once you are done with all the refactoring

alright, I'll do this. let's get this merged, thanks for your patience...

PR #142, opened 2023-06-12
diaglayer: refactor the communication parameter handling

andlaus:
This continues the long quest to clean up `DiagLayer`:

it turns out that the names of the communication parameters are not unambiguous: e.g., "CP_UniqueRespIdTable" parameter is not so unique in the sense that it exists for CAN (comparam subset "ISO_15765_2") as well as for DoIP (CP subset "ISO_13400_2_DIS_2015") and both parameters have a completely different definition. (The same issue applies to e.g., "CP_TesterPresentTime", but this being a simple integer parameter, at least the returned value is of the same type.)

To get out of that pickle, the full parameter ID now has to be specified for `DiagLayer.get_communication_parameter()` (e.g., "ISO_15765_2.CP_UniqueRespIdTable").

### DIFF
--- a/odxtools/comparam_subset.py
+++ b/odxtools/comparam_subset.py
@@ -108,6 +108,7 @@ class ComplexComparam(BaseComparam):
             if cp_el.tag in ("COMPARAM", "COMPLEX-COMPARAM"):
                 self.comparams.append(create_any_comparam_from_et(cp_el, doc_frags))
 
+        self.complex_physical_default_value = None
         if cpdv_elem := et_element.find("COMPLEX-PHYSICAL-DEFAULT-VALUE"):
             self.complex_physical_default_value = create_complex_value_from_et(cpdv_elem)
 
--- a/odxtools/diaglayer.py
+++ b/odxtools/diaglayer.py
@@ -521,8 +521,7 @@ class DiagLayer:
 
     def get_can_func_req_id(self, protocol_name: Optional[str] = None) -> Optional[int]:
         """CAN Functional Request Id."""
-        com_param = self.get_communication_parameter(
-            "ISO_15765_2.CP_CanFuncReqId", protocol_name=protocol_name)
+        com_param = self.get_communication_parameter("CP_CanFuncReqId", protocol_name=protocol_name)
         if com_param is None:
             return None
 
--- a/odxtools/diaglayer.py
+++ b/odxtools/diaglayer.py
@@ -54,14 +54,9 @@ class DiagLayer:
         # DOP, units, etc
         self.local_diag_data_dictionary_spec = diag_layer_raw.diag_data_dictionary_spec
 
-        # Communication parameters, e.g. CAN-IDs
-        self._local_communication_parameters = diag_layer_raw.communication_parameters
-
         # Properties that include inherited objects
         self._services: NamedItemList[Union[DiagService,
                                             SingleEcuJob]] = NamedItemList(short_name_as_id)
-        self._communication_parameters: NamedItemList[CommunicationParameterRef] = NamedItemList(
-            short_name_as_id)
         self._data_object_properties: NamedItemList[DopBase] = NamedItemList(short_name_as_id)
 
     @staticmethod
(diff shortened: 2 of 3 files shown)


odxtools/comparam_subset.py
```diff
@@ -108,6 +108,7 @@ def __init_from_et__(self, et_element, doc_frags: List[OdxDocFragment]) -> None:
             if cp_el.tag in ("COMPARAM", "COMPLEX-COMPARAM"):
                 self.comparams.append(create_any_comparam_from_et(cp_el, doc_frags))
 
+        self.complex_physical_default_value = None
```
kayoub5, line 111: Why?

odxtools/diaglayer.py
```diff
@@ -416,24 +384,79 @@ def decode_response(self, response: Union[bytes, bytearray],
...
+        Note that, although communication parameters use inheritance,
+        it is *not* the "value inheritance" scheme used by e.g. DOPs,
+        tables, state charts, ...
+        """
+        return self._communication_parameters
+
     def get_communication_parameter(
         self,
-        name: str,
+        cp_id: str,
```
kayoub5, line 441: Is this ID or shortname?

odxtools/comparam_subset.py
andlaus: without this, it is possible that a comparam does not exhibit a `complex_physical_default_value` attribute, i.e., 
```python
print(comparam.complex_physical_default_value)
```
will raise an `AttributeError` instead of printing `None`...

odxtools/diaglayer.py
kayoub5: I have seems PDX files where the short name is CP_UniqueRespIdTable, the Diag Layer is called ISO_15765_2 BUT the XML REF and ID are NOT `ISO_15765_2.CP_UniqueRespIdTable`, rather they are randomly generated UUID from the modeling tool.

odxtools/comparam_subset.py
kayoub5: They need to put a fine on "driving and fixing bug", same way there is a fine on "driving and texting" 😛

andlaus:
@kayoub5: do you see any problems with this?

odxtools/diaglayer.py
andlaus: I suppose we are in a pickle, then: there are parameters which exhibit the same short names (`CP_UniqueRespIdTable`, `CP_TesterPresentTime`, ...) yet, depending on the COMPARAM-SUBSET they were defined in (`ISO_15765_2` (CAN) or `ISO_13400_2_DIS_2015` (DoIP)), they exhibit either a completely different definition (`CP_UniqueRespIdTable`) or at least different semantics (`CP_TesterPresentTime`). I don't really like the current solution of relying on the `protocol_name` to hopefully retrieve the correct one, but if there are no better ideas, I guess we have to rely on that again. Any ideas?

andlaus: BTW: I consider the fact that the comparams are not fully standardized to be -- err -- unfortunate...

andlaus: alright: I reverted the comparam retrieval mechanism back to using short names: f8119a7ae22a1f51c4004b6a98b3448a9807de1b

This has the unfortunate consequence that two COMPARAMs with the same short name cannot be specified for the same diag_layer object (e.g. `CP_UniqueRespIdTable` for both CAN and DoIP). That said, for inheritance the ODXLINK IDs are still used, because the ID generation scheme of the editor does not matter here and I believe that this is "more correct".

odxtools/diaglayer.py
```diff
@@ -492,7 +521,8 @@ def get_send_id(self) -> Optional[int]:
 
     def get_can_func_req_id(self, protocol_name: Optional[str] = None) -> Optional[int]:
         """CAN Functional Request Id."""
-        com_param = self.get_communication_parameter("CP_CanFuncReqId", protocol_name=protocol_name)
+        com_param = self.get_communication_parameter(
+            "ISO_15765_2.CP_CanFuncReqId", protocol_name=protocol_name)
```
kayoub5, line 525: sure about the name here?

andlaus: I'm sure. That it is wrong. thanks for the catch. Fixed: c353954686f7366f825961203c52e41f39cdc2f0...

odxtools/comparam_subset.py
andlaus: Right, I suppose that we should create a "drive-by fix jar" instead of a "swear jar". If there ever will be an odxtools-developer meeting, this basically will cover all your bases :D

kayoub5 (approved):

PR #144, opened 2023-06-16
Rebase latest main

chenyin0126:


andlaus (commented):
the changes themselfs are fine but incomplete: given a database, how is it possible to distinguish if the `<ADDRESSING>` tag of a service was not set at all or if it was set to `"PHYSICAL"`? The pattern used by odxtools in such cases is to store the unmodified value from `et_elem.findtext('ADDRESSING')`, name the corresponding class attribute `addressing_raw`, and add an `addressing` `@property` function which returns `PHYSICAL` if the tag was not specified.

besides that, write support would be great to have (cf [printService.xml.jinja2](https://github.com/mercedes-benz/odxtools/blob/main/odxtools/templates/macros/printService.xml.jinja2)), I'd love to see this tested by one of the unit tests,  and I'd like to see the whole commit series squashed into a single commit with a meaningful commit message.

If you want, I'll help you out with any of these issues...

chenyin0126:
@andlaus  Thanks for your idea. I can update it to use `et_element.findtext("ADDRESSING")`
I am not sure how "add an addressing @property function", since I define the addressing value inside " def from_et(et_element, doc_frags: List[OdxDocFragment]):" this function is a helper function without self as parameter. 
How can I add @property for this? since this function is for returning DiagService object. That is why I define addressing value with no @property function, since it would be easier to implement and code looks easier

andlaus:
> How can I add https://github.com/Property for this?

you add it to the `DiagService` class:

```python
class DiagService:
  addressing_raw: Optional[str]
  
  ...

  @property
  def addressing(self) -> str:
    return "PHYSICAL" if self.addressing_raw is None else self.addressing_raw
```

That said, I believe that the functionality of this PR is now fully contained in #231. Would be cool if you gave #231 a spin...

andlaus:
#231 has been merged, so let's close this. if #231 did not help, feel free to reopen this issue...

PR #145, opened 2023-06-19
Untangle diaglayer (Part 6)

andlaus:
This PR collects various smaller fixes for the diaglayer code (in that sense, it consists purely of "drive-by" fixes). The most notable of those changes is that the "logical" view now exposes all objects that are contained by the DIAG-DATA-DICTIONARY-SPEC tag using a separate `.data_dictionary_spec` attribute instead of providing e.g. the `.data_object_properties` directly. (This is not completely trivial because some of the sub-objects of the diag data dictionary spec are subject to value inheritance).

### DIFF
--- a/odxtools/diaglayer.py
+++ b/odxtools/diaglayer.py
@@ -618,7 +618,6 @@ class DiagLayer:
         """Timeout on inactivity in seconds.
 
         This is defined by the communication parameter "CP_TesterPresentTime".
-        If the variant does not define this parameter, the default value 3.0 is returned.
 
         Description of the comparam: "Time between a response and the
         next subsequent tester present message (if no other request is
--- a/odxtools/diaglayer.py
+++ b/odxtools/diaglayer.py
@@ -12,15 +12,19 @@ from .admindata import AdminData
 from .audience import AdditionalAudience, Audience
 from .communicationparameter import CommunicationParameterRef
 from .companydata import CompanyData, create_company_datas_from_et
-from .dataobjectproperty import DopBase
+from .dataobjectproperty import DopBase, DtcDop
 from .diagdatadictionaryspec import DiagDataDictionarySpec
 from .diaglayerraw import DiagLayerRaw
 from .diaglayertype import DiagLayerType
 from .ecu_variant_patterns import EcuVariantPattern, create_ecu_variant_patterns_from_et
+from .endofpdufield import EndOfPduField
+from .envdata import EnvironmentData
+from .envdatadesc import EnvironmentDataDescription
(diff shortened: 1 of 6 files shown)


odxtools/diaglayer.py
```diff
@@ -193,13 +244,9 @@ def services(self) -> NamedItemList[Union[DiagService, SingleEcuJob]]:
...
 
     @property
-    def data_object_properties(self) -> NamedItemList[DopBase]:
-        """All data object properties including inherited ones.
-        This attribute corresponds to all specializations of DOP-BASE
-        defined in the DIAG-DATA-DICTIONARY-SPEC of this diag layer as well as
-        in the DIAG-DATA-DICTIONARY-SPEC of any parent.
-        """
-        return self._data_object_properties
+    def diag_data_dictionary_spec(self) -> DiagDataDictionarySpec:
```
kayoub5, line 248: property name too long, could be potentially shortened without issues

odxtools/endofpdufield.py
```diff
@@ -156,10 +156,12 @@ def _resolve_odxlinks(self, odxlinks: OdxLinkDatabase) -> None:
...
         """Recursively resolve any short-name references"""
         if self.structure_snref is not None:
-            self._structure = diag_layer.data_object_properties[self.structure_snref]
+            self._structure = diag_layer.diag_data_dictionary_spec.data_object_props[
+                self.structure_snref]
 
         if self.env_data_desc_snref is not None:
-            self._env_data_desc = diag_layer.data_object_properties[self.env_data_desc_snref]
+            self._env_data_desc = diag_layer.diag_data_dictionary_spec.data_object_props[
+                self.env_data_desc_snref]
```
kayoub5, line 164: diag_data_dictionary_spec makes a simple property access breaks multiple lines

Maybe put a local variable `spec = diag_layer.diag_data_dictionary_spec` to simplify code?

kayoub5 (approved):

odxtools/diaglayer.py
andlaus: I agree. the problem is that the ODX spec calls this aberration DIAG-DATA-DICTIONARY-SPEC, and IMO it would be quite confusing to use something else. We can probably think about moving all objects which currently it contains to the diaglayer, but that would either make writing PDX files pretty hard and it would potentially confuse people who are familiar with the standard...

odxtools/endofpdufield.py
andlaus: good point. fixed: 76b4606189d7f645741d3348843fca72741a7aed

PR #146, opened 2023-06-20
DiagLayer: refactor the value inheritance mechanism

andlaus:
This is the final PR in the long quest to refactor `DiagLayer`: An overhaul of the code that implements value inheritance. Now, all objects that the spec says shall be subject to the value inheritance mechanism should be covered (modulo bugs), and the mechanism is IMO implemented in a more straight forward way (i.e., with less code duplication as well as all value-inheritance related code being kept in the `DiagLayer` class).

### DIFF
--- a/odxtools/diaglayer.py
+++ b/odxtools/diaglayer.py
@@ -196,7 +196,10 @@ class DiagLayer:
 
         #####
         # compute the communication parameters applicable to the
-        # diagnostic layer
+        # diagnostic layer. Note that communication parameters do
+        # *not* use value inheritance, but a slightly different
+        # scheme, cf the docstring of
+        # _compute_available_commmunication_parameters().
         #####
         self._communication_parameters = self._compute_available_commmunication_parameters()
 
@@ -533,7 +536,11 @@ class DiagLayer:
 
     @property
     def protocols(self) -> NamedItemList["DiagLayer"]:
-        """Return the set of all protocols which are applicable to the diagnostic layer"""
+        """Return the set of all protocols which are applicable to the diagnostic layer
+
+        Note that protocols are *not* explicitly inherited objects,
+        but the parent diagnostic layers of variant type "PROTOCOL".
+        """
         result_dict: Dict[str, DiagLayer] = dict()
 
         for parent_ref in self._get_parent_refs_sorted_by_priority():
--- a/odxtools/diaglayer.py
+++ b/odxtools/diaglayer.py
@@ -3,7 +3,7 @@
 import warnings
 from copy import copy
 from itertools import chain
-from typing import Any, Dict, Iterable, List, Optional, Tuple, Union, cast
+from typing import Any, Callable, Dict, Iterable, List, Optional, Tuple, TypeVar, Union
 from xml.etree import ElementTree
 
 from deprecation import deprecated
@@ -37,6 +37,8 @@ from .table import Table
 from .units import UnitGroup, UnitSpec
 from .utils import create_description_from_et, short_name_as_id
 
+T = TypeVar("T")
+
 
 class DiagLayer:
     """This class represents a "logical view" upon a diagnostic layer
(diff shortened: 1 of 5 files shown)


odxtools/diaglayer.py
```diff
@@ -170,6 +200,11 @@ def _finalize_init(self, odxlinks: OdxLinkDatabase) -> None:
         #####
         self._communication_parameters = self._compute_available_commmunication_parameters()
 
+        #####
+        # resolve all SNREFs. TODO: We allow SNREFS to inherited
```
kayoub5, line 204: typo: to be inherited

odxtools/diaglayer.py
```diff
@@ -256,60 +338,141 @@ def diag_data_dictionary_spec(self) -> DiagDataDictionarySpec:
...
+    def _get_parent_refs_sorted_by_priority(self, reverse=False) -> Iterable[ParentRef]:
         return sorted(
             self.diag_layer_raw.parent_refs,
             key=lambda pr: pr.layer.variant_type.inheritance_priority,
             reverse=reverse)
 
-    def __gather_local_services(
-            self, odxlinks: OdxLinkDatabase) -> List[Union[DiagService, SingleEcuJob]]:
-        diagcomms_by_name: Dict[str, Union[DiagService, SingleEcuJob]] = {}
+    def _compute_available_objects(
```
kayoub5, line 347: _compute_available_objects not used to simplify protocols and _compute_available_commmunication_parameters ?

odxtools/diaglayer.py
```diff
@@ -256,60 +338,141 @@ def diag_data_dictionary_spec(self) -> DiagDataDictionarySpec:
...
 
-        if self.local_diag_data_dictionary_spec:
-            for dop in self.local_diag_data_dictionary_spec.all_data_object_properties:
-                result_dict[dop.short_name] = dop
+        This is not completely trivial as it requires to resolving the
+        references specified in the <DIAG-COMMS> XML tag.
+        """
+        result_dict: Dict[str, Union[DiagService, SingleEcuJob]] = {}
+
+        # TODO (?): add objects from the import-refs
```
kayoub5, line 405: Forgot about this?

odxtools/diaglayer.py
```diff
@@ -256,60 +338,141 @@ def diag_data_dictionary_spec(self) -> DiagDataDictionarySpec:
...
+        get_local_objects_fn = lambda dl: dl.diag_layer_raw.global_negative_responses
+        not_inherited_fn = lambda parent_ref: parent_ref.not_inherited_global_neg_responses
+        return self._compute_available_objects(get_local_objects_fn, not_inherited_fn)
+
+    def _compute_available_data_object_props(self) -> Iterable[DataObjectProperty]:
+        get_local_objects_fn = lambda dl: \
+            [] if dl.diag_layer_raw.diag_data_dictionary_spec is None \
+            else dl.diag_layer_raw.diag_data_dictionary_spec.data_object_props
+        not_inherited_fn = lambda parent_ref: parent_ref.not_inherited_dops
+        return self._compute_available_objects(get_local_objects_fn, not_inherited_fn)
```
kayoub5, line 448: Why the lambda? why not just pass the parent_ref.not_inherited_dops list?

odxtools/diaglayer.py
```diff
@@ -238,12 +273,59 @@ def ecu_variant_patterns(self) -> List[EcuVariantPattern]:
...
     #######
+    @property
+    def diag_comms(self) -> NamedItemList[Union[DiagService, SingleEcuJob]]:
+        """All diagnostic communication primitives applicable to this DiagLayer
+
+        Diagnostic communication primitives are diagnostic services as
+        well as single-ECU jobs. This list has all references
+        resolved.
+        """
+        return self._diag_comms
```
kayoub5, line 284: Missing setting initial values of those private fields in __init__

kayoub5 (changes requested):

odxtools/diaglayer.py
andlaus: not really, this is more a reminder for future me (or other contributors): I have not come across files that which import-refs, and I am not 100% sure about their semantics...

odxtools/diaglayer.py
andlaus: because it is possible to have nested inheritance hierarchies where each diaglayer specifies multiple parentrefs...

odxtools/diaglayer.py
andlaus: they are created in `._finalize_init()`. I'm not really happy about the visibility of this solution, but given that this is a private implementation detail, `mypy` does not complain about it, and that this is a pretty common pattern in odxtools, I don't want to fix that in this PR. (as I see it, the main issue with the current solution is that accessing `diag_layer.diag_comms` will raise `AttributeError` if the `diag_layer` has not been properly initialized which is not very intuitive. That said, I guess that "normal" users will never encounter this?)

odxtools/diaglayer.py
andlaus: protocols are not inherited, but in some sense they *are* the inheritance hierarchy. Regarding the comparams, for some arcane reason, they use a slightly different inheritance mechanism: https://github.com/mercedes-benz/odxtools/pull/146/files#diff-3a939bdfc46d96678ce4207d04602c4123b800b483f627e8288cab31d3c2270bR352 (cf section 7.3.6). It's very cool that you notice such things, I probably wouldn't ;)

odxtools/diaglayer.py
andlaus: I wanted to say that it is allowed to specify an SNREF to an object that was inherited. I reformulated that a bit: 6e9ce255aed8f684098f86fb81063e65b030a573, thanks bringing this up.

odxtools/diaglayer.py
kayoub5: Could you put this as comments? so that the next guy does not try to "simplify" the code and break logic?

odxtools/structures.py
```diff
@@ -507,7 +507,8 @@ def __init__(self, *, response_type: str, **kwargs):  # "POS-RESPONSE" or "NEG-R
         self.response_type = response_type
 
     def encode(self, coded_request: Optional[ByteString] = None, **params) -> ByteString:
-        logger.info(f"Compose response message to the request {coded_request}")
+        if coded_request is not None:
+            logger.info(f"Compose response message to the request {coded_request.hex()}")
         if coded_request is not None:
```
kayoub5, line 512: Duplicate `if coded_request is not None:` check

andlaus: oops. right.

andlaus: fixed, thanks!

odxtools/diaglayer.py
kayoub5: Other than this, LGTM

andlaus: good point. done: 84003041aa08c2c223c532a7b2aac2f2c1b737de

kayoub5 (approved):

andlaus:
alright. I guess I should tag a release...

PR #147, opened 2023-06-22
Add support for decoding global negative responses

andlaus:
This is surprisingly difficult given the fact that global negative responses are only negative responses that can be returned by any service of a given ECU...
    
Note that, in order to be able to decode these, `DiagLayer.decode_response()` must be used, *not* `DiagService.decode_message()`. This is because the diagnostic service object does not know the diagnostic layers where it is contained within and -- even if it knew -- it is impossible to determine which diag layer is the "correct" one.

Further, for the same reason as decoding cannot be done via the diag service, encoding of global negative responses works directly via the respective `Response` object instead of `DiagService.encode_negative_response()`. This object can be obtained via `DiagLayer.global_neg_responses.<response short name>`...

Besides decoding support, this PR also adds a global negative response to the "Somersault" ECUs and adds the missing output writing and includes global negative responses in the output of the `odxtools list` command.


odxtools/diaglayer.py
```diff
@@ -843,67 +845,90 @@ def _build_coded_prefix_tree(self):
...
+        # dictionary (this is quite hacky...)
+        if sub_tree.get(-1) is None:
+            sub_tree[-1] = [service]
+        else:
+            cast(List[DiagService], sub_tree[-1]).append(service)
+
+    def _find_services_for_uds(self, message: Union[bytes, bytearray]) -> List[DiagService]:
         if not hasattr(self, "_prefix_tree"):
             # Compute the prefix tree the first time this decode function is called.
             self._prefix_tree = self._build_coded_prefix_tree()
```
kayoub5, line 892: maybe use the cached decorator? it usual simplify this kind of logic

odxtools/diaglayer.py
```diff
@@ -843,67 +845,90 @@ def _build_coded_prefix_tree(self):
...
+                for gnr in self.global_negative_responses:
+                    try:
+                        decoded_gnr = gnr.decode(message)
+                        decoded_messages.append(
+                            Message(
+                                coded_message=message,
+                                service=service,
+                                structure=gnr,
+                                param_dict=decoded_gnr))
+                    except DecodeError:
```
kayoub5, line 925: I have seen the decode logic throw all kind of exceptions including but not limited to assert exceptions, are you sure about the logic here?

odxtools/diaglayer.py
```diff
@@ -924,7 +949,20 @@ def decode_response(self, response: Union[bytes, bytearray],
...
+                for gnr in self.global_negative_responses:
+                    try:
+                        decoded_gnr = gnr.decode(response)
+                        decoded_messages.append(
+                            Message(
+                                coded_message=response,
+                                service=service,
+                                structure=gnr,
+                                param_dict=decoded_gnr))
+                    except DecodeError:
```
kayoub5, line 963: duplicate block?

odxtools/diaglayer.py
```diff
@@ -843,67 +845,90 @@ def _build_coded_prefix_tree(self):
...
+        sub_tree = prefix_tree
+        for b in coded_prefix:
+            if b not in sub_tree:
+                sub_tree[b] = {}
+            sub_tree = cast(PrefixTree, sub_tree[b])
+
+        # Store the object as in the prefix tree. This is done by
+        # assigning the list of possible objects to the key -1 of the
+        # dictionary (this is quite hacky...)
+        if sub_tree.get(-1) is None:
```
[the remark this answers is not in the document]
andlaus: I'd rather not to: even though this class is part of the standard library, I doubt that many people heard of it (I certainly didn't), making this code harder to understand. (Also, insertion is just needed once...)

odxtools/diaglayer.py
andlaus: thanks. implemented: 495cdf83ed657c3b0f3c0a166637351cae6b77de

odxtools/diaglayer.py
andlaus: pretty much: if a non-`DecodeError` exception is thrown, something serious went belly-up. usually that implies that the database or the input file is corrupted, or we stepped on a landmine (i.e. an implementation bug). If some data simply cannot be decoded, this should never happen (if it does, I'll consider it to be a bug).

odxtools/diaglayer.py
andlaus: kind of: the hunk above is in the `.decode()` method, while this here for `.decode_response()`. I agree that it's rather inelegant to copy-and paste such a non-trivial piece of code, but generalizing `.decode()` and `.decode_response()` would probably not make things easier to read. (and IMO the de-/encoding logic is quite complicated enough already...)

odxtools/diaglayer.py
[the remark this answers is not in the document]
kayoub5: defaultdict is part of the standard library

odxtools/diaglayer.py
kayoub5: found one https://github.com/mercedes-benz/odxtools/blob/8ddf730985d7f92f8399b20a550197ed0783ead3/odxtools/parameters/matchingrequestparameter.py#L36

odxtools/diaglayer.py
kayoub5: maybe move it to a private function used by both decode / decode response

andlaus: I'll have a look...

andlaus: okay, done: turns out this was way easier than I anticipated and a pretty good idea: 8859c3cbf0c09ea60546de026ee894206f31b7fd

odxtools/diaglayer.py
andlaus: fixed: 62e55eb5d9fea0714d0e9de9cb9cca45717b524d

kayoub5 (approved):

PR #149, opened 2023-06-28
Fix bugs found when using v4

kayoub5:


odxtools/companydata.py
```diff
@@ -100,13 +100,10 @@ class CompanySpecificInfo:
...
-        if related_docs is not None:
-            rdlist = list()
-            for rd in related_docs.iterfind("RELATED-DOC"):
-                rdlist.append(RelatedDoc.from_et(rd))
-
-            related_docs = rdlist
+        related_docs = [
+            RelatedDoc.from_et(rd)
+            for rd in et_element.iterfind("RELATED-DOCS/RELATED-DOC")
+        ]
```
kayoub5, line 106: Existing code caused related_docs to be None, causing errors in `_resolve_odxlinks`

odxtools/parameters/parameterwithdop.py
```diff
@@ -49,7 +49,8 @@ def _resolve_snrefs(self, diag_layer: "DiagLayer") -> None:
         super()._resolve_snrefs(diag_layer)
 
         if self.dop_snref:
-            self._dop = diag_layer.diag_data_dictionary_spec.data_object_props[self.dop_snref]
+            spec = diag_layer.diag_data_dictionary_spec
+            self._dop = spec.data_object_props.get(self.dop_snref) or spec.structures.get(self.dop_snref)
```
kayoub5, line 53: Both data_object_props and structures can be used here, Both inherit from DopBase

odxtools/companydata.py
andlaus: using an empty list instead of `None` if no element is defined is the default pattern in odxtools anyway..

odxtools/parameters/parameterwithdop.py
```diff
@@ -49,7 +49,11 @@ def _resolve_snrefs(self, diag_layer: "DiagLayer") -> None:
         super()._resolve_snrefs(diag_layer)
 
         if self.dop_snref:
-            self._dop = diag_layer.diag_data_dictionary_spec.data_object_props[self.dop_snref]
+            spec = diag_layer.diag_data_dictionary_spec
+            self._dop = (
+                spec.data_object_props.get(self.dop_snref)
+                or spec.structures.get(self.dop_snref)
+            )
```
andlaus, line 56: nice catch. (weird specification...)

andlaus (approved):
once the CI is happy (looks like you forgot to run the `reformat-source.sh` script), merge at will.

odxtools/parameters/parameterwithdop.py
andlaus: possibly. let's fix them on an "as needed" basis. I also consider it to be possible that something like this can happen with the short name resolution of different kinds of objects...

BTW: the short name resolution is not entirely correct at the moment: `NamedItemList` transforms short names into valid python identifiers, so short names that start with a digit get an underscore prefix and will thus currently not be properly resolved. (we would need something like
```python
obj = [ x if x.short_name == snref for x in objlist ][0]
```
but I don't consider this to be a big enough issue to take immediate action...

andlaus:
done

PR #161, opened 2023-07-17
slightly refactor the length key code

andlaus:
now, instead of using hacks with ODXLINK IDs, proper references to `LengthKeyParameter` are used by `ParamLengthInfoType`. This necessitates to include the diag coded types in the reference resolution machinery because `ParamLengthInfoType` is a diag coded type and it needs to resolve the reference to the length key parameter.

### DIFF
--- a/odxtools/diagcodedtypes.py
+++ b/odxtools/diagcodedtypes.py
@@ -2,7 +2,7 @@
 # Copyright (c) 2022 MBition GmbH
 import abc
 import math
-from typing import Any, List, Optional, Union
+from typing import TYPE_CHECKING, Any, List, Optional, Union
 
 import bitstruct
 
@@ -10,8 +10,12 @@ from .decodestate import DecodeState
 from .encodestate import EncodeState
 from .exceptions import DecodeError, EncodeError
 from .globals import logger, xsi
-from .odxlink import OdxDocFragment, OdxLinkId
+from .odxlink import OdxDocFragment, OdxLinkDatabase, OdxLinkId, OdxLinkRef
 from .odxtypes import DataType, odxstr_to_bool
+
+if TYPE_CHECKING:
+    from .diaglayer import DiagLayer
+    from .parameters.lengthkeyparameter import LengthKeyParameter
 
 ODX_TYPE_TO_FORMAT_LETTER = {
     DataType.A_INT32: "s",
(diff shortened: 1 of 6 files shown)


odxtools/diagcodedtypes.py
```diff
@@ -478,11 +493,28 @@ def __init__(
             base_type_encoding=base_type_encoding,
             is_highlow_byte_order_raw=is_highlow_byte_order_raw,
         )
-        self.length_key_id = length_key_id
+        self.length_key_ref = length_key_ref
```
kayoub5, line 496: small breaking change

odxtools/diagcodedtypes.py
```diff
@@ -519,14 +551,14 @@ def convert_bytes_to_internal(self, decode_state: DecodeState, bit_position: int
             # if isinstance(param_value.parameter, LengthKeyParameter) would be prettier,
             # but leads to cyclic import...
             if (parameter.parameter_type == "LENGTH-KEY" and
-                    parameter.odx_id == self.length_key_id  # type: ignore
+                    parameter.odx_id == self.length_key.odx_id  # type: ignore
```
kayoub5, line 554: type ignore still needed?

odxtools/diagcodedtypes.py
```diff
@@ -519,14 +551,14 @@ def convert_bytes_to_internal(self, decode_state: DecodeState, bit_position: int
...
                ):
                 # The bit length of the parameter to be extracted is given by the length key.
                 assert isinstance(value, int)
                 bit_length = value
                 break
 
-        assert bit_length is not None, f"Did not find any length key with ID {self.length_key_id}"
+        assert bit_length is not None, f"Did not find any length key with ID {self.length_key.odx_id}"
```
kayoub5, line 561: raise odx exception instead of assert

odxtools/diagcodedtypes.py
andlaus: in principle its a breaking change, but I'm will bet a box of Champaign that nobody outside of this repository ever used this attribute...

odxtools/diagcodedtypes.py
andlaus: unfortunately it is still required:
```
odxtools/diagcodedtypes.py:554: error: "Parameter" has no attribute "odx_id"  [attr-defined]
Found 1 error in 1 file (checked 110 source files)
```

this is due to the fact that the base class for parameters  does not exhibit an ID. (only some do, amongst them `LengthKeyParameter`.)

We should only ignore the `attr-defined` error here though. [Fixed](40dfba66b057779aeb45fa4dd0496f14d447ef6b), thanks!

odxtools/diagcodedtypes.py
andlaus: I'd prefer not to open this can of worm in this PR. (I guess I should do the `assert` -> `odxassert` soon.)

kayoub5 (approved):

PR #163, opened 2023-07-19
Smallish improvements

andlaus:
This PR features a bunch of smallish quality-of-life changes which might have minor functional and semantic consequences.

### DIFF
--- a/odxtools/diagcodedtypes.py
+++ b/odxtools/diagcodedtypes.py
@@ -547,8 +547,8 @@ class ParamLengthInfoType(DiagCodedType):
     def convert_bytes_to_internal(self, decode_state: DecodeState, bit_position: int = 0):
         # Find length key with matching ID.
         bit_length = 0
-        for parameter, value in decode_state.parameter_value_pairs:
-            if parameter.short_name == self.length_key.short_name:
+        for parameter_name, value in decode_state.parameter_values.items():
+            if parameter_name == self.length_key.short_name:
                 # The bit length of the parameter to be extracted is given by the length key.
                 assert isinstance(value, int)
                 bit_length = value
--- a/odxtools/diagcodedtypes.py
+++ b/odxtools/diagcodedtypes.py
@@ -514,7 +514,7 @@ class ParamLengthInfoType(DiagCodedType):
 
     def convert_internal_to_bytes(self, internal_value, encode_state: EncodeState,
                                   bit_position: int) -> bytes:
-        bit_length = encode_state.length_keys.get(self.length_key.odx_id, None)
+        bit_length = encode_state.length_keys.get(self.length_key.short_name, None)
 
         if bit_length is None:
             if self.base_data_type in [
@@ -534,7 +534,7 @@ class ParamLengthInfoType(DiagCodedType):
                 bit_length = math.ceil(bit_length / 8.0) * 8
 
         assert bit_length is not None
-        encode_state.length_keys[self.length_key.odx_id] = bit_length
+        encode_state.length_keys[self.length_key.short_name] = bit_length
 
         return self._to_bytes(
             internal_value,
@@ -548,17 +548,13 @@ class ParamLengthInfoType(DiagCodedType):
         # Find length key with matching ID.
         bit_length = 0
         for parameter, value in decode_state.parameter_value_pairs:
-            # if isinstance(param_value.parameter, LengthKeyParameter) would be prettier,
-            # but leads to cyclic import...
-            if (parameter.parameter_type == "LENGTH-KEY" and
-                    parameter.odx_id == self.length_key.odx_id  # type: ignore[attr-defined]
-               ):
+            if parameter.short_name == self.length_key.short_name:
                 # The bit length of the parameter to be extracted is given by the length key.
                 assert isinstance(value, int)
                 bit_length = value
                 break
 
-        assert bit_length is not None, f"Did not find any length key with ID {self.length_key.odx_id}"
+        assert bit_length is not None, f"Did not find any length key with short name {self.length_key.short_name}"
 
         # Extract the internal value and return.
         return self._extract_internal(
--- a/odxtools/encodestate.py
+++ b/odxtools/encodestate.py
@@ -26,7 +26,7 @@ class EncodeState(NamedTuple):
     """a mapping from short name to value for each parameter"""
     triggering_request: Optional[bytes] = None
     """If encoding a response: request that triggered the response"""
-    length_keys: Dict[OdxLinkId, int] = {}
-    """Mapping from IDs to bit lengths (specified by LengthKeyParameters)"""
+    length_keys: Dict[str, int] = {}
+    """Mapping from short names to bit lengths (specified by LengthKeyParameters)"""
     is_end_of_pdu: bool = False
     """Flag whether the parameter is the last on the PDU (needed for MinMaxLengthType)"""
--- a/odxtools/encodestate.py
+++ b/odxtools/encodestate.py
@@ -1,12 +1,14 @@
 # SPDX-License-Identifier: MIT
 # Copyright (c) 2022 MBition GmbH
+from dataclasses import dataclass, field
 from typing import Any, Dict, NamedTuple, Optional, Union
 
 from .odxlink import OdxLinkId
 from .odxtypes import AtomicOdxType
 
 
-class EncodeState(NamedTuple):
+@dataclass
+class EncodeState:
     """Utility class to be used while encoding a message.
 
     While encoding parameters may update the dicts with new keys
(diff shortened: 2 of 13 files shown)


odxtools/diagcodedtypes.py
```diff
@@ -514,7 +514,7 @@ def length_key(self) -> "LengthKeyParameter":
 
     def convert_internal_to_bytes(self, internal_value, encode_state: EncodeState,
                                   bit_position: int) -> bytes:
-        bit_length = encode_state.length_keys.get(self.length_key.odx_id, None)
+        bit_length = encode_state.length_keys.get(self.length_key.short_name, None)
```
kayoub5, line 517: sure about using short name instead of oid?

odxtools/diagcodedtypes.py
```diff
@@ -547,18 +547,14 @@ def convert_internal_to_bytes(self, internal_value, encode_state: EncodeState,
...
         # Find length key with matching ID.
         bit_length = 0
-        for parameter, value in decode_state.parameter_value_pairs:
-            # if isinstance(param_value.parameter, LengthKeyParameter) would be prettier,
-            # but leads to cyclic import...
-            if (parameter.parameter_type == "LENGTH-KEY" and
-                    parameter.odx_id == self.length_key.odx_id  # type: ignore[attr-defined]
-               ):
+        for parameter_name, value in decode_state.parameter_values.items():
+            if parameter_name == self.length_key.short_name:
```
kayoub5, line 551: does this logic assume the length and the value are in the same struct? or it can handle length and value belonging to different parameters in the hierarchy?

odxtools/nameditemlist.py
```diff
@@ -49,6 +50,11 @@ def append(self, item: T):
                           f"Encountered name '{item_name}' which is not an "
                           f"identifier!")
 
+        # make sure that the name of the item in question is not a
+        # python keyword (this would lead to syntax errors)
+        if iskeyword(item_name):
+            item_name = f"{item_name}_"
```
kayoub5, line 56: does this affect trying to access value with `[“name“] `?

odxtools/parameters/lengthkeyparameter.py
```diff
@@ -44,13 +44,15 @@ def is_optional(self):
...
-        physical_value = encode_state.parameter_values.get(self.short_name,)
+        physical_value = encode_state.parameter_values.get(self.short_name)
 
         if physical_value is None:
             raise TypeError(f"A value for the length key '{self.short_name}'"
                             f" must be specified.")
 
         # Set the value of the length key in the length key dict.
-        encode_state.length_keys[self.odx_id] = physical_value
+        encode_state.length_keys[self.short_name] = physical_value
```
kayoub5, line 54: this will cause collision if two lengths in different structs use a common generic name like "len", "length", "size"...

odxtools/parameters/parameterbase.py
```diff
@@ -108,15 +108,16 @@ def decode_from_pdu(self, decode_state: DecodeState):
         pass
 
     def encode_into_pdu(self, encode_state: EncodeState) -> bytes:
-        """Insert the encoded value of a parameter into the coded RPC.
+        """Insert the encoded value of a parameter into a binary blob of a message.
```
kayoub5, line 111: you are not inserting here, right? you are returning

odxtools/parameters/parameterbase.py
```diff
@@ -131,33 +132,37 @@ def encode_into_pdu(self, encode_state: EncodeState) -> bytes:
...
-        if len(old_rpc) < min_length:
-            # Make byte code longer if necessary
-            new_rpc += bytearray([0] * (min_length - len(old_rpc)))
-        for byte_idx_val, byte_idx_rpc in enumerate(
-                range(byte_position, byte_position + len(byte_value))):
+            byte_position = len(msg_blob)
+
+        return self._insert_into_blob(msg_blob, param_blob, byte_position)
+
+    def _insert_into_blob(self, blob: bytes, new_data: bytes, pos: Optional[int] = None) -> bytes:
```
kayoub5, line 145: this is not really an "insert", input is not modified

odxtools/structures.py
```diff
@@ -224,30 +224,27 @@ def convert_physical_to_bytes(self,
             is_end_of_pdu=encode_state.is_end_of_pdu,
         )
 
-    def convert_bytes_to_physical(self, decode_state: DecodeState, bit_position: int = 0):
+    def convert_bytes_to_physical(self,
+                                  decode_state: DecodeState,
+                                  bit_position: int = 0) -> Tuple[ParameterValueDict, int]:
```
kayoub5, line 229: maybe make an alias for `Tuple[ParameterValueDict, int] `?

tests/test_diag_coded_types.py
```diff
@@ -124,7 +123,7 @@ def test_encode_leading_length_info_type_unicode2string(self):
             base_type_encoding=None,
             is_highlow_byte_order_raw=None,
         )
-        state = EncodeState(coded_message=bytes([0x12]), parameter_values={})
+        state = EncodeState(coded_message=bytes([0x12]), parameter_values={}, length_keys={})
```
kayoub5, line 126: why reset length_keys?

odxtools/diagcodedtypes.py
andlaus: I'm sure that the name of a parameter is unique within the context of a given (sub-)structure, and I prefer to use the short names as indices because they are plain strings and thus quite a bit more intuitive to work with IMO...

odxtools/diagcodedtypes.py
andlaus: as far as I can see, it is the latter as long as the length key is physically in front of the value. note that using IDs does not change these aspects of the coding logic.

odxtools/nameditemlist.py
andlaus: yes, but that already happens for names that are not valid python identifiers (like `4ever`, `2do`, etc.) The main purpose of this class is to make interactive exploration of the dataset as convenient as possible, so this is IMO okay. 

The bullet-proof way of accessing parameters is to go through the list directly, e.g.
```python
nil = NamedItemList(lambda x: x[0], [("foo", 1), ("bar", 2)])
foo = [ x for x in nil if x[0] == "foo"][0]
print(foo) # ("foo", 1)
```

odxtools/parameters/lengthkeyparameter.py
andlaus: yes, but I assume that short names must be unique within each structure (as far as I remember, there are SNREFs to parameters somewhere...)

odxtools/parameters/parameterbase.py
andlaus: yep. I have trouble coming up with a better name, though. Do you have any ideas?

odxtools/structures.py
andlaus: not sure: this is pretty deep in the decoding logic and it is easier to see what's returned if this is directly specified. 

That said, I think the decoding and encoding code would benefit from a major overhaul. In this case, it would possibly be better if the position of the next byte would be put into `decode_state`. (but maybe not, because this would make the arguments non-const...)

tests/test_diag_coded_types.py
andlaus: this is not a reset, the length keys are just explicitly specified. ([removed](7c1486c12d7d0b4a1eb840588e22cd1d5165946b)]

On that note, the current code is wrong: `length_key` is defaulted by a mutable object, so we get this carnage:
```python
from odxtools.encodestate import EncodeState

es = EncodeState(b'', parameter_values={})
es.length_keys['abc'] = 123

es2 = EncodeState(b'', parameter_values={})

print(es2.length_keys)
# -> {'abc': 123}
```

to fix this, we need to make this a dataclass and initialize these members using `field()`...

odxtools/parameters/parameterbase.py
kayoub5:
```suggestion
        """Encode the value of a parameter into a binary blob and return it
```

odxtools/parameters/parameterbase.py
andlaus: okay, I changed it to `encode`...

andlaus:
okay, I fixed the "mutable object used as default value" issue for EncodeState and DecodeState: https://github.com/mercedes-benz/odxtools/pull/163/commits/1e312f0b32da811da3bd0066869efff492aa13be . Unfortunately, this involved a bit of copying of DecodeState objects. (While quite inelegant, performance should actually slightly improve because the `_replace` methods of NamedTuple does basically the same and we now only need to copy the decode state once per function call instead of once per parameter...)

kayoub5 (approved):

odxtools/encodestate.py
```diff
@@ -20,13 +22,18 @@ class EncodeState(NamedTuple):
```
kayoub5, line 22: You forgot to update docs here

kayoub5:
@andlaus 

You forgot to update end of pdu field

https://github.com/mercedes-benz/odxtools/blob/c5954adf9d9354bdb844bb4c595f4e39a70b53ff/odxtools/endofpdufield.py#L127

andlaus:
thanks for the catch. I hope to have fixed this in #164 (https://github.com/mercedes-benz/odxtools/pull/164/commits/bdc56ee9dc763de6e6735f138c97dfa188d1dafe ). can you check if it works?

PR #164, opened 2023-07-21
Refactor table classes

andlaus:
this PR refactors the table classes once more:

- Table rows now get a reference to the table which contains them. This is required to resolve SNREFs in table rows.
- In Tables, the table rows are now specified as a mixed list of ODXLINK references and table row objects. The approach is identical to the one for diagnostic communication objects for diagnostic layer. The reason for this is that the XML specification defines rows this way (as with the diagnostic services, `table.table_rows` provides a fully resolved view, while `table.table_rows_raw` is the table row specification as it is found in the XML)
-  The `TableRow` class is put to a separate file and `TableBase` is subsumed into its only user, `Table`.
- The `DiagLayer` class does not hold the table objects directly anymore. instead, they are moved to the `DiagDataDictionarySpec` like in the XML.
- various smaller cleanups and refactorings

### DIFF
--- a/odxtools/diaglayer.py
+++ b/odxtools/diaglayer.py
@@ -109,9 +109,6 @@ class DiagLayer:
         global_negative_responses = self._compute_available_global_neg_responses(odxlinks)
         self._global_negative_responses = NamedItemList(short_name_as_id, global_negative_responses)
 
-        tables = self._compute_available_tables()
-        self._tables = NamedItemList(short_name_as_id, tables)
-
         functional_classes = self._compute_available_functional_classes()
         self._functional_classes = NamedItemList(short_name_as_id, functional_classes)
 
@@ -157,6 +154,7 @@ class DiagLayer:
         ############
 
         dops = NamedItemList(short_name_as_id, self._compute_available_data_object_props())
+        tables = NamedItemList(short_name_as_id, self._compute_available_tables())
         dtc_dops: NamedItemList[DtcDop]
         structures: NamedItemList[BasicStructure]
         end_of_pdu_fields: NamedItemList[EndOfPduField]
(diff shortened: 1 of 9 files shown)


odxtools/diaglayer.py
```diff
@@ -313,11 +311,6 @@ def global_negative_responses(self) -> NamedItemList[Response]:
         """All global negative responses applicable to this DiagLayer"""
         return self._global_negative_responses
 
-    @property
-    def tables(self) -> NamedItemList[Table]:
```
kayoub5, line 317: The `diaglayer.tables` property is gone, how is the user supposed to access tables now?

kayoub5:
I tried to test this PR with the PDX files I have, found out that https://github.com/mercedes-benz/odxtools/pull/163 caused breaking changes

odxtools/diaglayer.py
andlaus: via `diag_layer.diag_data_dictionary_spec.tables`. You have a point though, I should add a deprecated method which does that. (fixed: https://github.com/mercedes-benz/odxtools/pull/164/commits/bdc56ee9dc763de6e6735f138c97dfa188d1dafe )

andlaus:
> I tried to test this PR with the PDX files I have, found out that https://github.com/mercedes-benz/odxtools/pull/163 caused breaking changes

right, [fixed](https://github.com/mercedes-benz/odxtools/pull/164/commits/bdc56ee9dc763de6e6735f138c97dfa188d1dafe) (hopefully). note that this is a gap in the unit test coverage. Unfortunately, it's not completely trivial to add a test for End-of-PDU encoding...

andlaus:
@kayoub5: Do you any showstoppers with this or can it be merged?

kayoub5:
> @kayoub5: Do you any showstoppers with this or can it be merged?

No show stoppers, just make a release once merged

andlaus:
>  No show stoppers, just make a release once merged

cool, thanks. Since the latest release did not come with the EndOfPduField regression, I don't think that this is critical and I would like to wait for my next PR (for en- and decoding TableStruct parameters) is merged...

PR #165, opened 2023-07-24
Table parameter coding

andlaus:
This PR adds support for en- and decoding TABLE-KEY and TABLE-STRUCT parameters. (Tables are basically a quite generic multiplexing mechanism.)


odxtools/parameters/tablekeyparameter.py
```diff
@@ -37,37 +40,107 @@ def _resolve_odxlinks(self, odxlinks: OdxLinkDatabase) -> None:
         super()._resolve_odxlinks(odxlinks)
 
         if self.table_ref:
-            self._table = odxlinks.resolve(self.table_ref)
+            if TYPE_CHECKING:
+                self._table = odxlinks.resolve(self.table_ref, Table)
+            else:
+                self._table = odxlinks.resolve(self.table_ref)
```
kayoub5, line 46: Why?

odxtools/parameters/tablekeyparameter.py
```diff
@@ -37,37 +40,107 @@ def _resolve_odxlinks(self, odxlinks: OdxLinkDatabase) -> None:
...
+
+        self._table_row: Optional[TableRow] = None
         if self.table_row_ref:
-            self._table_row = odxlinks.resolve(self.table_row_ref)
+            if TYPE_CHECKING:
+                self._table_row = odxlinks.resolve(self.table_row_ref, TableRow)
+            else:
+                self._table_row = odxlinks.resolve(self.table_row_ref)
```
kayoub5, line 53: Why different code based on TYPE_CHECKING ?

odxtools/parameters/tablekeyparameter.py
```diff
@@ -37,37 +40,107 @@ def _resolve_odxlinks(self, odxlinks: OdxLinkDatabase) -> None:
...
+                "reference is defined, a table must also be specified."
+            self._table_row = self._table.table_rows[self.table_row_snref]
 
     @property
     def table(self) -> "Table":
         return self._table
 
     @property
-    def table_row(self) -> "TableRow":
+    def table_row(self) -> Optional["TableRow"]:
```
kayoub5, line 73: Why optional?

odxtools/parameters/tablekeyparameter.py
```diff
@@ -37,37 +40,107 @@ def _resolve_odxlinks(self, odxlinks: OdxLinkDatabase) -> None:
...
-
-    def get_coded_value(self):
-        raise NotImplementedError("Encoding a TableKeyParameter is not implemented yet.")
-
-    def get_coded_value_as_bytes(self):
-        raise NotImplementedError("Encoding a TableKeyParameter is not implemented yet.")
-
-    def decode_from_pdu(self, coded_message, default_byte_position=None):
-        raise NotImplementedError("Decoding a TableKeyParameter is not implemented yet.")
+        return not is_required
```
kayoub5, line 80:
```suggestion
        return not self.is_required
```

kayoub5: Isn't is_optional always the oppsite of is_required? or is there a case where they could be both True/False?

odxtools/parameters/tablekeyparameter.py
```diff
@@ -37,37 +40,107 @@ def _resolve_odxlinks(self, odxlinks: OdxLinkDatabase) -> None:
...
+        key_dop = self.table.key_dop
+        bit_position = 0 if self.bit_position is None else self.bit_position
+        if key_dop is None:
+            raise EncodeError(f"Table '{self.table.short_name}' does not define "
+                              f"a KEY-DOP, but is used in TABLE-KEY parameter "
+                              f"'{self.short_name}'")
+        return key_dop.convert_physical_to_bytes(tr.key, encode_state, bit_position=bit_position)
+
+    def encode_into_pdu(self, encode_state: EncodeState) -> bytes:
+        # Set the value of the length key in the length key dict.
```
kayoub5, line 97: length key?

odxtools/parameters/tablestructparameter.py
```diff
@@ -14,17 +25,94 @@ def __init__(self, *, table_key_ref, table_key_snref, **kwargs):
...
+        super()._resolve_odxlinks(odxlinks)
+
+        if self.table_key_ref is not None:
+            self._table_key = odxlinks.resolve(self.table_key_ref)
+
+    def _resolve_snrefs(self, diag_layer: "DiagLayer") -> None:
+        super()._resolve_snrefs(diag_layer)
+
+        if self.table_key_snref is not None:
+            raise NotImplementedError("Table keys cannot yet defined using SNREFs"
```
kayoub5, line 41: Warn, don't block file loading

odxtools/parameters/tablekeyparameter.py
```diff
@@ -37,37 +40,107 @@ def _resolve_odxlinks(self, odxlinks: OdxLinkDatabase) -> None:
...
         return self._table
 
     @property
-    def table_row(self) -> "TableRow":
+    def table_row(self) -> Optional["TableRow"]:
         return self._table_row
 
     def is_required(self):
-        raise NotImplementedError("TableKeyParameter.is_required is not implemented yet.")
+        return self._table_row is None
```
kayoub5, line 77: It should be possible to figure out the value for this from the TableStructParameter, same way we figure out the length parameter, no?

odxtools/parameters/tablestructparameter.py
```diff
@@ -14,17 +25,94 @@ def __init__(self, *, table_key_ref, table_key_snref, **kwargs):
...
     def is_required(self):
-        raise NotImplementedError("TableStructParameter.is_required is not implemented yet.")
+        return True
 
     def is_optional(self):
-        raise NotImplementedError("TableStructParameter.is_optional is not implemented yet.")
+        return False
+
+    def get_coded_value(self, physical_value=None):
+        raise EncodeError("TableStructParameters cannot be converted to "
```
kayoub5, line 55: Maybe use the same we did for mutex, with the wrapper dict that have one key?

odxtools/parameters/tablestructparameter.py
```diff
@@ -14,17 +25,94 @@ def __init__(self, *, table_key_ref, table_key_snref, **kwargs):
...
+            inner_params = encode_state.parameter_values[self.short_name]
+            if not isinstance(inner_params, dict):
+                raise EncodeError(f"The sub-parameters for '{self.short_name}' must be "
+                                  f"specified using a key->value dictionary")
+            return structure.convert_physical_to_bytes(
+                inner_params,  # type: ignore [arg-type]
+                encode_state)
+        else:
+            # the table row associated with the key neither defines a
+            # DOP not a structure -> ignore it
```
kayoub5, line 89:
```suggestion
            # DOP nor a structure -> ignore it
```

odxtools/structures.py
```diff
@@ -121,16 +122,15 @@ def print_free_parameters_info(self) -> None:
...
-
+                                     param_values: ParameterValueDict,
+                                     triggering_coded_request: Optional[bytes],
+                                     is_end_of_pdu: bool = True) -> bytes:
         coded_rpc = bytearray()
         encode_state = EncodeState(
             coded_rpc,
             dict(param_values),
+            length_keys={},
+            table_keys={},
```
kayoub5, line 133: Why  length_keys here?

odxtools/parameters/tablekeyparameter.py
andlaus: to prevent mypy complaining about a the type of `._table`. (we can't do this unconditionally because that would lead to a circular import.)

odxtools/parameters/tablekeyparameter.py
andlaus: to prevent mypy from complaining without getting circular imports.

odxtools/parameters/tablekeyparameter.py
andlaus: because table keys with a defined table row is the static case (i.e., in this case, the interpretation of the associated TABLE-STRUCT-PARAMs does not depend on the received message.)

odxtools/parameters/tablekeyparameter.py
andlaus: not sure what you mean: IIRC, `is_required()` says whether the parameter must be specified for `encode()` or if it can be defaulted or not. this is the the case iff we are not in the static case, i.e., the table row is not defined.

odxtools/parameters/tablekeyparameter.py
kayoub5: You can break the circular import by making the import lazy, and moving it inside the function
Doing import inside function is usually discoureged, but it is one of the simplest and most effective ways to break circular dependency without any side effects
```suggestion
            from .table import Table
            self._table = odxlinks.resolve(self.table_ref, Table)
```

odxtools/parameters/tablekeyparameter.py
kayoub5: It is still possible to figure out the tablekeyparameter value from the tablestructparameter, same way we figure out the lengthkeyparameter, no?

odxtools/parameters/tablekeyparameter.py
andlaus: IIRC, `is_optional` says whether the the parameter must be explicitly specified for encoding purposes or not, while `.is_required` specifies if the parameter has some effect on the result. (i.e., value parameters for which a default value is specified are optional but required.) If this is the case (don't remember for sure), the names of these methods should change IMO...

odxtools/parameters/tablekeyparameter.py
```diff
@@ -37,37 +40,107 @@ def _resolve_odxlinks(self, odxlinks: OdxLinkDatabase) -> None:
...
+    def get_coded_value(self, physical_value=None):
+        key_dop = self.table.key_dop
+        return key_dop.convert_physical_to_internal(physical_value)
+
+    def get_coded_value_as_bytes(self, encode_state: EncodeState) -> bytes:
+        tr = encode_state.table_keys[self.short_name]
+        key_dop = self.table.key_dop
+        bit_position = 0 if self.bit_position is None else self.bit_position
+        if key_dop is None:
+            raise EncodeError(f"Table '{self.table.short_name}' does not define "
```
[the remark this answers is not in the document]
andlaus: that's because mypy did not typecheck `get_coded_value` because it did not specify a return type. fixed...

odxtools/parameters/tablekeyparameter.py
andlaus: copy and paste: Best friend, worst enemy...

fixed, thanks

odxtools/parameters/tablekeyparameter.py
```diff
@@ -37,37 +40,107 @@ def _resolve_odxlinks(self, odxlinks: OdxLinkDatabase) -> None:
...
+            assert key_dop is not None
+            bit_position_int = self.bit_position if self.bit_position is not None else 0
+            phys_val, next_byte_position = key_dop.convert_bytes_to_physical(
+                decode_state, bit_position=bit_position_int)
+
+            table_row_candidates = [x for x in self.table.table_rows if x.key == phys_val]
+            if len(table_row_candidates) == 0:
+                raise DecodeError(f"No table row exhibiting the key '{str(phys_val)}' found")
+            elif len(table_row_candidates) > 1:
+                raise DecodeError(f"Multiple rows exhibiting key '{str(phys_val)}' found in table")
```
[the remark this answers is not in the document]
andlaus: the paragraph about TABLE-KEY in section 7.3.6.11 says "[...]The physical result shall
match with exactly one KEY-value in the TABLE."

odxtools/parameters/tablestructparameter.py
andlaus: I don't really get what you mean here: The issue with table structs is that `physical_value` is depends in which row is selected, but the selected row is not specified in the table struct but the corresponding TABLE-KEY parameter...

odxtools/structures.py
andlaus: historical artifact: in a previous version of the patch I removed the mutable default values for `EncodeState` instead of making it a dataclass.

odxtools/parameters/tablekeyparameter.py
andlaus: not really: for this, we would need to know what struct was selected and to determine this we would need the assumption that the key->struct mapping is one-to-one, i.e. that given a struct, we can find a unique key (this is not the case since tables are free to map multiple keys to the same struct).

The reason why we can figure out the value of length keys implicitly while encoding is that they are just the length of the list of the associated parameter...

odxtools/parameters/tablekeyparameter.py
andlaus: yeah, that's used in a few places already. I always wonder what the performance implications of this are, though...

andlaus: the performance implications can be significant:
```python
import timeit
import odxtools

def eager():
    return type(odxtools.load_file)

def lazy():
    import odxtools as odxtools_lazy
    return type(odxtools_lazy.load_file)

t = timeit.timeit("eager()", number=1000*1000, globals=globals())
print(f"eager: {t} seconds elapsed")

t = timeit.timeit("lazy()", number=1000*1000, globals=globals())
print(f"lazy: {t} seconds elapsed")
```

produces
```
eager: 0.038458367984276265 seconds elapsed
lazy: 0.12220152601366863 seconds elapsed
```
on my machine. I'd say performance are acceptable if there is no other solution, but lazy imports should be avoided if possible...

odxtools/parameters/tablekeyparameter.py
kayoub5: Each row have a short_name, the idea is that instead of the tablestructparameter containing directly the struct, it would be wrapped in a dict with one key-value, key being the row short_name, and value being the struct that needs to be encoded

This would be simular to how we encode mutex at the moment.

odxtools/parameters/tablestructparameter.py
```diff
@@ -38,8 +39,9 @@ def _resolve_snrefs(self, diag_layer: "DiagLayer") -> None:
         super()._resolve_snrefs(diag_layer)
 
         if self.table_key_snref is not None:
-            raise NotImplementedError("Table keys cannot yet defined using SNREFs"
-                                      " for TableStructParameters.")
+            warnings.warn(
+                "Table keys cannot yet defined using SNREFs"
```
kayoub5, line 43: grammar police

"cannot yet be defined" or "cannot yet define"

odxtools/parameters/tablekeyparameter.py
andlaus: do you mean to encode PDUs with table-struct parameters like this:
```python
rq.encode(table_struct_param={"nice_table_row_key":{"table_struct_param1":"table_struct_value1", }})
```

instead of
```python
rq.encode(table_key_param="nice_table_row_key", table_struct_param={"table_struct_param1":"table_struct_value1", })
```
?

the issue with this is that ODX separate the encoding of the table-keys from the table structs, and there is no "back link" from the table keys to the table structs. This means that (in principle), multiple table struct parameters can use the same table key. (I doubt that many people use this, but ...). Of course, these issues can be solved, but I wonder if they are worth it. Possibly, it is better to remove the implicit encoding of length-keys instead...

kayoub5: > do you mean to encode PDUs with table-struct parameters like this

Yes

Let me explain the reasons for my request a bit
* Having to specify two parameters independently will lead to weird error messages when user change one and forget to change the other
* I am (in my internal projects) trying to generate JSON schema from the PDX parameters (and use it to generate a GUI), so having a consitent behavior for each parameter (and not chaning what it takes as value) would be nicer

> there is no "back link" from the table keys to the table structs. This means that (in principle), multiple table struct parameters can use the same table key.

That's another point, people who have those "weird" uses cases, can do this
```
rq.encode(table_key_param=table_row_key_physical_value, table_struct_param={"table_row_short_name":{"table_struct_param1":"table_struct_value1", }})
```
table_key_param will have a default value from table_struct_param (the same way we encode length paramters)
but user can still specify it explicitly

andlaus:
okay. I've implemented the implicit scheme which you proposed. Since it is impossible to distinguish implicit from explicit coding if the referenced structure has only one parameter, one now always has to specify the short name of the table row for TABLE-STRUCT parameters. example:

```python
ecu.positive_responses[0].encode(my_table_struct_param={"table_row_5":{"foo_param":1, "bar_param": 2, "foobar_param": 3}})
```

specifying the corresponding table-key parameter is optional, but if it is specified, its value must be the same as the selected row for the corresponding TABLE-STRUCT-PARAMs. Further, if table struct parameters use a table key stemming from a lower nesting level (i.e., a struct with a table-key parameter that contains structs with table-struct parameters referencing the outer table key), the table key parameter must be explicitly specified. The same applies to length keys. (AFAICS, it always has been this way. IMO these are quite rare corner cases, though.)

odxtools/parameters/lengthkeyparameter.py
```diff
@@ -45,23 +45,14 @@ def is_optional(self):
...
-        physical_value = encode_state.parameter_values.get(self.short_name)
-
-        if physical_value is None:
-            raise TypeError(f"A value for the length key '{self.short_name}'"
-                            f" must be specified.")
-
-        # Set the value of the length key in the length key dict.
-        encode_state.length_keys[self.short_name] = physical_value
-
         return super().encode_into_pdu(encode_state)
```
kayoub5, line 56: This makes the overload of this function useless, it could be just removed

odxtools/templates/macros/printResponse.xml.jinja2
```diff
@@ -10,7 +10,9 @@
 {%- macro printResponse(resp, tag_name="POS-RESPONSE") -%}
 <{{tag_name}} ID="{{resp.odx_id.local_id}}">
  <SHORT-NAME>{{resp.short_name}}</SHORT-NAME>
+{%- if resp.long_name %}
  <LONG-NAME>{{resp.long_name|e}}</LONG-NAME>
+{%- endif %}
```
kayoub5, line 15: Drive by fix?

odxtools/templates/macros/printDOP.xml.jinja2
```diff
@@ -152,7 +152,9 @@
 {%- macro printDOP(dop, tag_name) %}
 <{{tag_name}} ID="{{dop.odx_id.local_id}}">
  <SHORT-NAME>{{dop.short_name}}</SHORT-NAME>
+ {%- if dop.long_name %}
  <LONG-NAME>{{dop.long_name|e}}</LONG-NAME>
+ {%- endif %}
```
kayoub5, line 157: Drive by fix?

odxtools/parameters/lengthkeyparameter.py
andlaus: yes. I thought it would improve the visibility (`get_coded_value_as_bytes()` is not supposed to be called externally if I understand this coed correctly). The en-/decoding code should be fundamentally refactored soon because it is currently quite a bowl of spaghetti IMO...

odxtools/templates/macros/printResponse.xml.jinja2
andlaus: yes. That one was caused because for some reason, with the other changes included in this PR, my test script started to fail because a internal file which was passed through [`examples/pdxcopy.py`](https://github.com/mercedes-benz/odxtools/blob/main/examples/pdxcopy.py) started produce different results than the original for the `odxtools list` subcommand. there are probably a few more of these (sorry for that)...

kayoub5:
Drive by bug found, should be `..diaglayer` instead `.diaglayer`

https://github.com/mercedes-benz/odxtools/pull/165/files#diff-42a83494ffe5dde99cf6e7ca32e2103f5aa5e58e634766d9f0721fa87c024e4bR15

andlaus:
okay, I've cleaned up the way how to specify TABLE-STRUCT parameters a bit: the PR now uses tuples instead of dictionaries and the decoding routine should produce the same format as the encoding one: https://github.com/mercedes-benz/odxtools/pull/165/commits/039cb150f9ec2d7d38d5144e78ac4f50c4405468 .

@kayoub5: do you see any further issues with this or can it be merged?

andlaus:
> Drive by bug found, should be `..diaglayer` instead `.diaglayer`
> 
> https://github.com/mercedes-benz/odxtools/blob/main/odxtools/parameters/tablekeyparameter.py#L12

right. I think current mypy versions do not seem care, though. fixed: https://github.com/mercedes-benz/odxtools/pull/165/commits/9ab9e3213d4f815ec362bfb111f6839063597cb5

kayoub5:
@andlaus could you publish a version with this change?

PR #166, opened 2023-07-27
Fix crash caused by sdg not being hashashable

kayoub5:


andlaus:
looks good, merging...

andlaus:
BTW: what does "hashashable" mean? something involving THC? ;-P

kayoub5:
> BTW: what does "hashashable" mean? something involving THC? ;-P

hashable means primitive type (int, float, str) or class that impliments `__hash__` function, objects that are hashable are supposed to be immutable (can be used as dict keys)

andlaus:
that's hashable, sure. but I was wondering about "hashashable" ;)

PR #168, opened 2023-07-31
refactor exception handling

andlaus:
These changes make it possible to continue the program execution if a deviation from the ODX specification was detected or a hard-coded assumption was violated. (If any of this happens, the behavior after that point is undefined.) The new mechanism consists of the following parts:

- The `odxtools.exceptions.strict_mode` variable indicates whether program execution should be attempted to be continued if a violation is detected or not. The default is to not continue. Switching to non-strict mode can be achieved using
  ```python
  import odxtools

  odxtools.exceptions.strict_mode = False
  ```
- The `odxraise()` function raises an exception in strict mode and does nothing in non-strict mode. For type checkers, it looks like it always raises an exception.
- The `odxassert()` function is very similar to the build-in `assert`  statement, except that if the asserted condition is not met and if odxtools is in non-strict mode, program execution will continue. Unfortunately, this does not work well with type checkers as these do not consider the internal behavior of functions, so `assert` statements which's purpose is to narrow the type of a given object for the benefit of type checkers must be replaced by `if...odxraise()` constructs. E.g.
  ```python
  # prevent mypy from complaining
  assert isinstance(x, int)
  ```
  becomes
  ```python
  # prevent mypy from complaining
  if not isinstance(x, int): odxraise()
  ```
- In strict mode, the `odxrequire()` function raises an exception if its argument is `None`, in all other cases it passes through its argument without modifications. This allows to conveniently ensure that a given object was specified whilst convincing type checkers that it is not `None`. E.g., ensuring that a `SHORT-NAME` exists for an `ElementTree` node can be achieved using
  ```python
  ```

### DIFF
--- a/README.md
+++ b/README.md
@@ -200,6 +200,30 @@ python3 -m odxtools list -a "$YOUR_PDX_FILE"
   # -> decoded response: [session()]
   ```
 
+## Using the non-strict mode
+
+By default, odxtools raises exceptions if it suspects that it cannot
+fulfill a requested operation correctly. For example, if the dataset
+it is instructed to load is detected to be not conformant with the ODX
+specification, or if completing the operation requires missing
+features of odxtools. To be able to deal with such cases, odxtools
+provides a "non-strict" mode where such issues are ignored, but where
+the results are undefined. The following snippet shows how to instruct
+odxtools to load a non-conforming file in non-strict mode, and after
+this is done, enables the safety checks again:
+
+  ```python
+  import odxtools
+
+  [...]
+
+  odxtools.exceptions.strict_mode = False
+  botched_db = odxtools.load_file("my_non-conforming_database.pdx")
+  odxtools.exceptions.strict_mode = True
+
+  [...]
+  ```
+
 ## Interactive Usage
 
 ### Python REPL
--- a/examples/somersaultecu.py
+++ b/examples/somersaultecu.py
@@ -647,7 +647,8 @@ somersault_positive_responses = {
                         description=None,
                         diag_coded_type=somersault_diagcodedtypes["uint8"],
                         byte_position=0,
-                        coded_value=uds.positive_response_id(SID.ForwardFlip.value),  # type: ignore
+                        coded_value=uds.positive_response_id(
+                            SID.ForwardFlip.value),  # type: ignore[attr-defined]
                         bit_position=None,
                         sdgs=[],
                     ),
--- a/odxtools/admindata.py
+++ b/odxtools/admindata.py
@@ -100,8 +100,6 @@ class CompanyRevisionInfo:
 
     @property
     def company_data(self) -> CompanyData:
-        if self._company_data is None:
-            odxraise()
         return self._company_data
 
     @staticmethod
--- a/odxtools/admindata.py
+++ b/odxtools/admindata.py
@@ -5,6 +5,7 @@ from typing import TYPE_CHECKING, Any, Dict, List, Optional
 from xml.etree import ElementTree
 
 from .companydata import CompanyData, TeamMember
+from .exceptions import odxraise, odxrequire
 from .nameditemlist import NamedItemList
 from .odxlink import OdxDocFragment, OdxLinkDatabase, OdxLinkId, OdxLinkRef
 from .specialdata import SpecialDataGroup, create_sdgs_from_et
@@ -33,8 +34,8 @@ class CompanyDocInfo:
     def from_et(et_element: ElementTree.Element,
                 doc_frags: List[OdxDocFragment]) -> "CompanyDocInfo":
         # the company data reference is mandatory
-        company_data_ref = OdxLinkRef.from_et(et_element.find("COMPANY-DATA-REF"), doc_frags)
-        assert company_data_ref is not None
+        company_data_ref = odxrequire(
+            OdxLinkRef.from_et(et_element.find("COMPANY-DATA-REF"), doc_frags))
         team_member_ref = OdxLinkRef.from_et(et_element.find("TEAM-MEMBER-REF"), doc_frags)
         doc_label = et_element.findtext("DOC-LABEL")
         sdgs = create_sdgs_from_et(et_element.find("SDGS"), doc_frags)
@@ -99,15 +100,16 @@ class CompanyRevisionInfo:
 
     @property
     def company_data(self) -> CompanyData:
-        assert self._company_data is not None
+        if self._company_data is None:
+            odxraise()
         return self._company_data
 
     @staticmethod
     def from_et(et_element: ElementTree.Element,
                 doc_frags: List[OdxDocFragment]) -> "CompanyRevisionInfo":
 
-        company_data_ref = OdxLinkRef.from_et(et_element.find("COMPANY-DATA-REF"), doc_frags)
-        assert company_data_ref is not None
+        company_data_ref = odxrequire(
+            OdxLinkRef.from_et(et_element.find("COMPANY-DATA-REF"), doc_frags))
         revision_label = et_element.findtext("REVISION_LABEL")
         state = et_element.findtext("STATE")
 
@@ -148,8 +150,7 @@ class DocRevision:
         team_member_ref = OdxLinkRef.from_et(et_element.find("TEAM-MEMBER-REF"), doc_frags)
         revision_label = et_element.findtext("REVISION-LABEL")
         state = et_element.findtext("STATE")
-        date = et_element.findtext("DATE")
-        assert date is not None
+        date = odxrequire(et_element.findtext("DATE"))
         tool = et_element.findtext("TOOL")
 
         crilist = [
--- a/odxtools/companydata.py
+++ b/odxtools/companydata.py
@@ -4,6 +4,7 @@ from dataclasses import dataclass, field
 from typing import TYPE_CHECKING, Any, Dict, List, Optional
 from xml.etree import ElementTree
 
+from .exceptions import odxrequire
 from .nameditemlist import NamedItemList
 from .odxlink import OdxDocFragment, OdxLinkDatabase, OdxLinkId
 from .specialdata import SpecialDataGroup, create_sdgs_from_et
@@ -148,19 +149,12 @@ class TeamMember:
 
     @staticmethod
     def from_et(et_element: ElementTree.Element, doc_frags: List[OdxDocFragment]) -> "TeamMember":
-        odx_id = OdxLinkId.from_et(et_element, doc_frags)
-        assert odx_id is not None
-        short_name = et_element.findtext("SHORT-NAME")
-        assert short_name is not None
+        odx_id = odxrequire(OdxLinkId.from_et(et_element, doc_frags))
+        short_name = odxrequire(et_element.findtext("SHORT-NAME"))
         long_name = et_element.findtext("LONG-NAME")
         description = create_description_from_et(et_element.find("DESC"))
 
-        roles = list()
-        if (roles_elem := et_element.find("ROLES")) is not None:
-            for role_elem in roles_elem.iterfind("ROLE"):
-                role = role_elem.text
-                assert role is not None
-                roles.append(role)
+        roles = [odxrequire(role_elem.text) for role_elem in et_element.iterfind("ROLES/ROLE")]
 
         department = et_element.findtext("DEPARTMENT")
         address = et_element.findtext("ADDRESS")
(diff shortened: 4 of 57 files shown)


README.md
```diff
@@ -200,6 +200,30 @@ python3 -m odxtools list -a "$YOUR_PDX_FILE"
...
+this is done, enables the safety checks again:
+
+  ```python
+  import odxtools
+
+  [...]
+
+  odxtools.exceptions.strict_mode = False
+  botched_db = odxtools.load_file("my_non-conforming_database.pdx")
+  odxtools.exceptions.strict_mode = True
```
kayoub5, line 222: To avoid issues with threads and changing global variables side effect, it would be better to either 
* Make `strict_mode` an argument to `load_file`
* Use https://docs.python.org/3/library/warnings.html with default error for OdxWarning, and users can change behavior using https://docs.python.org/3/library/warnings.html#warning-filter

odxtools/admindata.py
```diff
@@ -99,15 +100,16 @@ class CompanyRevisionInfo:
 
     @property
     def company_data(self) -> CompanyData:
-        assert self._company_data is not None
+        if self._company_data is None:
+            odxraise()
         return self._company_data
```
kayoub5, line 105: Company data is a required element in XML?

odxtools/utils.py
```diff
@@ -34,8 +36,7 @@ def short_name_as_id(obj: Any) -> str:
     prepends an underscore to such such shortnames.
     """
 
-    sn = obj.short_name
-    assert isinstance(sn, str)
+    sn: str = odxrequire(obj.short_name)
```
kayoub5, line 39: checking something is not None and something is string is not the same thing

examples/somersaultecu.py
```diff
@@ -1780,7 +1781,7 @@ class SomersaultSID(IntEnum):
                 ProgCode(
                     code_file="jobs.jar",
                     encryption=None,
-                    syntax="JAR",
+                    syntax=ProgCodeSyntax.JAR,
```
kayoub5, line 1784: Unrelated to PR

odxtools/companydata.py
```diff
@@ -148,19 +149,12 @@ class TeamMember:
 
     @staticmethod
     def from_et(et_element: ElementTree.Element, doc_frags: List[OdxDocFragment]) -> "TeamMember":
-        odx_id = OdxLinkId.from_et(et_element, doc_frags)
-        assert odx_id is not None
-        short_name = et_element.findtext("SHORT-NAME")
-        assert short_name is not None
+        odx_id = odxrequire(OdxLinkId.from_et(et_element, doc_frags))
+        short_name = odxrequire(et_element.findtext("SHORT-NAME"))
```
kayoub5, line 153: odxrequire checks that element is not None, but an empty string is not None, but is still not a valid short_name

odxtools/exceptions.py
```diff
@@ -1,13 +1,14 @@
 # SPDX-License-Identifier: MIT
 # Copyright (c) 2022 MBition GmbH
 import warnings
+from typing import TYPE_CHECKING, NoReturn, Optional, Type, TypeVar
 
 
 class OdxError(Exception):
```
kayoub5, line 7:
```suggestion
class OdxError(Warning, Exception):
```

odxtools/exceptions.py
```diff
@@ -19,5 +20,61 @@ class OdxWarning(Warning):
     """Any warning that happens during interacting with diagnostic objects."""
 
 
-# Mark DecodeError warnings as errors by default
+# Mark DecodeError and EncodeErrpr warnings as errors by default
 warnings.simplefilter("error", DecodeError)
+warnings.simplefilter("error", EncodeError)
```
kayoub5, line 25:
```suggestion
warnings.simplefilter("error", EncodeError)
warnings.simplefilter("error", OdxError)
```

odxtools/exceptions.py
```diff
@@ -19,5 +20,61 @@ class OdxWarning(Warning):
...
+    """
+    Raise an exception but only if in strict mode.
+
+    Also, convince type checkers that the exception is always raised.
+    """
+    if TYPE_CHECKING or strict_mode:
+        if message is None:
+            raise error_type()
+        else:
+            raise error_type(message)
```
kayoub5, line 45: Current approach will make all error appear as if they originated here, caller should be the one where the first line number show up, use `stacklevel` argument to reflect this
```suggestion
            warnings.warn(message, error_type, stacklevel=2)
```

odxtools/exceptions.py
```diff
@@ -19,5 +20,61 @@ class OdxWarning(Warning):
...
+
+#: Specify whether violations of the ODX specification or failed
+#: assumptions in the ODX library code should cause the routine in
+#: question to abort or whether it should continue anyway. Be aware that
+#: in non-strict mode, the behavior of odxtools is undefined and it
+#: might start eating children...
+strict_mode = True
+
+
+def odxraise(message: Optional[str] = None, error_type: Type[Exception] = OdxError) -> NoReturn:
```
kayoub5, line 35: message should always be included, without a message, it would be very difficult to know what went wrong

Not all applications preserve the call stack.

odxtools/nameditemlist.py
```diff
@@ -69,10 +66,19 @@ def append(self, item: T):
...
         return self._list.sort(key=key, reverse=reverse)
 
-    def __contains__(self, x) -> bool:
+    def keys(self) -> Iterable[str]:
+        return self.__dict__.keys()
+
+    def values(self) -> Iterable[T]:
+        return self.__dict__.values()
+
+    def items(self) -> Iterable[Tuple[str, T]]:
```
kayoub5, line 78: * You re-invented the dict, subclass `collections.UserDict` instead
* Those changes are not related to original PR, please move them to their own PR

odxtools/nameditemlist.py
```diff
@@ -69,10 +66,19 @@ def append(self, item: T):
             else:
                 tmp = f"{item_name}_{i}"
 
-    def sort(self, key=None, reverse=False):
+    def sort(self, key: Optional[Callable[[T], str]] = None, reverse: bool = False):
         return self._list.sort(key=key, reverse=reverse)
 
-    def __contains__(self, x) -> bool:
+    def keys(self) -> Iterable[str]:
```
kayoub5, line 72: What happens if someone have a parameter called "keys" ?

README.md
andlaus: I considered this, but that would mean to drag the `strict` argument through to all constructors, `from_et()` functions and `odxraise()`, `odxassert()` and `odxrequire()` making them *very* clunky IMO. Also, I came to the conclusion that using one thread in strict mode and another in non-strict mode is quite a corner case...

> Use https://docs.python.org/3/library/warnings.html with default error for OdxWarning, and users can change behavior using https://docs.python.org/3/library/warnings.html#warning-filter

does this work with type checking? (making mypy happy is the reason for most of these assertations...)

odxtools/admindata.py
andlaus: within COMPANY-REVISON-INFO, yes: COMPANY-DATA-REF is mandatory and if the reference is not resolved, the object has not been properly initialized. the check should IMO be removed, though: we will get an `AttributeException` if something went wrong, but on most occasions where the same pattern applies we do not check for this either. (that said, this would categorize as drive-by fix. ;))

odxtools/companydata.py
andlaus: yes, ensuring that all content strings of the XML are valid would be another (quite large) PR because this would affect a substantial part of the XML tags containing "primitive" objects. I tried not to commingle too much into one here. (I did not always succeed.)

README.md
kayoub5: * Yes it works with type checking
* I left you a few review comments on how it should look like in `odxtools/exceptions.py`, the only change from your current code would be

Usage for end user
```py
  with warnings.filterwarnings("ignore", OdxError): # Could be "once", "default", "always"
      botched_db = odxtools.load_file("my_non-conforming_database.pdx")
```

examples/somersaultecu.py
andlaus: true. It's IMO too small to warrant its own PR, so I at least moved this into a separate patch...

README.md
andlaus: ok, I'll have a look. BTW: this also seems to basically be a global variable to me?

odxtools/exceptions.py
andlaus: I agree in principle, but that would be *a lot* of work. For now, the primary intention is to be able to look in a debugger what went wrong...

odxtools/exceptions.py
andlaus: > use stacklevel argument to reflect this

thanks for the hint, I wasn't aware of this so far :)

odxtools/nameditemlist.py
andlaus: it must be accessed via `.keys_` (same as python keywords like `None`, `global`, etc ...)

odxtools/nameditemlist.py
andlaus: Good point, but this class is also (primarily?) a list, so it also needs to inherit from `UserList`. I thus tried

```python
class NamedItemList(UserDict[str, T], UserList[T]):
  [...]
```

But this produced mypy errors a la
```
odxtools/nameditemlist.py:11: error: Definition of "data" in base class "UserDict" is incompatible with definition in base class "UserList"  [misc]
```

any idea?

odxtools/utils.py
andlaus: good point. fixed.

odxtools/nameditemlist.py
kayoub5: Could you move those changes to their own PR? we can discuss there possible solutions

odxtools/nameditemlist.py
kayoub5: But `keys` is not a keyword

README.md
andlaus: I now have an (almost) working version of this, but I've got very mixed feelings: On one hand, it feels more standard to do this using the `warnings` module, on the other hand, it is much more brittle because the unit testing modules (`unittest` and `pytest`) seem to regard the `warnings` module as their personal property and thus mess with it pretty freely. In particular `unittest`: It modifies `warnings`'s filters *by default*, and one explicitly has to  tell it to leave `warnings` alone (i.e., `unittest.main(warnings=???)`. unfortunately, I haven't yet figured out what it expects for the `warnings` argument). `pytest` is a bit less offensive as it only messes with the filters if its config says so. (ours currently does which confused me for quite a while.)

@kayoub5: Do you think going the `warnings` route is worth it  anyway? if yes, do you know how to convince `unittests` to leave the filters of `warnings` alone?

odxtools/exceptions.py
```diff
@@ -38,11 +32,13 @@ def odxraise(message: Optional[str] = None, error_type: Type[Exception] = OdxErr
...
-            raise error_type()
-        else:
-            raise error_type(message)
+    if TYPE_CHECKING:
+        raise error_type()
+
+    if message is None:
+        warnings.warn("ODX conformity error detected", error_type, stacklevel=2)
+    else:
+        warnings.warn(message, error_type, stacklevel=2)
```
kayoub5, line 41: Wrappers of `odxraise`, such as `odxassert` and `odxrequire` need a different stacklevel=3

odxtools/nameditemlist.py
andlaus: done. I figured that the patch can be removed completely. the `ProgCodeSyntax` patch now gets its separate PR...

README.md
andlaus: > Can be provided in the unittest.main https://python.readthedocs.io/en/stable/library/unittest.html#unittest.main

yes, but what must it be set to? setting `warning="error"` will break some tests because some currently non-fatal warnings will become raised exceptions, whilst setting it to "ignore" (or not specifying it at all) makes those tests fail which expect an `OdxError` to be raised. (The `unittest` module seems to always set some catch-all warning filters and it seems like it cannot be instructed to leave these filters alone...)

tl;dr: I think we should stick with a "custom" global variable instead of trying to use that of the `warnings` module, at least for now...

kayoub5: You can specify per type `error::OdxError`

See https://docs.python.org/3/library/warnings.html#default-warning-filter

andlaus: yes, but not with `unittest` (at least not without some severe hack), because `unittest` always adds some catch-all filters to the beginning of the list...

Problems like this are probably the reason why global variables ought to be avoided (i.e., multiple separate parts of the code fighting for supremacy over it)...

kayoub5: Rest of the code looks OK, if you think that using `warnings` is not worth the effort, then LGTM

kayoub5: > yes, but not with unittest (at least not without some severe hack), because unittest always adds some catch-all filters to the beginning of the list...

That's one of the many reasons people abandoned unittest and switched to pytest

andlaus: okay, let's get this merged then. Note that the same issue can happen with user code that using odxtools: if it modifies the filters of `warnings`, and if we relied `warnings` for controlling odxtools' behavior, the side effects could be very non-intuitive.

if you want to have a look at this issue yourself, check out [this branch](https://github.com/andlaus/odxtools/tree/WIP)...

PR #170, opened 2023-08-02
Followup fixes

andlaus:
These changes are an extended version of some of the "lateral" fixes that emerged during the review of https://github.com/mercedes-benz/odxtools/pull/168. in particular, I fixed a few issues with `NamedItemList`:

- Better type checking. Note that the `__iter__()` method is still not properly annotated, because doing so currently leads to *a lot* of typechecking errors.
- Add `keys()`, `values()` and `items()` methods. This makes `NamedItemList` objects more similar to normal dictionaries. Be aware that `.items()` is currently quite slow.
- Get rid of the `_typed_dict` attribute. We now use the untyped `__dict__` plus separate lists for the item names and values...
- Ensure that item names are valid before extending the list.
- Add unit test for `NamedItemList`


odxtools/nameditemlist.py
```diff
@@ -115,13 +149,13 @@ def __eq__(self, other: object) -> bool:
         if not isinstance(other, NamedItemList):
             return False
         else:
-            return self._list == other._list
+            return self._names == other._names and self._values == other._values and self._item_to_name_fn == other._item_to_name_fn
 
-    def __iter__(self):
-        return iter(self._list)
+    def __iter__(self):  # -> Iterator[T]: # <- this leads to *many* type checking errors
+        return iter(self._values)
```
kayoub5, line 155: this will be confusion to some users, it have a dict interface, but iterate like a list

odxtools/nameditemlist.py
```diff
@@ -55,57 +53,93 @@ def append(self, item: T):
...
+    @overload
+    def __getitem__(self, key: int) -> T:
+        ...
+
+    @overload
+    def __getitem__(self, key: str) -> T:
+        ...
+
+    @overload
+    def __getitem__(self, key: slice) -> List[T]:
```
kayoub5, line 113: use case for slice?

odxtools/nameditemlist.py
```diff
@@ -55,57 +53,93 @@ def append(self, item: T):
...
-        return len(self._list)
+    def __contains__(self, x: T) -> bool:
+        return x in self._values
 
-    def __getitem__(self, key: Union[int, str, slice]) -> T:
+    def __len__(self) -> int:
+        return len(self._values)
+
+    @overload
+    def __getitem__(self, key: int) -> T:
```
kayoub5, line 105: how would that work syntax wise? `foo.5`

odxtools/nameditemlist.py
```diff
@@ -55,57 +53,93 @@ def append(self, item: T):
...
+        return self._values
 
-    def __contains__(self, x) -> bool:
-        return x in self._list
+    def items(self) -> List[Tuple[str, T]]:
+        return list(zip(self._names, self._values))
 
-    def __len__(self):
-        return len(self._list)
+    def __contains__(self, x: T) -> bool:
```
kayoub5, line 98: dict contains logic is key based

odxtools/singleecujob.py
```diff
@@ -167,11 +167,17 @@ def dop(self) -> Optional[DopBase]:
         return self._dop
 
 
+class ProgCodeSyntax(Enum):
```
kayoub5, line 170: maybe use StrEnum?

odxtools/singleecujob.py
```diff
@@ -183,7 +189,14 @@ def from_et(et_element: Element, doc_frags: List[OdxDocFragment]) -> "ProgCode":
...
 
-        syntax = et_element.findtext("SYNTAX")
+        syntax = ProgCodeSyntax("JAR")
+        try:
+            syntax = ProgCodeSyntax(odxrequire(et_element.findtext("SYNTAX")))
+        except ValueError as e:
+            try:
+                odxraise(f"Could not parse program code syntax")
+            except Exception as ee:
+                raise ee from e
```
kayoub5, line 199: what's going on here?

odxtools/nameditemlist.py
andlaus: the class is primarily intended to be usable as a normal list with some extra features. (i.e., direct item access via attributes. That's why it's called `NamedItemList`, not `NamedItemDict`. ;) )

odxtools/nameditemlist.py
```diff
@@ -55,57 +53,93 @@ def append(self, item: T):
...
                 tmp = f"{item_name}{i}"
             else:
                 tmp = f"{item_name}_{i}"
+        item_name = tmp
+
+        self.__dict__[item_name] = item
+        self._names.append(item_name)
+        self._values.append(item)
+
+    def sort(self, key: Optional[Callable[[T], str]] = None, reverse: bool = False) -> None:
```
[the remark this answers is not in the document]
andlaus: to sort the list, a la

```python
foo = NamedItemList(lambda x: f"item_{x}", [5, 1, 7])
print(foo)
# [item_5, item_1, item_7]
foo.sort()
# [item_1, item_5, item_7]
```

odxtools/nameditemlist.py
andlaus: example:

```python
print(foo[1:])
# [5, 7]
```

odxtools/nameditemlist.py
andlaus: not for lists. we want
```python
2 in foo
# False
```
to work...

odxtools/singleecujob.py
andlaus: that seems to be not a part of the python standard library. While in this case, it would slightly better than `Enum`, adding a dependency is not worth it IMO...

odxtools/singleecujob.py
andlaus: we want to raise an `OdxException` instead of `ValueError` while keeping the traceback of the original exception, but we only want to throw anything if we are in strict mode. Maybe odxraise could be extended so that an alternative traceback can be specified, but that would probably not really make things more intuitive?

odxtools/nameditemlist.py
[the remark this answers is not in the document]
kayoub5: I mean why would the user of library need to change the order of the NamedItemList ?

There is a difference between getting a sorted list (using `sorted` function), and changing the order inside the list itself

odxtools/singleecujob.py
kayoub5: It is, but added in Python 3.11

https://docs.python.org/3.11/library/enum.html#enum.StrEnum

odxtools/singleecujob.py
kayoub5: a simple `if value in ProgCodeSyntax.__members__:` would have been enough

odxtools/nameditemlist.py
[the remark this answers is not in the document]
andlaus: > I mean why would the user of library need to change the order of the NamedItemList ?

not sure. it might come in handy at some point?

> There is a difference between getting a sorted list (using sorted function), and changing the order inside the list itself

yes: using `sorted()` makes a normal list, i.e., the ability to access items by name gets lost. `.sort()` gets around this. That said,
```python
foo = NamedItemList(lambda x: x, ["a", "c", "b"])
foo.sort()
```
and
```python
foo = NamedItemList(lambda x: x, ["a", "c", "b"])
foo = NamedItemList(lambda x: x, sorted(foo))
```

should be equivalent...

odxtools/singleecujob.py
andlaus: good point. fixed!

PR #171, opened 2023-08-04
Clean up the submodule structure and the import statements

andlaus:
This commit features no functional changes but quite a bit of "long and boring" housekeeping work:

- Most classes get separate files with the same name as the class name but all-lowercase. This slightly reduces the mental burden while developing because the class context of a given function is always clear from the file name. (e.g., no file should contain more than one `.from_et()` function.)
- Underscores in file names are (mostly) removed to make all files adhere to the naming convention.
- The import statements are cleaned up: All imported classes or functions should now be used by the importing submodule. (except for the imports of `__init__.py` which are for the benefit of the importing code.)
- The "Copyright (c) 202? MBition Gmbh" comments at the beginning of most files are removed. These comments have no legal significance anyway and they might deter people from modifying these files because of perceived "ownership"...
- A few `.from_et()` constructors get type annotations.


kayoub5:
> The "Copyright (c) 202? MBition Gmbh" comments at the beginning of most files are removed. These comments have no legal significance anyway and they might deter people from modifying these files because of perceived "ownership"...

Are you sure that? see https://opensource.stackexchange.com/a/300

examples/isotp_send.py
```diff
@@ -1,7 +1,6 @@
 #! /usr/bin/python3
```
kayoub5, line 1: This line only makes sense if the file is an executable script (for example in the `__main__.py` file), otherwise it's just noise

odxtools/__init__.py
```diff
@@ -63,24 +62,9 @@
...
 from .load_file import load_file
 from .load_odx_d_file import load_odx_d_file
 from .load_pdx_file import load_pdx_file
-from .parameters import (CodedConstParameter, DynamicParameter, LengthKeyParameter,
-                         MatchingRequestParameter, PhysicalConstantParameter, ReservedParameter,
-                         SystemParameter, TableEntryParameter, TableKeyParameter,
-                         TableStructParameter, ValueParameter)
-from .service import DiagService
-from .singleecujob import SingleEcuJob
-from .structures import Request, Response
```
kayoub5, line 83: This is a breaking change, you have users who do `from odxtools import Request`

kayoub5:
> This commit features no functional changes

Changing how a class gets imported is a functional breaking change for package consumers

andlaus:
> Are you sure that "These comments have no legal significance" ? only a lawyer can say that with confidence.
see https://opensource.stackexchange.com/a/300

I can't claim to know this for for all relevant jusristictions, but for Germany I'm pretty sure: There, the copyright ("Urheberrecht") of a work is always with whomever created the work, regardless of where the work was created. This right is not even transferable, only the commercial rights ("Verwertungsrechte") are. That said, in Germany the copyright itself is basically just the moral right to claim that you created a given work. When it comes to the commercial rights, these are usually covered by some contract between the creator and the holder of these rights (e.g., by means of an employment contract) and notices in the files have no influence on these contractual regulations.

(Also, for MIT-licensed projects like odxtools, who holds the commercial rights does not matter in practice anyway...)

odxtools/__init__.py
andlaus: yes. Let's make the next release 5.0.0...

andlaus:
> Changing how a class gets imported is a functional breaking change for package consumers

yes, I meant that the logic of these classes does not change, only their location...

kayoub5 (approved):

PR #172, opened 2023-08-07
Convert Literal string to Enum classes

andlaus:
Since the values of these objects are read from the XML (i.e., at runtime), using `typing.Literal` for them does not work very well: While loading a dataset it is currently not verified that the specified values are valid, and mypy must be instructed to ignore converting the dynamic string to the literal strings. Using `Enum` classes fixes these issues...

### DIFF
--- a/odxtools/singleecujob.py
+++ b/odxtools/singleecujob.py
@@ -52,7 +52,6 @@ class SingleEcuJob:
     odx_id: OdxLinkId
     short_name: str
     prog_codes: List[ProgCode]
-    """Pointers to the code that is executed when calling this job."""
     # optional xsd:elements inherited from DIAG-COMM (and thus shared with DIAG-SERVICE)
     oid: Optional[str]
     long_name: Optional[str]
--- a/odxtools/singleecujob.py
+++ b/odxtools/singleecujob.py
@@ -1,13 +1,14 @@
 # SPDX-License-Identifier: MIT
 from dataclasses import dataclass
+from enum import Enum
 from itertools import chain
-from typing import TYPE_CHECKING, Any, Dict, List, Literal, Optional, Union
+from typing import TYPE_CHECKING, Any, Dict, List, Optional, Union
 from xml.etree.ElementTree import Element
 
 from .admindata import AdminData
 from .audience import Audience
 from .createsdgs import create_sdgs_from_et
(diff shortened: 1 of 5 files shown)


odxtools/singleecujob.py
```diff
@@ -137,12 +137,18 @@ def from_et(et_element: Element, doc_frags: List[OdxDocFragment]) -> "SingleEcuJ
...
         is_mandatory_raw = odxstr_to_bool(et_element.get("IS-MANDATORY"))
         is_executable_raw = odxstr_to_bool(et_element.get("IS-MANDATORY"))
         is_final_raw = odxstr_to_bool(et_element.get("IS-FINAL"))
 
-        sdgs = create_sdgs_from_et(et_element.find("SDGS"), doc_frags)
+        diag_class_type: Optional[DiagClassType] = None
+        if (diag_class_type_str := et_element.get("DIAGNOSTIC-CLASS")) is not None:
+            if diag_class_type_str not in DiagClassType.__members__:
+                odxraise(f"Encountered unknown diagnostic class type '{diag_class_type_str}'")
+            diag_class_type = DiagClassType(diag_class_type_str)
```
kayoub5, line 151: `DiagClassType(diag_class_type_str)` will throw even if `odxraise` does not

odxtools/unitgroup.py
```diff
@@ -11,7 +12,10 @@
 if TYPE_CHECKING:
     from .diaglayer import DiagLayer
 
-UnitGroupCategory = Literal["COUNTRY", "EQUIV-UNITS"]
+
+class UnitGroupCategory(Enum):
+    COUNTRY = "COUNTRY"
+    EQUIV_UNITS = "EQUIV-UNITS"
```
kayoub5, line 18: Are you sure Category only accepts those values?

odxtools/singleecujob.py
andlaus: yes, but when using the non-strict mode, all bets are off: In non-strict mode there are plenty of instances where methods are called for objects that can be `None` in principle (e.g., [here](https://github.com/mercedes-benz/odxtools/blob/main/odxtools/statetransition.py#L45-L46))...

odxtools/unitgroup.py
andlaus: I just checked. it is weird IMO, but true:
image

PR #173, opened 2023-08-07
Add type annotations to all `from_et()` and `create_*_from_et()` functions

andlaus:
Also, remove the spurious duplicate `create_*_from_et()` functions which were forgotten to be deleted from the non-`create*.py` files during the recent restructuring.

### DIFF
--- a/odxtools/compumethods/createanycompumethod.py
+++ b/odxtools/compumethods/createanycompumethod.py
@@ -1,8 +1,9 @@
 # SPDX-License-Identifier: MIT
 import warnings
 from typing import Any, Dict, List
+from xml.etree import ElementTree
 
-from ..exceptions import OdxWarning, odxassert
+from ..exceptions import OdxWarning, odxassert, odxraise, odxrequire
 from ..globals import logger
 from ..odxlink import OdxDocFragment
 from ..odxtypes import DataType
@@ -89,8 +90,8 @@ def _parse_compu_scale_to_linear_compu_method(
     return LinearCompuMethod(**kwargs)
 
 
-def create_any_compu_method_from_et(et_element, doc_frags: List[OdxDocFragment],
-                                    internal_type: DataType,
+def create_any_compu_method_from_et(et_element: ElementTree.Element,
+                                    doc_frags: List[OdxDocFragment], internal_type: DataType,
                                     physical_type: DataType) -> CompuMethod:
     compu_category = et_element.findtext("CATEGORY")
     odxassert(compu_category in [
@@ -120,7 +121,7 @@ def create_any_compu_method_from_et(et_element, doc_frags: List[OdxDocFragment],
 
     if compu_category == "TEXTTABLE":
         odxassert(physical_type == DataType.A_UNICODE2STRING)
-        compu_internal_to_phys = et_element.find("COMPU-INTERNAL-TO-PHYS")
+        compu_internal_to_phys = odxrequire(et_element.find("COMPU-INTERNAL-TO-PHYS"))
 
         internal_to_phys: List[CompuScale] = []
         for scale in compu_internal_to_phys.iterfind("COMPU-SCALES/COMPU-SCALE"):
@@ -128,9 +129,9 @@ def create_any_compu_method_from_et(et_element, doc_frags: List[OdxDocFragment],
             upper_limit = Limit.from_et(scale.find("UPPER-LIMIT"), internal_type=internal_type)
 
             if (vt := scale.find("COMPU-INVERSE-VALUE/VT")) is not None:
-                compu_inverse_value = internal_type.from_string(vt.text)
+                compu_inverse_value = internal_type.from_string(odxrequire(vt.text))
             elif (v := scale.find("COMPU-INVERSE-VALUE/V")) is not None:
-                compu_inverse_value = internal_type.from_string(v.text)
+                compu_inverse_value = internal_type.from_string(odxrequire(v.text))
             else:
                 compu_inverse_value = None
 
@@ -142,7 +143,7 @@ def create_any_compu_method_from_et(et_element, doc_frags: List[OdxDocFragment],
                     lower_limit=lower_limit,
                     upper_limit=upper_limit,
                     compu_inverse_value=compu_inverse_value,
-                    compu_const=scale.find("COMPU-CONST").findtext("VT"),
+                    compu_const=odxrequire(scale.findtext("COMPU-CONST/VT")),
                 ))
 
         kwargs["internal_to_phys"] = internal_to_phys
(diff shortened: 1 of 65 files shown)


odxtools/compumethods/createanycompumethod.py
```diff
@@ -168,19 +169,27 @@ def create_any_compu_method_from_et(et_element, doc_frags: List[OdxDocFragment],
...
+        physical_points = []
+        for scale_elem in et_element.iterfind("COMPU-INTERNAL-TO-PHYS/COMPU-SCALES/COMPU-SCALE"):
+            internal_point = internal_type.from_string(
+                odxrequire(scale_elem.findtext("LOWER-LIMIT")))
+            physical_point = physical_type.from_string(
+                odxrequire(scale_elem.findtext("COMPU-CONST/V")))
+
+            if not isinstance(internal_point, (float, int)):
+                odxraise()
+            if not isinstance(physical_point, (float, int)):
```
kayoub5, line 182: What are you basing the assumption that physical and internal points can only be float or int?

odxtools/diagnostictroublecode.py
```diff
@@ -18,7 +20,7 @@ class DiagnosticTroubleCode:
     short_name: Optional[str]
     text: Optional[str]
     display_trouble_code: Optional[str]
-    level: Union[bytes, None]
+    level: Union[int, None]
```
kayoub5, line 23: Are you sure about this change?

odxtools/diagnostictroublecode.py
```diff
@@ -27,14 +29,15 @@ def is_temporary(self) -> bool:
...
+                doc_frags: List[OdxDocFragment]) -> "DiagnosticTroubleCode":
         if et_element.find("DISPLAY-TROUBLE-CODE") is not None:
             display_trouble_code = et_element.findtext("DISPLAY-TROUBLE-CODE")
         else:
             display_trouble_code = None
 
-        if et_element.find("LEVEL") is not None:
-            level = et_element.findtext("LEVEL")
+        if (level_str := et_element.findtext("LEVEL")) is not None:
+            level = int(level_str)
```
kayoub5, line 40: level_str can be in hex?

odxtools/parameters/createanyparameter.py
```diff
@@ -212,20 +221,23 @@ def create_any_parameter_from_et(et_element, doc_frags):
...
+        if (table_snref_elem := et_element.find("TABLE-SNREF")) is not None:
+            table_snref = odxrequire(table_snref_elem.get("SHORT-NAME"))
+        else:
+            table_snref = None
+
+        table_row_ref = OdxLinkRef.from_et(et_element.find("TABLE-ROW-REF"), doc_frags)
+        if (table_row_snref_elem := et_element.find("TABLE-ROW-SNREF")) is not None:
+            table_row_snref = odxrequire(table_row_snref_elem.get("SHORT-NAME"))
+        else:
+            table_row_snref = None
```
kayoub5, line 233: maybe a helper function to read SNREF?

odxtools/physicaltype.py
```diff
@@ -56,21 +57,21 @@ def __post_init__(self):
...
-            precision = int(precision)
+    def from_et(et_element: ElementTree.Element, doc_frags: List[OdxDocFragment]):
+        base_data_type_str = et_element.get("BASE-DATA-TYPE")
+        if base_data_type_str not in DataType.__members__:
+            odxraise(f"Encountered unknown base data type '{base_data_type_str}'")
+        base_data_type = DataType(base_data_type_str)
+
+        display_radix_str = et_element.get("DISPLAY-RADIX")
+        if display_radix_str is not None:
+            if display_radix_str not in Radix.__members__:
```
kayoub5, line 68: It would help more if radix was an int, 10 for decimal, 16 for hex.. etc

odxtools/compumethods/createanycompumethod.py
andlaus: it is pretty hard to interpolate between anything else? (what is the middle between the strings "foo" and "bar"?)

odxtools/diagnostictroublecode.py
andlaus: quite: the XSD says its an `xsd:unsignedByte`, which -- if I understand this correctly -- just an integer in the 0-255 range...

odxtools/diagnostictroublecode.py
andlaus: I don't think so: https://www.data2type.de/xml-xslt-xslfo/xml-schema/datentypen-referenz/xs-byte (sorry, I couldn't quickly find a good English source)

odxtools/parameters/createanyparameter.py
andlaus: good point, but let's do this in another PR...

odxtools/compumethods/createanycompumethod.py
kayoub5: bytes and strings can be compared (for equality and ordering), python natively support that, if they can be compared, then they can be used for table based compu method

andlaus:
okay. I couldn't resist to fix most non-spurious complaints produced by the `ruff` linter: https://github.com/mercedes-benz/odxtools/pull/173/commits/db3ed9f582e8420063c61459fff98d2be8b607b7 (these are "collateral fixes", sorry for that)

odxtools/compumethods/createanycompumethod.py
andlaus: yes, but the "TAB-INTP" compu methods interpolates between the specified points. e.g. if points (0, 0) and (2, pi) are specified by the PDX file, encountering a 1 while decoding would be mapped to pi/2. (I might be wrong, I haven't actually implemented this code...)

odxtools/physicaltype.py
kayoub5: We could make it so that the enum is an IntEnum, that won't be breaking the spec

andlaus: how to create this object from a string? i.e. something like this, but working:
```python
class Radix(IntEnum):
  DEC = 10
  HEX = 16

r = Radix("HEX")
```

andlaus: okay, found a solution: 9387b34e49cb35176811b3e26654fa7bad1da866 . (thanks for your persistence)

note though, that this solution only works as long as all members are valid python identifiers, in particular this approach won't work if the names contain a minus...

odxtools/progcode.py
```diff
@@ -34,9 +34,10 @@ def from_et(et_element: ElementTree.Element, doc_frags: List[OdxDocFragment]) ->
         encryption = et_element.findtext("ENCRYPTION")
 
         syntax_str = odxrequire(et_element.findtext("SYNTAX"))
-        if syntax_str not in ProgCodeSyntax.__members__:
+        try:
+            syntax = ProgCodeSyntax(syntax_str)
+        except ValueError:
             odxraise(f"Encountered unknown program code syntax '{syntax_str}'")
```
kayoub5, line 40: You need to add `syntax = None` inside except, otherwise the lines below will raise `NameError`

Same for other changes in this commit

andlaus: right. next try: df6db1118c9c947e0accce14eb968667a678ea53

kayoub5 (approved):

PR #177, opened 2023-08-10
make the parameters data classes

andlaus:
The second part of the "use `@dataclass` where appropriate" quest...


odxtools/parameters/codedconstparameter.py
```diff
@@ -15,15 +16,15 @@
...
-        super().__init__(parameter_type="CODED-CONST", **kwargs)
+    diag_coded_type: DiagCodedType
+    coded_value: AtomicOdxType
 
-        self.diag_coded_type = diag_coded_type
-        if not isinstance(coded_value, (int, bytes, bytearray)):
-            odxraise()
-        self.coded_value = coded_value
+    @property
+    def parameter_type(self) -> str:
```
kayoub5, line 26: TIP:

Consider using a `Literal` typehint for return value

odxtools/parameters/parameterwithdop.py
```diff
@@ -17,22 +18,13 @@
...
-    ) -> None:
-        super().__init__(parameter_type=parameter_type, **kwargs)
-        self.dop_ref = dop_ref
-        self.dop_snref = dop_snref
-
-        self._dop: Optional[DopBase] = None
-        if dop_snref is None and dop_ref is None:
+    def __post_init__(self) -> None:
+        if self.dop_snref is None and self.dop_ref is None:
             logger.warn(f"Param {self.short_name} without DOP-(SN)REF should not exist!")
```
kayoub5, line 28: Why not use warnings module?

kayoub5: The warning should indicate to what parameter it is referring

odxtools/parameters/physicalconstantparameter.py
```diff
@@ -57,7 +80,7 @@ def decode_from_pdu(self, decode_state: DecodeState):
         return phys_val, next_byte_position
 
     def __repr__(self) -> str:
-        repr_str = f"PhysicalConstantParameter(short_name='{self.short_name}', physical_constant_value={self.physical_constant_value}"
+        repr_str = f"PhysicalConstantParameter(short_name='{self.short_name}', physical_constant_value={self.physical_constant_value!r}"
```
kayoub5, line 83: What's the r for at the end?

odxtools/parameters/reservedparameter.py
```diff
@@ -1,27 +1,35 @@
...
     def is_optional(self):
         return False
 
+    @property
+    def bit_length(self) -> int:
+        # this is a bit hacky: the parent class already specifies
+        # bit_length as a function property, and we need to change
+        # this to return the value from the XML here. Since function
+        # attributes cannot be overridden by non-function ones, we
+        # need to take the "bit_length_raw" detour...
```
kayoub5, line 30: Another approach is to explicitly make an `__init__` in this class

tests/test_encoding.py
```diff
@@ -162,6 +162,7 @@ def test_encode_linear(self):
         )
 
         param1._resolve_odxlinks(odxlinks)
+        param1._resolve_snrefs(None)
```
kayoub5, line 165: Why?

tests/test_decoding.py
```diff
@@ -1359,6 +1362,8 @@ def test_min_max_length_type_end_of_pdu_in_structure(self):
 
         req_param1._resolve_odxlinks(odxlinks)
         req_param2._resolve_odxlinks(odxlinks)
+        req_param1._resolve_snrefs(None)
+        req_param2._resolve_snrefs(None)
```
kayoub5, line 1366: Shouldn't calling the  resolve snrefs on the parent structure be enough?

odxtools/parameters/codedconstparameter.py
andlaus: good idea. implemented, thanks: 27a43512cd58203aa44d55aabaa2dded1e4cfd49

odxtools/parameters/parameterwithdop.py
andlaus: historical artifact. instead of `warnings`, this should use the use the standard odxtools error handling mechanism, though (i.e., `odxassert()`). fixed: b282cb8f7c1ac63e6635f139fbab9281b0a1add8

odxtools/parameters/physicalconstantparameter.py
andlaus: without it, `mypy` complains that the result of the f-string changed for `bytes` objects a few python versions ago (3.7?)

odxtools/parameters/reservedparameter.py
andlaus: yes, the problem with this is that I don't know how the constructor of the dataclass can be called. (I don't want to have to explicitly handle all attributes...)

tests/test_decoding.py
andlaus: it would be. The problem there is that this might require to construct a full `DiagLayer` object. (The code here is a pretty bad hack which takes advantage of the fact that these parameter classes do not actually access the passed `diag_layer` object...)

tests/test_encoding.py
andlaus: basically because of this: https://github.com/mercedes-benz/odxtools/pull/177/files/a4db1e5bb5af9caf07e75a4ff88190657e69a6e3#diff-a57adee126e4dd38d0a35068034b5c77a85e084fedbc394fa8418cb6330c1929R33-R40

there, conversion from a string to the internal value requires a DOP, but the DOP can be either referenced via ODXLINK or via SNREF and thus the conversion can only be done if we are sure that the DOP has been resolved (i.e., in `.resolve_snrefs()`). Before this patch the conversion was done at runtime very time the property was accessed. (drive-by fix, sorry for that...)

andlaus:
> If using dataclass, then the custom __repr__ can be removed, right?

good point. done: 202f76667c2b45059857e2a1f99a3d6a2ac0e384

kayoub5 (approved):

PR #178, opened 2023-08-11
Moar dataclasses, part 3

andlaus:
the third batch of dataclass conversions. this one converts quite a few classes, but they are all quite simple...

### DIFF
--- a/odxtools/diaglayercontainer.py
+++ b/odxtools/diaglayercontainer.py
@@ -153,9 +153,3 @@ class DiagLayerContainer:
 
     def __getitem__(self, key: Union[int, str]) -> DiagLayer:
         return self.diag_layers[key]
-
-    def __repr__(self) -> str:
-        return f"DiagLayerContainer('{self.short_name}')"
-
-    def __str__(self) -> str:
-        return f"DiagLayerContainer('{self.short_name}')"
--- a/odxtools/diaglayercontainer.py
+++ b/odxtools/diaglayercontainer.py
@@ -1,4 +1,5 @@
 # SPDX-License-Identifier: MIT
+from dataclasses import dataclass
 from itertools import chain
 from typing import List, Optional, Union
 from xml.etree import ElementTree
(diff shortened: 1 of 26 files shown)


odxtools/diaglayercontainer.py
```diff
@@ -71,26 +56,26 @@ def from_et(et_element: ElementTree.Element) -> "DiagLayerContainer":
...
-        ]
-        base_variants = [
+        ])
+        base_variants = NamedItemList(short_name_as_id, [
             DiagLayer.from_et(dl_element, doc_frags)
             for dl_element in et_element.iterfind("BASE-VARIANTS/BASE-VARIANT")
-        ]
-        ecu_variants = [
+        ])
+        ecu_variants = NamedItemList(short_name_as_id, [
```
kayoub5, line 75: is short_name_as_id actually needed anymore?

odxtools/request.py
```diff
@@ -1,12 +1,12 @@
...
 from .basicstructure import BasicStructure
 
 
+@dataclass
 class Request(BasicStructure):
 
-    def __init__(self, **kwargs):
-        super().__init__(**kwargs)
-
     def __repr__(self) -> str:
```
kayoub5, line 10: useless repr, use dataclass auto implementation

odxtools/diaglayercontainer.py
andlaus: yes: without it, `NamedItemList` does not know how to map an item to a name. also, ODX short names can start with a number, so we cannot simply pass through the short name a la `lambda x: x.short_name`...

kayoub5: do you have actually a case where NamedItemList have another id mapper other than short_name_as_id?

odxtools/request.py
andlaus: right. fixed: 544249a51be18d56a3ee9f44bf40668becffc7ef

odxtools/diaglayercontainer.py
andlaus: not within odxtools. (I think that this class ought to be implemented generically, IMO). To make its usage a bit more convenient, we could introduce a `ShortNameItemList` class for this purpose, though...

kayoub5 (approved):

Issue #179, opened 2023-08-11
ModuleNotFoundError: No module named 'odxtools.compumethods' with v5.0.0


matthtran:
Installation using `pip` succeeds, but there's an error on import. Looking at the files and on https://pypi-browser.org/package/odxtools/odxtools-5.0.0.tar.gz, it appears the `compumethods` folder is not included in the package.

andlaus:
okay, I hope to have fixed this with version 5.0.4. If the issue persists, feel free to reopen it...

PR #180, opened 2023-08-14
add a MANIFEST.in

andlaus:
for whatever reason, the `setuptools` heuristic for automatically determining which files ought to be distributed broke down recently, so that all sub-directories of `odxtools/` were no longer part of the pypi package. Explicitly instructing `setuptools` what to include fixes that...


andlaus:
this hopefully closes #179 ...

andlaus:
since this is IMO pretty urgent and cannot break the pypi package even more than it already is, I take the liberty of merging this without prior review...

PR #181, opened 2023-08-14
migrate the build to `pyproject.toml`

andlaus:
This has been the standard way for python projects to configure build systems/CI tools/packaging systems for a while...


.github/workflows/tests.yml
```diff
@@ -51,8 +51,9 @@ jobs:
 
       - name: Install odxtools development and testing dependencies
         run: |
-          pip install -r requirements.txt
-          pip install -r test-requirements.txt
+          pip install --upgrade pip
+          pip install -e .
```
kayoub5, line 55: don't use editable mode in CI, you will miss bugs such as  #179 if you do

pyproject.toml
```diff
@@ -1,12 +1,64 @@
 [build-system]
 
 requires = [
-    "setuptools>=42",
+    "setuptools",
```
kayoub5, line 4: specify min version

pyproject.toml
```diff
@@ -1,12 +1,64 @@
...
+
+[project.scripts]
+odxtools = "odxtools.__init__:_main"
+
+[tool.setuptools]
+
+packages = [ "odxtools" ]
+
+[tool.setuptools.dynamic]
+version = {attr = "odxtools.version.__version__"}
```
kayoub5, line 60: You can't do this, and then put `setuptools-scm` in build requires

Tip, use setuptools-scm, and change version.py as per https://github.com/pypa/setuptools_scm docs

test-requirements.txt
```diff
@@ -1,4 +0,0 @@
-types-Deprecated>=1.2.9
-coveralls>=3.3.1
-mypy>=0.991
-pytest>=7.2.0
```
kayoub5, line 4: Why?

Tip: use poetry

pyproject.toml
```diff
@@ -1,12 +1,64 @@
...
+    'License :: OSI Approved :: MIT License',
+    'Operating System :: POSIX',
+]
+dependencies = [
+    "bitstruct >= 6.0.0",
+    "argparse_addons",
+    "PyInquirer",
+    "jinja2",
+    "python-can",
+    "can-isotp",
```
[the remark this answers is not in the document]
andlaus: `can` is needed by the snoop tool, but besides this there currently there is nothing. (the `snoop` tool needs to implement its own ISO-TP state machine because it must work in "passive mode"...)

A much more problematic dependency is `PyInquirer` (because it seems like it constantly like to break installations). Since this is only needed by the somewhat obscure `browse` tool, I put it into the `browse-tool` dependency group which is not selected by default...

.github/workflows/tests.yml
andlaus: okay. As you've probably noticed, I'm not familiar with this new-fanged `pyproject.toml` stuff, so I used the cantools files as a template. Anyway, it is good to follow the [official advise](https://pip.pypa.io/en/stable/topics/local-project-installs/#regular-installs), thanks for pointing this out!

pyproject.toml
andlaus: cool, having to modify `version.py` before each release is error-prone and annoying. To keep the changes here minimal, I'll remove the `setuptools_scm` dependency and do the switch in the next PR...

test-requirements.txt
andlaus: I've switched to dependency groups...

test-requirements.txt
```diff
@@ -1,4 +0,0 @@
-types-Deprecated>=1.2.9
-coveralls>=3.3.1
```
kayoub5, line 2: coverage command line missing, you need to add it to test dependencies, most likely was previously installed as dependency of coveralls

pyproject.toml
```diff
@@ -1,12 +1,77 @@
...
+requires-python = ">=3.8"
+keywords = ['can', 'can bus', 'DoIP', 'odx', 'pdx', 'obd', 'uds', 'automotive', 'diagnostics']
+classifiers = [
+    'Programming Language :: Python :: 3',
+    'Programming Language :: Python :: 3.8',
+    'Development Status :: 5 - Production/Stable',
+    'Environment :: Console',
+    'Intended Audience :: Developers',
+    'License :: OSI Approved :: MIT License',
+    'Operating System :: POSIX',
```
kayoub5, line 31: Maybe use "OS Independent" for "Operating System", the library is pure python with no native dependencies

test-requirements.txt
andlaus: oops. should be fixed. (until we actually use coveralls.io, I switched this to the `coverage` package, though.)

.gitignore
```diff
@@ -5,6 +5,7 @@ __pycache__/
 
 # Distribution / packaging
 *.egg-info/
+odxtools/version.py
```
kayoub5, line 8: Why?

pyproject.toml
```diff
@@ -70,8 +71,8 @@ odxtools = "odxtools.__init__:_main"
 [tool.setuptools]
 packages = [ "odxtools" ]
 
-[tool.setuptools.dynamic]
-version = {attr = "odxtools.version.__version__"}
+[tool.setuptools_scm]
+write_to = "odxtools/version.py"
```
kayoub5, line 75: write_to is deprecated, use version_file instead

pyproject.toml
andlaus: ok, done. the `snoop` tool needs `python-can`, but I think `python-can` is available on most platforms and I don't consider the command line tools to be part of the core library...

.gitignore
andlaus: failed experiment: I wanted git to ignore changes to the file even if it was tracked (i.e., so that the file is present even if `pip install -e .` has been forgotten, but changes to it are not shown in `git diff`. that doesn't work that easily with git. anyway, fixed: 18257b6cb1663e1ea326f3554d0701b0d741fa0b

pyproject.toml
andlaus: in this case, the [documentation](https://pypi.org/project/setuptools-scm/) is outdated as well. also, when using `version_file=` pip produces
```
TypeError: Configuration.__init__() got an unexpected keyword argument 'version_file'
```

kayoub5: version_file was added in setuptools_scm v8, what version are you using?

See https://github.com/pypa/setuptools_scm/blob/main/CHANGELOG.rst

kayoub5: Nevermind v8 of setuptools_scm has not been published yet

kayoub5 (approved):

PR #182, opened 2023-08-15
Remove short_name_as_id

kayoub5:


odxtools/nameditemlist.py
```diff
@@ -1,10 +1,51 @@
...
+    prepends an underscore to such such shortnames.
+    """
+    if not isinstance(obj, OdxNamed):
+        odxraise()
+    sn = obj.short_name
+    if not isinstance(sn, str):
+        odxraise()
+
+    if not sn.isidentifier():
+        warnings.warn(("For NamedItemList objects to work properly, all "
```
andlaus, line 35: possibly, it is better to change this `if` to `odxassert()`

odxtools/nameditemlist.py
```diff
@@ -1,10 +1,51 @@
...
+
+    # make sure that the name of the item in question is not a
+    # python keyword (this would lead to syntax errors)
+    if iskeyword(sn):
+        return f"{sn}_"
+
+    return sn
 
 
 class NamedItemList(Generic[T]):
```
andlaus, line 51: IMO, it would be better to leave `NamedItemList` as is and derives a `ShortNameItemList` which  hard-codes the usage of `.short_name` as item names. possibly, I'm missing an issue with this, though?

odxtools/database.py
```diff
@@ -9,9 +9,8 @@
 from .diaglayer import DiagLayer
 from .diaglayercontainer import DiagLayerContainer
 from .globals import logger
-from .nameditemlist import NamedItemList
+from .nameditemlist import NamedItemList, short_name_as_key
 from .odxlink import OdxLinkDatabase
-from .utils import short_name_as_id
 
 
 def version(v: str):
```
andlaus, line 16: this is a drive-by (sorry), but this function should probably become a static member of the `Database` class and be renamed to something like `parse_model_version()`?

andlaus (commented):
looks good. The only thing I'd really like to have is to keep the generic version of the class and derive the one which uses `.short_name` from it...

odxtools/nameditemlist.py
kayoub5: unless there is another sub class of NamedItemList other than ShortNameItemList, adding ShortNameItemList would be simply over engineering

andlaus: I would normally agree, but I'm thinking about using this class for other projects such as a residual bus simulator based on [cantools](https://github.com/cantools/cantools) where I need a different item name function.

(possibly, it is worth thinking about factoring this class out into a completely separate module...)

kayoub5: NamedItemList have a base class ItemList, that allows overloading the key getter

odxtools/database.py
kayoub5: Out of PR scope, current PR overloaded as is

andlaus:
ok, now I'm happy :). merging!

PR #183, opened 2023-08-15
Make the compumethods dataclasses

andlaus:
The next part of the quest for the database containing mostly dataclasses. This PR also contains a few drive-by fixes, in particular that we now have a proper class for COMPU-RATIONAL-COEFFS, the CompuScale and Limit classes no longer uses default values, and that CompuScale has a `.from_et()` method.

### DIFF
--- a/odxtools/compumethods/createanycompumethod.py
+++ b/odxtools/compumethods/createanycompumethod.py
@@ -66,7 +66,7 @@ def _parse_compu_scale_to_linear_compu_method(
         internal_type=internal_type,
     )
     if internal_lower_limit is None:
-        internal_lower_limit = Limit(float("-inf"), IntervalType.INFINITE)
+        internal_lower_limit = Limit(0, IntervalType.INFINITE)
     kwargs["internal_lower_limit"] = internal_lower_limit
 
     # Read upper limit
@@ -76,7 +76,7 @@ def _parse_compu_scale_to_linear_compu_method(
     )
     if internal_upper_limit is None:
         if not is_scale_linear:
-            internal_upper_limit = Limit(float("inf"), IntervalType.INFINITE)
+            internal_upper_limit = Limit(0, IntervalType.INFINITE)
         else:
             odxassert(internal_lower_limit is not None and
                       internal_lower_limit.interval_type == IntervalType.CLOSED)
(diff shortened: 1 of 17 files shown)


odxtools/compumethods/createanycompumethod.py
```diff
@@ -76,7 +76,7 @@ def _parse_compu_scale_to_linear_compu_method(
     )
     if internal_upper_limit is None:
         if not is_scale_linear:
-            internal_upper_limit = Limit(float("inf"), IntervalType.INFINITE)
+            internal_upper_limit = Limit(0, IntervalType.INFINITE)
```
kayoub5, line 79: Sure about this?

odxtools/compumethods/createanycompumethod.py
```diff
@@ -124,28 +124,13 @@ def create_any_compu_method_from_et(et_element: ElementTree.Element,
...
-                    lower_limit=lower_limit,
-                    upper_limit=upper_limit,
-                    compu_inverse_value=compu_inverse_value,
-                    compu_const=odxrequire(scale.findtext("COMPU-CONST/VT")),
-                ))
+                CompuScale.from_et(
+                    scale_elem, doc_frags, internal_type=internal_type,
+                    physical_type=physical_type))
 
+        kwargs["physical_type"] = DataType.A_UNICODE2STRING
```
kayoub5, line 133: A_UNICODE2STRING means 2 bytes per char, are you sure?

odxtools/parameterinfo.py
```diff
@@ -58,7 +58,7 @@ def parameter_info(param_list: Iterable[Union[Parameter, EndOfPduField]]) -> str
         if isinstance(cm, TexttableCompuMethod):
             result += f": enum; choices:\n"
             for scale in cm.internal_to_phys:
-                result += f"  '{scale.compu_const}'\n"
+                result += f"  '{str(scale.compu_const)}'\n"
```
kayoub5, line 61: Why?

odxtools/compumethods/createanycompumethod.py
andlaus: pretty much: it is impossible to convert an infinite `float` to `int` (try `int(float("inf"))`), and `Limit.complies_to_{upper,lower}()` ignore the value for `IntervalType.INFINITE` anyway...

odxtools/compumethods/createanycompumethod.py
andlaus: not really. this is littered across the codebase whenever a string is to be represented, though. probably we should change this to `A_UTF8STRING`... (let's see if something blows up)

odxtools/compumethods/createanycompumethod.py
kayoub5: what about signed int, 0 is not the min value in that case

odxtools/parameterinfo.py
andlaus: because mypy starts to complain about the change in behaviour for byte strings a few versions back. we also could use `f"{scale.compu_const!r}"` (what do you prefer?)

odxtools/compumethods/createanycompumethod.py
andlaus: okay, I checked: on the python side, `A_UNICODE2STRING` always maps to `str` (which, if I'm not mistaken, uses UTF8 internally), so there is no difference  declaring the physical type `A_UNICODE2STRING` compared to `A_UTF8STRING`. I think Katrin just picked `A_UNICODE2STRING` on a whim here, so let's stick with it for now...

odxtools/compumethods/createanycompumethod.py
andlaus: the problem with this is that in python integer classes are arbitrary length, so there is no analogon to INT_MAX in C...

odxtools/parameterinfo.py
andlaus: I agree, but we don't know what type `scale.compu_const` actually is (it can be the representation of any atomic physical value, i.e.,  `int`, `str`, `bytes` or `float`), so we need to use something that works for all of these...

andlaus: (since this is a TEXTTABLE, the actual type must most likely be `str`, but `mypy` can't know that...)

odxtools/compumethods/createanycompumethod.py
kayoub5: The trick is knowing the type bit_length
```python
# bit length
bit_length = 4

def unsigned_to_signed(value):
    sign_mask = 1 << (bit_length - 1));
    if (sign_mask & raw) == 0:
        return value
    return (value ^ sign_mask) - sign_mask

max_unsigned = (1 << bit_length) - 1 # 15
min_unsigend = 0

min_signed = unsigned_to_signed(1 << (bit_length - 1)) # -8
max_signed =unsigned_to_signed((1 << (bit_length - 1)) - 1) # 7
```

andlaus: what would be the point? IMO, passing `None` would better, but I think that allowing `None` for the value would regress the normal cases (as I see it, infinite limits are a corner case)...

kayoub5 (approved):

PR #184, opened 2023-08-15
Simplify common properties parsing

kayoub5:
* Use TypedDict to ensure that kwargs actually match class init functions (any deviations will be caught by mypy)
* Few fixes where common properties were missing in some classes


odxtools/outputparam.py
```diff
@@ -28,23 +24,12 @@ def __post_init__(self) -> None:
...
     def from_et(et_element: ElementTree.Element, doc_frags: List[OdxDocFragment]) -> "OutputParam":
 
-        odx_id = odxrequire(OdxLinkId.from_et(et_element, doc_frags))
-        short_name = odxrequire(et_element.findtext("SHORT-NAME"))
-        long_name = et_element.findtext("LONG-NAME")
-        description = create_description_from_et(et_element.find("DESC"))
+        kwargs = IdentifiableElement.get_kwargs(et_element, doc_frags)
         dop_base_ref = odxrequire(OdxLinkRef.from_et(et_element.find("DOP-BASE-REF"), doc_frags))
         semantic = et_element.get("SEMANTIC")
         oid = et_element.get("OID")
```
kayoub5, line 30: @andlaus I noticed that most of the time (there are like only two exceptions), when element have an ID, it also have an optional OID attribute

Refactoring of OID could be done in different PR

tests/test_decoding.py
```diff
@@ -1154,6 +1154,8 @@ def test_decode_dtc(self):
         dtc1 = DiagnosticTroubleCode(
             odx_id=OdxLinkId("dtcID1", doc_frags),
             short_name="P34_sn",
+            long_name=None,
+            description=None,
```
kayoub5, line 1158: Using a common parse function allowed to catch rogue cases like this one

odxtools/element.py
```diff
@@ -0,0 +1,53 @@
...
+
+
+class BaseElementKwargs(TypedDict):
+    short_name: str
+    long_name: Optional[str]
+    description: Optional[str]
+
+
+@dataclass
+class BaseElement:
```
andlaus, line 18: `NamedElement` would possibly be better name for this class?

andlaus (commented):
I like this. I'd prefer avoiding building keyword dictionaries and implement this using the `__init_from_et__()` route, though.

odxtools/additionalaudience.py
```diff
@@ -1,38 +1,26 @@
...
                 doc_frags: List[OdxDocFragment]) -> "AdditionalAudience":
 
-        odx_id = odxrequire(OdxLinkId.from_et(et_element, doc_frags))
-        short_name = odxrequire(et_element.findtext("SHORT-NAME"))
-        long_name = et_element.findtext("LONG-NAME")
-        description = create_description_from_et(et_element.find("DESC"))
-
-        return AdditionalAudience(
-            odx_id=odx_id, short_name=short_name, long_name=long_name, description=description)
+        return AdditionalAudience(**IdentifiableElement.get_kwargs(et_element, doc_frags))
```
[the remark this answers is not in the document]
kayoub5: Problem with `__init_from_et__` is that it does not check if a property was assigned twice, or was not assigned at all

The TypedDict solution gets checked for correctness with mypy

odxtools/basecomparam.py
```diff
@@ -27,11 +28,7 @@ class Usage(Enum):
 
 
 @dataclass
-class BaseComparam:
-    odx_id: OdxLinkId
-    short_name: str
-    long_name: Optional[str]
-    description: Optional[str]
```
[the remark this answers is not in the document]
kayoub5: That would mean that I will change the class to use `from_et` as well, and refactor other methods as well

That would be a big change for another PR

.gitignore
```diff
@@ -24,3 +24,4 @@ __pycache__/
 .venv
 env
 venv
+/odxtools/version.py
```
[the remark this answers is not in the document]
andlaus: good point. shouldn't most of the other entries in that file get a preceding slash then, though?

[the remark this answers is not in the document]
kayoub5: Possibly, but even without the slash it will work, since "**" will match current directory as well

odxtools/additionalaudience.py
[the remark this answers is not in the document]
andlaus: fair. The `__init_from_et__` pattern still feels more natural to me, but if the TypedDicts lead to better mypy coverage, let's go that route. That said, I think that There Can Be Only One (TM), so the current users of `__init_from_et__()` should be converted sooner or later.

BTW: how to "stack" these TypedDict objects? (i.e., The one for ComplexComparam would contain the one for BaseComparam which would contain the one for IdentifiableElement)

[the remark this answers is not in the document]
andlaus: okay: the "attribute missing" problem of the `__init_from_et__` pattern can be solved by calling

```python
def dataclass_is_initialized(dc_obj: Any) -> bool:
    for field in dc_obj.__dataclass_fields__.values():
        if not hasattr(dc_obj, field.name):
            return False

    return True
```

at the end of all `from_et()`  methods. That said, since this would be a runtime check, a static solution would be preferable IMO. (Also, it might still be possible to assign objects of the wrong type to the attributes, but I suppose that mypy will catch that (haven't checked).)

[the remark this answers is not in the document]
kayoub5: @andlaus check the new mechanism in the last commit, works with mypy 1.5

odxtools/element.py
```diff
@@ -0,0 +1,56 @@
...
+    short_name: str
+    long_name: Optional[str]
+    description: Optional[str]
+
+
+@dataclass
+class NamedElement:
+    short_name: str
+    long_name: Optional[str]
+    description: Optional[str]
```
andlaus, line 21: the TypedDict and the dataclass exhibit essentially the same attributes. it would be great if that repetition  could somehow be avoided. (if this is impossible, so be it...)

odxtools/element.py
```diff
@@ -0,0 +1,56 @@
...
+
+    @staticmethod
+    def get_kwargs(
```
andlaus, line 24: The name is a bit too generic for my taste. Also, it should IMO not be a public method. What do you think about
```suggestion
    def __attribdict_from_et__(
```

odxtools/element.py
```diff
@@ -0,0 +1,56 @@
...
+from typing import List, Optional, TypedDict, cast
+from xml.etree import ElementTree
+
+from odxtools.exceptions import odxrequire
+from odxtools.utils import create_description_from_et
+
+from .odxlink import OdxDocFragment, OdxLinkId
+
+
+class BaseElementKwargs(TypedDict):
```
andlaus, line 11: make this a non-public class?
```suggestion
class _NamedElementAttribDict(TypedDict):
```

andlaus: (maybe it could also be moved into the `NamedElement` class, a la `NamedElement._AttribDict`)

odxtools/element.py
andlaus: okay, it can be done using only dataclasses and mypy also seems to get it correctly:
```python
from dataclasses import dataclass, asdict

@dataclass
class Foo:
    hello: int

@dataclass
class Bar(Foo):
    world: float

print(Bar(**asdict(Foo(1)), world=2.0))
# Bar(hello=1, world=2.0)
```

(I guess it will not work if the base class is abstract and cannot be instantiated, though.)

kayoub5: are you sure mypy is not just ignoring it?

andlaus: it works:
```python
from dataclasses import dataclass, asdict

@dataclass
class Foo:
    hello: int

@dataclass
class Bar(Foo):
    world: float

print(Bar(**asdict(Foo(1)), world=2.0)) # ok
print(Bar(**asdict(Foo("foobar")), world=2.0)) # wrong type for "hello"
print(Bar(world=2.0)) # missing "hello"
```

produces
```
> mypy bla.py 
bla.py:12: error: Argument 1 to "Foo" has incompatible type "str"; expected "int"
bla.py:13: error: Missing positional argument "hello" in call to "Bar"
Found 2 errors in 1 file (checked 1 source file)
```

(sometimes I think that `mypy` is close to passing the turing test. usually that feeling goes away a minute later when it complains about something stupid ;))

odxtools/element.py
kayoub5: The class was removed

odxtools/element.py
kayoub5: asdict is recursive, it will convert sub properties to dict as well

odxtools/element.py
andlaus: hm, okay. so far I assumed that `__foo__` was meant not to be called outside if the class or its descendants. (basically because of `__init__()` and `__post_init__()`). If we were not to replace that pattern anyway, I guess that we should rename `__init_from_et__()` to `__init_from_et()`?

kayoub5: `__init_from_et` or `_init_from_et` depending if other files may call it or not

odxtools/element.py
```diff
@@ -0,0 +1,44 @@
...
+    @staticmethod
+    def _from_et(
+        et_element: ElementTree.Element,
+        doc_frags: List[OdxDocFragment],
+    ) -> "NamedElement":
+
+        return NamedElement(
+            short_name=odxrequire(et_element.findtext("SHORT-NAME")),
+            long_name=et_element.findtext("LONG-NAME"),
+            description=create_description_from_et(et_element.find("DESC")),
```
andlaus, line 26: I think that this is now the only user of `create_description_from_et()`? If so, the function should be made a private static method of the class?

odxtools/element.py
andlaus: maybe replacing `asdict()` with

```python
def extract_attribs(obj: Any) -> Dict[str, Any]:
    return { x.name: getattr(obj, x.name) for x in obj.__dataclass_fields__.values() }
```

it still seems to work with mypy...

andlaus:
@kayoub5: if you want, I can push the fixes branch into `chore/simplify_common_props` for you (the fixes branch should make the CI pass); I just wanted to avoid the confusion that usually emerges once two people push to a branch where this is not expected. (I'm still scratching my head why I can't open a PR for your PR. Possibly it's because `chore/simplify_common_props` is on the  main odxtools repository instead on your personal fork or maybe there is some SAML screwup...)

odxtools/element.py
kayoub5: There is one extra user for it

odxtools/element.py
andlaus: agreed. Anyway, that's stuff for another day...

odxtools/element.py
andlaus: okay, let's leave it like it is for now...

PR #192, opened 2023-08-19
Support BIT-MASK

kayoub5:


odxtools/compumethods/texttablecompumethod.py
```diff
@@ -39,8 +39,10 @@ def __is_internal_in_scale(self, internal_value, scale: CompuScale):
         if scale.lower_limit is not None and not scale.lower_limit.complies_to_lower(
                 internal_value):
             return False
-        if scale.upper_limit is not None and not scale.upper_limit.complies_to_upper(
-                internal_value):
+        # If no UPPER-LIMIT is defined
+        # the COMPU-SCALE will be applied only for the value defined in LOWER-LIMIT
+        upper_limit = scale.upper_limit or scale.lower_limit
+        if not upper_limit.complies_to_upper(internal_value):
```
kayoub5, line 45: @andlaus what I understood from the spec is that if no upper limit is defined for TEXTTABLE, SCALE-LINEAR, and SCALE-RAT-FUNC, then the internal value must exactly match the lower limit

odxtools/compumethods/compuscale.py
```diff
@@ -25,7 +25,7 @@ class CompuScale:
         description="<p>fancy description</p>", # optional: provide a description
         lower_limit=Limit(0), # required: lower limit
         upper_limit=Limit(3), # required: upper limit
-        compu_inverse_value=2, # required iff lower_limit != upper_limit
+        compu_inverse_value=2, # required if lower_limit != upper_limit
```
andlaus, line 28: small note: in math, the term "iff" means "if and only if", i.e., a bidirectional implication. let's use "if" here because it does not make much of a difference, and I'm not sure whether `compu_inverse_value` is mandatory for intervals containing more than one element...

odxtools/compumethods/texttablecompumethod.py
```diff
@@ -39,8 +39,10 @@ def __is_internal_in_scale(self, internal_value, scale: CompuScale):
         if scale.lower_limit is not None and not scale.lower_limit.complies_to_lower(
                 internal_value):
             return False
-        if scale.upper_limit is not None and not scale.upper_limit.complies_to_upper(
-                internal_value):
+        # If no UPPER-LIMIT is defined
+        # the COMPU-SCALE will be applied only for the value defined in LOWER-LIMIT
+        upper_limit = scale.upper_limit or scale.lower_limit
```
andlaus, line 44: what happens if `upper_limit` is `0`?
```suggestion
        upper_limit = scale.upper_limit
        if upper_limit is None:
            upper_limit = scale.lower_limit
```

odxtools/createanydiagcodedtype.py
```diff
@@ -65,7 +65,15 @@ def create_any_diag_coded_type_from_et(et_element: ElementTree.Element,
         bit_length = int(odxrequire(et_element.findtext("BIT-LENGTH")))
         bit_mask = None
         if (bit_mask_str := et_element.findtext("BIT-MASK")) is not None:
-            bit_mask = int(bit_mask_str, base=16)
+            # The XSD uses the type xsd:hexBinary
+            # The type xsd:hexBinary allow for leading/trailing whitespace
+            # The type xsd:hexBinary allow an empty string as value
+            # The type xsd:hexBinary only allows an even number of hex digits
+            # The ODX examples show an odd number of hex digits
```
andlaus, line 72:
```suggestion
            # xsd:hexBinary allows for leading/trailing whitespace, empty strings, and it only allows an even
            # number of hex digits, while some of the examples shown in the  ODX specification exhibit an
            # odd number of hex digits.
```

odxtools/standardlengthtype.py
```diff
@@ -18,24 +20,47 @@ class StandardLengthType(DiagCodedType):
...
+
+    def __apply_mask(self, internal_value: AtomicOdxType) -> AtomicOdxType:
+        if self.bit_mask is None:
+            return internal_value
+        if self.is_condensed_raw is True:
+            raise NotImplementedError("Serialization of condensed bit mask is not supported")
+        if isinstance(internal_value, int):
+            return internal_value & self.bit_mask
+        if isinstance(internal_value, bytes):
+            int_value: int = int.from_bytes(internal_value, 'big')
```
andlaus, line 39: the explicit type specification is not necessary here: mypy deduces the type of `int_value` correctly, and by convention, `.from_*()` constructors return an object of the type it was called for. That said, if you prefer to leave it in, that's perfectly fine with me...

tests/test_encoding.py
```diff
@@ -317,17 +333,57 @@ def test_issue_70(self):
             CodedConstParameter(
                 short_name="p6", diag_coded_type=uint1, bit_position=7, **param_kwargs),
         ]
-        req = Request(
-            odx_id=OdxLinkRef("request_id", doc_frags),
-            short_name="request_sn",
-            parameters=params,
-            long_name=None,
+        req = self._create_request(params)
+        self.assertFalse(req._BasicStructure__message_format_lines())
```
andlaus, line 337: drive-by:

```suggestion
        self.assertEqual(req._BasicStructure__message_format_lines(), [])
```

andlaus (commented):
Besides the usual nitpicks, this looks good. (Note that, since I do not have files which use masking, I cannot test this in the real world...)

odxtools/compumethods/texttablecompumethod.py
kayoub5: scale.lower_limit have type Limit, not int, so even if `scale.lower_limit.value = 0`, `scale.lower_limit` will still be truthy

odxtools/standardlengthtype.py
kayoub5: vscode was acting a bit weird, started behaving correctly when I added it

```suggestion
            int_value = int.from_bytes(internal_value, 'big')
```

andlaus:
all good. merging, thanks!

PR #193, opened 2023-08-20
Support COMPU-DEFAULT-VALUE

kayoub5:


odxtools/compumethods/texttablecompumethod.py
```diff
@@ -25,17 +28,27 @@ def __post_init__(self) -> None:
...
         if scale is not None:
             res = (
                 scale.compu_inverse_value
                 if scale.compu_inverse_value is not None else scale.lower_limit.value)
             odxassert(self.internal_type.isinstance(res))
             return res
+        raise EncodeError(f"Texttable compu method could not encode '{physical_value}'.")
 
     def __is_internal_in_scale(self, internal_value, scale: CompuScale):
+        if scale == self.compu_default_value:
```
andlaus, line 50: is it possible that the default value is not one of the specified scales? (that would be weird, but it would not be the first time...)

andlaus (commented):
after renaming `DataType.from_et_value_union()` to `DataType.from_et()` (and optionally adding type checking overloads), this looks good to me.

odxtools/compumethods/texttablecompumethod.py
kayoub5: Yes, it is possible, usually compu_default_value is used to auto assign "invalid", "unknown" to invalid internal values

odxtools/odxtypes.py
```diff
@@ -129,6 +130,16 @@ def as_python_type(self) -> type:
     def from_string(self, value: str) -> AtomicOdxType:
         return _PARSE_ODX_TYPE[self.value](value)
 
+    def from_et_value_union(self,
```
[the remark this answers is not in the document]
kayoub5: If you don't like from_et_value_union, the best next thing would be `parse_et`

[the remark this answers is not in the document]
andlaus: let's go for `parse_et()` or `create_from_et()` then. I still think that `from_et()` would be somewhat better, but I see your point...

odxtools/compumethods/texttablecompumethod.py
andlaus: okay, I assumed that in this case "invalid" was simply one of the entries of the table. never stop learning, I suppose ;)

andlaus:
cool! merging...

PR #194, opened 2023-08-20
Fix table lookup in TableKeyParameter

kayoub5:


odxtools/parameters/tablekeyparameter.py
```diff
@@ -71,7 +72,11 @@ def _resolve_snrefs(self, diag_layer: "DiagLayer") -> None:
 
     @property
     def table(self) -> "Table":
-        return self._table
+        if self._table is not None:
+            return self._table
+        if self._table_row is not None:
+            return self._table_row.table
+        odxraise(f'Could not resolve the table of {self.short_name}')
```
andlaus, line 79: I think it would be better to specify the return type of this property as `Optional["Table"]` and keep it always returning `self._table`...

odxtools/parameters/tablekeyparameter.py
```diff
@@ -81,7 +86,7 @@ def is_required(self):
         return self._table_row is None
 
     def is_optional(self):
-        return not self.is_required
+        return not self.is_required()
```
andlaus, line 89: thanks for the catch!

odxtools/parameters/tablekeyparameter.py
kayoub5: Tried that, got 13 mypy errors,

`self.table` is never `None`, but it may be resolved from `_table_row` instead of being referenced directly.

andlaus: In this case, I think that `self._table` should be declared to be `self.table: Optional["Table"]` and the static-case resolution should go to `._resolve_snrefs()`...

kayoub5: How would the declaration in `__post_init__` look like? I can't figure out a mypy happy declaration for it

andlaus: the same as it does now, but without the assignment:
```python
    def __post_init__(self) -> None:
        self._table: Optional["Table"]
        # [...]
```

odxtools/parameters/tablekeyparameter.py
```diff
@@ -24,6 +24,8 @@ class TableKeyParameter(Parameter):
     table_row_ref: Optional[OdxLinkRef]
 
     def __post_init__(self) -> None:
+        self._table: Optional[Table] = None
+        self._table_row: Optional[TableRow] = None
```
andlaus, line 28: I'm bit surprised that this works during normal program execution: the `Table` and `TableRow` classes should only be imported while type checking?
```suggestion
        self._table: "Table"
        self._table_row: Optional["TableRow"] = None
```

kayoub5: Since Python 3.9, type evaluation is lazy

andlaus: ah, good to know. That makes me wonder why the CI did not complain (because we should also run all tests on python 3.8...)

andlaus:
neat and clean. thanks for these improvements. merging, I'll also make a release...

PR #195, opened 2023-08-21
Fix TableKeyParameter again

kayoub5:


odxtools/parameters/tablekeyparameter.py
```diff
@@ -24,7 +24,7 @@ class TableKeyParameter(Parameter):
     table_row_ref: Optional[OdxLinkRef]
 
     def __post_init__(self) -> None:
-        self._table: "Table"
+        self._table: "Optional[Table]" = None
```
kayoub5, line 27: @andlaus we need to assign a value, otherwise table getter will raise `AttributeError: 'TableKeyParameter' object has no attribute '_table'`

andlaus: yes, but the  assignment should be done in either `_build_odxlinks()` or `_build_snrefs()`. What's missing is the static case (sorry that I missed that before merging the last PR):

```python
        # current code
        if self._table_row is not None:
            self._table = self._table_row.table
```

kayoub5: That logic is not correct, `self._table = self._table_row.table` must only be done if table was not resolved, otherwise `self._table_row.table` may return incorrect table, in case of the row being ref imported from one table to another.

andlaus: i believe it is correct, because specifying both, the table row and the table is illegal. (if the table row is specified, this is the static case.) the XSD agrees:
image
(I think we should not cater too much to invalid input files.)

andlaus: (in the first two cases, the table row is specified via SNREF and is optional. if a row is specified there, we just assign `.table` with the same object again...)

kayoub5: Double check the definition of `ODXLINK-TO-TABLE` and `SNREF-TO-TABLE`, unlike their name specify, they include optional row ref.

This means that user can:
* Specify both `TABLE-REF` and `TABLE-ROW-SNREF` (See `ODXLINK-TO-TABLE`), in this case, each link must be used independently (the row may defined in a different table and imported trough `TABLE > TABLE-ROW-REF`) 
* Specify only `TABLE-REF`  (See `ODXLINK-TO-TABLE`)
* Specify both `TABLE-SNREF` and `TABLE-ROW-SNREF` (See `SNREF-TO-TABLE`), in this case, each link must be used independently (the row may defined in a different table and imported trough `TABLE > TABLE-ROW-REF`) 
* Specify only `TABLE-SNREF` (See `SNREF-TO-TABLE`)
* Specify only `TABLE-ROW-REF`, This is where we have a problem now, where we need to read the table from the table row, only and only in this case must we ready the table from the row.

For reference
```xml
    <xsd:group name="ODXLINK-TO-TABLE">
        <xsd:sequence>
            <xsd:element maxOccurs="1" minOccurs="1" name="TABLE-REF" type="ODXLINK"/>
            <xsd:element maxOccurs="1" minOccurs="0" name="TABLE-ROW-SNREF" type="SNREF"/>
        </xsd:sequence>
    </xsd:group>
```

andlaus: not really: exactly one of "ODXLINK-TO-TABLE", "SNREF-TO-TABLE" or "TABLE-ROW-REF" has to be  specify, where the first two options may include an SNREF to a row, but if this is specified, the table where this row is located is -- by definition -- the table which contains the row. in this case, we would override `self._table` with the same object that it already points to if we go down the route proposed above (be aware that python assigns "non-primitive" objects by reference)...

(IMO the specification is **way** too twisted here.)

kayoub5: Logic updated

andlaus: your solution is clearer than what I proposed. cool!

andlaus:
okay, next try. let's merge.

PR #199, opened 2023-09-05
remove the `ProgCodeSyntax` enum class

andlaus:
Instead, make this a string because the SYNTAX subtag of SINGLE-ECU-JOB is specified to be a plain string in the ODX specification. It seems like we wanted to be more strict on our own files and tried to constrain the possible values for this. This was implemented using a string literal (which is like a type-checking only enum, i.e., it does not check if the value specified is valid if can only be determined at runtime). After changing this to an `Enum` class in #172, it got some (incorrect) teeth which bit users that used non-Java single ECU jobs.

### DIFF
--- a/odxtools/compumethods/texttablecompumethod.py
+++ b/odxtools/compumethods/texttablecompumethod.py
@@ -36,7 +36,7 @@ class TexttableCompuMethod(CompuMethod):
         return scales
 
     def convert_physical_to_internal(self, physical_value):
-        scale: CompuScale = next(
+        scale = next(
             filter(lambda scale: scale.compu_const == physical_value, self._get_scales()), None)
         if scale is not None:
             res = (
--- a/odxtools/compumethods/texttablecompumethod.py
+++ b/odxtools/compumethods/texttablecompumethod.py
@@ -36,7 +36,7 @@ class TexttableCompuMethod(CompuMethod):
         return scales
 
     def convert_physical_to_internal(self, physical_value):
-        scale = next(
+        scale: CompuScale = next(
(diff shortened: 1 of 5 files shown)


odxtools/compumethods/texttablecompumethod.py
```diff
@@ -36,7 +36,7 @@ def _get_scales(self):
         return scales
 
     def convert_physical_to_internal(self, physical_value):
-        scale: CompuScale = next(
+        scale = next(
```
kayoub5, line 39: why?

andlaus: mypy started to warn about this:
```
```
odxtools/compumethods/texttablecompumethod.py:39: note: By default the bodies of untyped functions are not checked, consider using --check-untyped-defs  [annotation-unchecked]
Success: no issues found in 150 source files
```

The real fix would be to add type annotations to all `convert_physical_to_internal()` methods, but I was not in the mood to open that can of worms...

andlaus: (note: comprehensive type annotations for everything are currently relatively high on my list of priorities...)

kayoub5: please keep it, even if mypy is not checking it, that type hint helps the IDE to provide correct suggestions.

mypy is informing you, it's not a warning or an error.

andlaus: good point. (as a user of emacs without a language server, I do not notice such things...) fixed: 75967cd9cc88baccba3961d2cb8655f8d23fd15f.

kayoub5 (approved):

PR #201, opened 2023-09-07
Don't assert about short names starting with digits

kayoub5:
OEM files are using short names that start with digit

The generates hundreds of errors/warnings for something that is not against ODX standard

The standard allows for short names to start with digits, so making an assert for it don't make much sense

Users who need to access an item that is a keyword or start with digit will get a syntax error anyway and their code won't run, so the extra assert is pointless.


andlaus:
LGTM, thanks!

PR #202, opened 2023-09-08
rework .is_required and .is_optional methods of parameters

andlaus:
`Parameter.is_optional()` and `Parameter.is_required()` were quite confusing as it can be assumed that one is just the negation of the other. We thus rename `.is_optional` to `.is_settable` to more clearly express that this indicates that a parameter is user-definable. (In the process, a few instances in the `browse` tool where this seems to have been mixed up are fixed.) While at it, make these functions properties so that they do not need to be called anymore...

### DIFF
--- a/odxtools/parameters/parameter.py
+++ b/odxtools/parameters/parameter.py
@@ -62,13 +62,28 @@ class Parameter(NamedElement, abc.ABC):
     def bit_length(self) -> Optional[int]:
         return None
 
-    @abc.abstractmethod
-    def is_required(self):
-        pass
+    @property
+    def is_required(self) -> bool:
+        """True if the parameter must be explicitly specified when
+        encoding a message.
 
-    @abc.abstractmethod
-    def is_optional(self):
-        pass
+        Required parameters are always settable, and parameters which
+        have a default value are settable but not required to be
+        specified.
+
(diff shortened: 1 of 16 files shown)


odxtools/parameters/parameter.py
```diff
@@ -62,13 +62,28 @@ def parameter_type(self) -> ParameterType:
...
+    @property
+    def is_settable(self) -> bool:
+        """True if the parameter can be specified when encoding a
+        message.
+
+        Required parameters are always settable, and parameters which
+        have a default value are settable but not required to be
+        specified.
+        """
+        raise NotImplemented
```
kayoub5, line 86: https://www.flake8rules.com/rules/F901.html

odxtools/parameters/parameter.py
```diff
@@ -62,13 +62,28 @@ def parameter_type(self) -> ParameterType:
...
 
-    @abc.abstractmethod
-    def is_optional(self):
-        pass
+        Required parameters are always settable, and parameters which
+        have a default value are settable but not required to be
+        specified.
+
+        """
+        raise NotImplemented
```
kayoub5, line 75: https://www.flake8rules.com/rules/F901.html

PR #203, opened 2023-09-11
Add type annotations, part 1

andlaus:
This is the kick-off PR of another long and winding quest; this time the goal is to add comprehensive type annotations to everything so that `mypy --disallow-untyped-defs` passes for odxtools without any complaints. Note that this quest will include many side-tasks; I indent to put them into separate PRs as much as possible, but it is about as certain as death and taxes that I'll occasionally miss a few of these "drive-by fixes"...

### DIFF
--- a/odxtools/cli/dummy_sub_parser.py
+++ b/odxtools/cli/dummy_sub_parser.py
@@ -1,7 +1,6 @@
 # SPDX-License-Identifier: MIT
 import argparse
 import sys
-from typing import List
 
 
 class DummyTool:
--- a/odxtools/cli/dummy_sub_parser.py
+++ b/odxtools/cli/dummy_sub_parser.py
@@ -1,6 +1,7 @@
 # SPDX-License-Identifier: MIT
 import argparse
 import sys
+from typing import List
 
 
 class DummyTool:
@@ -14,11 +15,11 @@ class DummyTool:
     should bail out.
     """
 
-    def __init__(self, tool_name, error):
+    def __init__(self, tool_name: str, error: Exception):
         self._odxtools_tool_name_ = tool_name
         self._error = error
 
-    def add_subparser(self, subparser_list):
+    def add_subparser(self, subparser_list: argparse._SubParsersAction) -> None:
         subparser_list.add_parser(
             self._odxtools_tool_name_,
             description=f"Tool '{self._odxtools_tool_name_}' is unavailable: {self._error}",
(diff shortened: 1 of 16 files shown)


odxtools/cli/dummy_sub_parser.py
```diff
@@ -14,19 +15,19 @@ class DummyTool:
     should bail out.
     """
 
-    def __init__(self, tool_name, error):
+    def __init__(self, tool_name: str, error: Exception):
         self._odxtools_tool_name_ = tool_name
         self._error = error
 
-    def add_subparser(self, subparser_list):
+    def add_subparser(self, subparser_list: argparse._SubParsersAction) -> None:
```
kayoub5, line 22: `_SubParsersAction` is private

odxtools/isotp_state_machine.py
```diff
@@ -28,29 +29,29 @@ class IsoTpStateMachine:
...
 
-        E.g., add some data to a telegram, etc.
+        E.g., add some data to a telegram, etc. Returns a generator of
+        (receive_id, payload_data) tuples.
         """
         try:
             telegram_idx = self._can_rx_ids.index(rx_id)
         except ValueError:
-            return False  # unknown CAN ID
+            return  # unknown CAN ID
```
kayoub5, line 52: behavior change?

andlaus: not really: this method was always called with `do_yield = True`, and I'm honestly not sure anymore why I've implemented that option in the first place...

odxtools/cli/dummy_sub_parser.py
andlaus: yes, but how to annotate this correctly? `Any`?

kayoub5: ## Option 1
https://github.com/python/typeshed/issues/7539#issuecomment-1076581049

## Option 2

Even if python makes changes and `_SubParsersAction` is no longer there, only mypy will report it, but the code itself will still work.

```suggestion
    def add_subparser(self, subparser_list: "argparse._SubParsersAction") -> None:
```

andlaus: cool. I picked option two: e8009c70294dc4c2b10c95a9011da0e3e6fe9d48

odxtools/cli/dummy_sub_parser.py
```diff
@@ -1,6 +1,7 @@
 # SPDX-License-Identifier: MIT
 import argparse
 import sys
+from typing import List
```
kayoub5, line 4: added but not used

kayoub5: other than this, LGTM

andlaus: okay, I've removed all unused imports (at least those flagged by `ruff`): 58c013ac695db3e3687d87e83e2228d8769637cf

PR #205, opened 2023-09-13
Support DynamicLengthField: Part 1

kayoub5:


odxtools/parameters/parameterwithdop.py
```diff
@@ -46,7 +46,12 @@ def _resolve_snrefs(self, diag_layer: "DiagLayer") -> None:
         if self.dop_snref:
             spec = diag_layer.diag_data_dictionary_spec
             self._dop = (
-                spec.data_object_props.get(self.dop_snref) or spec.structures.get(self.dop_snref))
+                spec.end_of_pdu_fields.get(self.dop_snref) or
+                spec.dynamic_length_fields.get(self.dop_snref) or
+                spec.data_object_props.get(self.dop_snref) or spec.structures.get(self.dop_snref) or
+                spec.muxs.get(self.dop_snref) or spec.dtc_dops.get(self.dop_snref) or
+                spec.muxs.get(self.dop_snref) or spec.env_data_descs.get(self.dop_snref) or
+                spec.env_datas.get(self.dop_snref))
```
andlaus, line 54: this is fun. we probably should provide an `.all_dops` property...

odxtools/positioneddataobjectproperty.py
```diff
@@ -0,0 +1,74 @@
+# SPDX-License-Identifier: MIT
```
andlaus, line 1: as said above, I'd prefer to cling to the XSD and duplicate this class into `SwitchKey` and `DetermineNumberOfItems`

andlaus (commented):
Besides my usual nits, I like this. (I have a branch for implementing additional field objects, bit the one of this PR is much more comprehensive.)

examples/somersaultecu.py
```diff
@@ -1200,7 +1200,7 @@ class SomersaultSID(IntEnum):
             is_visible_raw=None,
             sdgs=[],
             byte_position=0,
-            switch_key=MultiplexerSwitchKey(
+            switch_key=PositionedDataObjectProperty(
```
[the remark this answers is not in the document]
kayoub5: As a type alias in mutex file is acceptable?

odxtools/endofpdufield.py
```diff
@@ -67,32 +34,13 @@ def from_et(et_element: ElementTree.Element,
...
-            structure_ref=structure_ref,
-            structure_snref=structure_snref,
-            env_data_desc_ref=env_data_desc_ref,
-            env_data_desc_snref=env_data_desc_snref,
             min_number_of_items=min_number_of_items,
             max_number_of_items=max_number_of_items,
-            is_visible_raw=is_visible_raw,
             **kwargs)
 
         return eopf
```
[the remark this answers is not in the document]
kayoub5: that's existing code...

odxtools/parameters/parameterwithdop.py
kayoub5: I would have done so, if there was another class with same logic, but I couldn't find any

odxtools/endofpdufield.py
```diff
@@ -67,32 +33,13 @@ def from_et(et_element: ElementTree.Element,
...
 
-    @property
-    def structure(self) -> "BasicStructure":
-        """may be a Structure or a env-data-desc"""
-        return self._structure
```
kayoub5, line 87: @andlaus it seems env-data-desc was broken in existing code of endofpdufield

odxtools/endofpdufield.py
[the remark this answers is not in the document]
andlaus: yeah, I prefer to fix such tiny things if the code has to be touched anyway. you can leave it as-is if you are a follower of a strict "no drive-by fixes"-policy, though...

odxtools/parameters/parameterwithdop.py
andlaus: yeah, there currently isn't any. stuff for another PR...

odxtools/field.py
```diff
@@ -0,0 +1,90 @@
...
+    def bit_length(self):
+        return self.structure.bit_length
+
+    def convert_physical_to_internal(self, physical_value):
+        return self.structure.convert_physical_to_internal(physical_value)
+
+    def _resolve_odxlinks(self, odxlinks: OdxLinkDatabase) -> None:
+        """Recursively resolve any odxlinks references"""
+        if self.structure_ref is not None:
+            self._structure = odxlinks.resolve(self.structure_ref)
```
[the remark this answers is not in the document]
kayoub5: @andlaus existing code is broken, I just moved it over to new file

odxtools/endofpdufield.py
andlaus: true: We don't test it in the unittests and I also do not have any files which use ENV-DATA-DESC in END-OF-PDUFIELDs. Let's try to do it right this time (at least in principle); Without exercising this code, I have some doubts if all details will be correct, but let's burn that bridge when we come to it...

odxtools/endofpdufield.py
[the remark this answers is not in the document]
kayoub5: I avoid drive-by fixes

odxtools/endofpdufield.py
kayoub5: I updated the code, it still does not support ENV-DATA-DESC, but it least it reports it correctly once detected

andlaus (approved):

PR #208, opened 2023-09-21
Drive by fixes #1

andlaus:
This PR consists of fixes for issues I've noticed while preparing [#203](https://github.com/mercedes-benz/odxtools/pull/203). It turned out to be quite a bit larger than the motivating PR...

(note: all patches of this PR are fully self-contained and the PR is also bisectable. Since all commits should be basically independent of each other, it is probably easier to review the PR by looking at the patches individually instead of the cumulative diff.)

### DIFF
--- a/odxtools/basicstructure.py
+++ b/odxtools/basicstructure.py
@@ -50,8 +50,9 @@ class BasicStructure(DopBase):
                     byte_position_int = (
                         param.byte_position if param.byte_position is not None else 0)
                     offset = byte_position_int * 8 + bit_position_int
-                bit_length = 0 if param.bit_length is None else param.bit_length
-                offset += bit_length
+                if param.bit_length is None:
+                    return None
+                offset += param.bit_length
 
                 length = max(length, offset)
 
--- a/odxtools/basicstructure.py
+++ b/odxtools/basicstructure.py
@@ -8,7 +8,7 @@ from .dataobjectproperty import DataObjectProperty
 from .decodestate import DecodeState
 from .dopbase import DopBase
 from .encodestate import EncodeState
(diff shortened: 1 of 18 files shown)


odxtools/basicstructure.py
```diff
@@ -22,12 +22,11 @@
 
 if TYPE_CHECKING:
     from .diaglayer import DiagLayer
-    from .endofpdufield import EndOfPduField
 
 
 @dataclass
 class BasicStructure(DopBase):
-    parameters: NamedItemList[Union[Parameter, "EndOfPduField"]]
+    parameters: NamedItemList[Parameter]
```
kayoub5, line 29: Are you sure about this?

andlaus: yes: all fields are DOPs and thus encoded using `ValueParameter` and the code actually already works like this. I don't know how this sneaked, but I if I had to speculate, I would bet that it is an artifact from the time when EndOfPduFields were implemented...

odxtools/basicstructure.py
```diff
@@ -50,7 +50,8 @@ def bit_length(self) -> Optional[int]:
                     byte_position_int = (
                         param.byte_position if param.byte_position is not None else 0)
                     offset = byte_position_int * 8 + bit_position_int
-                offset += param.bit_length
+                bit_length = 0 if param.bit_length is None else param.bit_length
```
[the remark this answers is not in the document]
andlaus: right, my bad. I find this tangle of en- and decode functions utterly confusing. fixed: e5f4d98f8a4a8d03e55492e976fc12a24244b1b2

kayoub5 (approved):

PR #210, opened 2023-09-25
Refactor `MinMaxLengthType` coding functionality

andlaus:
This PR is one of the first baby steps towards untangling the de-/encoding functionality. It renames `DecodeState.next_byte_position` to `DecodeState.cursor_position`, and cleans up the de- and encoding code of the `MinMaxLengthType` DOP.

### DIFF
--- a/odxtools/minmaxlengthtype.py
+++ b/odxtools/minmaxlengthtype.py
@@ -34,9 +34,9 @@ class MinMaxLengthType(DiagCodedType):
     def dct_type(self) -> DctType:
         return "MIN-MAX-LENGTH-TYPE"
 
-    def __termination_character(self) -> bytes:
-        """Returns the termination character or None if it isn't defined."""
-        # The termination character is actually not specified by ASAM
+    def __termination_sequence(self) -> bytes:
+        """Returns the termination byte sequence if it isn't defined."""
+        # The termination sequence is actually not specified by ASAM
         # for A_BYTEFIELD but I assume it is only one byte.
         termination_char = b''
         if self.termination == "ZERO":
@@ -70,16 +70,16 @@ class MinMaxLengthType(DiagCodedType):
         if encode_state.is_end_of_pdu or len(value_bytes) == self.max_length:
             # All termination types may be ended by the end of the PDU
             # or once reaching the maximum length. In this case, we
-            # must not add the termination character
+            # must not add the termination sequence
             pass
         else:
-            termination_char = self.__termination_character()
-            if termination_char is not None:
-                # ensure that we don't try to encode an odd-length
-                # value whn using a two-character terminator
-                odxassert(len(value_bytes) % len(termination_char) == 0)
+            termination_char = self.__termination_sequence()
 
-                value_bytes.extend(termination_char)
+            # ensure that we don't try to encode an odd-length
+            # value when using a two-byte terminator
+            odxassert(len(value_bytes) % len(termination_char) == 0)
+
+            value_bytes.extend(termination_char)
 
         if len(value_bytes) < self.min_length:
             raise EncodeError(f"Encoded value for MinMaxLengthType "
@@ -98,7 +98,7 @@ class MinMaxLengthType(DiagCodedType):
 
         coded_message = decode_state.coded_message
         cursor_pos = decode_state.cursor_position
-        terminator_char = self.__termination_character()
+        terminator_char = self.__termination_sequence()
 
         # If no termination char is found, this is the next byte after the parameter.
         max_terminator_pos = len(coded_message)
@@ -107,16 +107,16 @@ class MinMaxLengthType(DiagCodedType):
 
         if self.termination != "END-OF-PDU":
             # The parameter either ends after the maximum length, at
-            # the end of the PDU or if a termination character is
+            # the end of the PDU or if a termination sequence is
             # found.
 
             terminator_pos = cursor_pos + self.min_length
             while True:
-                # Search the termination character
+                # Search the termination sequence
                 terminator_pos = coded_message.find(terminator_char, terminator_pos,
                                                     max_terminator_pos)
                 if terminator_pos < 0:
-                    # termination character was not found, i.e., we
+                    # termination sequence was not found, i.e., we
                     # are terminated by either the end of the PDU or
                     # our maximum size. (whatever is the smaller
                     # value.)
@@ -127,14 +127,14 @@ class MinMaxLengthType(DiagCodedType):
                     byte_length = terminator_pos - cursor_pos
                     break
                 elif (terminator_pos - cursor_pos) % len(terminator_char) == 0:
-                    # we found the termination character at a position
+                    # we found the termination sequence at a position
                     # and it is correctly aligned (two-byte
-                    # termination "characters" must be word aligned
+                    # termination sequences must be word aligned
                     # relative to the beginning of the parameter)!
                     byte_length = terminator_pos - cursor_pos
                     break
                 else:
-                    # we found the termination character, but its
+                    # we found the termination sequence, but its
                     # alignment was incorrect. Try again one byte
                     # further...
                     terminator_pos += 1
@@ -159,7 +159,7 @@ class MinMaxLengthType(DiagCodedType):
             if byte_pos != len(coded_message) and byte_pos - cursor_pos != self.max_length:
                 byte_pos += len(terminator_char)
 
-            # next byte starts after the actual data and the termination character
+            # next byte starts after the actual data and the termination sequence
             return value, byte_pos
         else:
             # If termination == "END-OF-PDU", the parameter ends after max_length
--- a/odxtools/minmaxlengthtype.py
+++ b/odxtools/minmaxlengthtype.py
@@ -34,11 +34,11 @@ class MinMaxLengthType(DiagCodedType):
     def dct_type(self) -> DctType:
         return "MIN-MAX-LENGTH-TYPE"
 
-    def __termination_character(self):
+    def __termination_character(self) -> bytes:
         """Returns the termination character or None if it isn't defined."""
         # The termination character is actually not specified by ASAM
         # for A_BYTEFIELD but I assume it is only one byte.
-        termination_char = None
+        termination_char = b''
         if self.termination == "ZERO":
             if self.base_data_type not in [DataType.A_UNICODE2STRING]:
                 termination_char = bytes([0x0])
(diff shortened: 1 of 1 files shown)


odxtools/minmaxlengthtype.py
```diff
@@ -34,11 +34,11 @@ def __post_init__(self):
     def dct_type(self) -> DctType:
         return "MIN-MAX-LENGTH-TYPE"
 
-    def __termination_character(self):
+    def __termination_character(self) -> bytes:
         """Returns the termination character or None if it isn't defined."""
```
kayoub5, line 38: docs don't match implementation

odxtools/minmaxlengthtype.py
```diff
@@ -53,94 +53,125 @@ def __termination_character(self):
...
-            is_highlow_byte_order=self.is_highlow_byte_order,
-        )
-
-        if encode_state.is_end_of_pdu or byte_length == self.max_length:
-            # All termination types may be ended by the PDU
-            return value_byte
+        bit_length = 8 * self.max_length if self.max_length is not None else 8 * len(internal_value)
+        bit_length = min(8 * len(internal_value), bit_length)
+        value_bytes = bytearray(
+            self._to_bytes(
```
kayoub5, line 59: internal value can be a variable size int, not just bytes and str

odxtools/minmaxlengthtype.py
```diff
@@ -53,94 +53,125 @@ def __termination_character(self):
...
             termination_char = self.__termination_character()
-            if self.termination == "END-OF-PDU":
-                termination_char = b''
-            odxassert(
-                termination_char is not None,
-                f"MinMaxLengthType with termination {self.termination}"
-                f"(min: {self.min_length}, max: {self.max_length}) failed encoding {internal_value}"
-            )
-            return value_byte + termination_char
+            if termination_char is not None:
```
kayoub5, line 77: you changed the code not to return None

odxtools/minmaxlengthtype.py
```diff
@@ -53,94 +53,125 @@ def __termination_character(self):
...
+                # ensure that we don't try to encode an odd-length
+                # value whn using a two-character terminator
```
kayoub5, line 79: you mean a two bytes terminator (it's still one character)

odxtools/minmaxlengthtype.py
```diff
@@ -53,94 +53,125 @@ def __termination_character(self):
...
+                odxassert(len(value_bytes) % len(termination_char) == 0)
+
+                value_bytes.extend(termination_char)
+
+        if len(value_bytes) < self.min_length:
+            raise EncodeError(f"Encoded value for MinMaxLengthType "
+                              f"must be at least {self.min_length} bytes long. "
```
kayoub5, line 86: bytes or characters?

kayoub5:
@andlaus please move the renaming to its own PR, The review of encode/decode logic is too complicated to throw a drive-by rename at it.

odxtools/minmaxlengthtype.py
kayoub5: docs says "Returns the termination character or None", I don't see a return None in the code.

andlaus: ah, you mean the `None`? right, fixed: c91ebaa7877d94fff2f10ae10900bc194813e418 (I also renamed this to `__termination_sequence`

odxtools/minmaxlengthtype.py
andlaus: the spec only allows standard length types here, and even these tend to come with severe restrictions (e.g. for A_FLOAT32 min_length == max_length == 32.) AFAICS, the only data type which has variable size are strings, cf. section 7.3.6.2 of the spec (p. 94 ff)

(Honestly, I have severe doubts about the usefulness of MinMaxLengthType in practice...)

andlaus:
> @andlaus please move the renaming to its own PR, The review of encode/decode logic is too complicated to throw a drive-by rename at it.

okay, done: #211. To avoid unnecessary work, I did not rebase the present PR, so the changes of #211 will continue to show up here until #211 is merged...

odxtools/minmaxlengthtype.py
kayoub5: Usually there is no termination sequence for variable size int, only valid value I can think of would be end-of-pdu.

andlaus: yeah, I suppose that being unable to specify integers that have their most-significant byte 0 is kind of a show stopper...

that said, I believe `._to_bytes()` works for arbitrary data types? (at least modulo bugs)

odxtools/minmaxlengthtype.py
```diff
@@ -34,113 +34,136 @@ def __post_init__(self):
...
+                termination_sequence = bytes([0x0, 0x0])
         elif self.termination == "HEX-FF":
             if self.base_data_type not in [DataType.A_UNICODE2STRING]:
-                termination_char = bytes([0xFF])
+                termination_sequence = bytes([0xFF])
             else:
-                termination_char = bytes([0xFF, 0xFF])
-        return termination_char
+                termination_sequence = bytes([0xFF, 0xFF])
+        return termination_sequence
```
kayoub5, line 52: Just a proposal to simplify the logic (existing logic is correct, and you may keep it as is)

```python
if self.termination == "ZERO":
    termination_sequence = bytes([0x0])
elif self.termination == "HEX-FF":
    termination_sequence = bytes([0xFF])
else:
    termination_sequence = b''
if self.base_data_type == DataType.A_UNICODE2STRING:
    termination_sequence *= 2
return termination_sequence
```

odxtools/minmaxlengthtype.py
```diff
@@ -34,113 +34,136 @@ def __post_init__(self):
...
-                termination_char is not None,
-                f"MinMaxLengthType with termination {self.termination}"
-                f"(min: {self.min_length}, max: {self.max_length}) failed encoding {internal_value}"
-            )
-            return value_byte + termination_char
+            termination_sequence = self.__termination_sequence()
+
+            # ensure that we don't try to encode an odd-length
+            # value when using a two-byte terminator
+            odxassert(len(value_bytes) % len(termination_sequence) == 0)
```
kayoub5, line 80: Are you sure about the `% len(termination_sequence)` thing?
```python
>>> 5 % 0
  File "<stdin>", line 1, in <module>
ZeroDivisionError: integer modulo by zero
```

odxtools/minmaxlengthtype.py
```diff
@@ -34,113 +34,136 @@ def __post_init__(self):
...
+
+            value_bytes.extend(termination_sequence)
+
+        if len(value_bytes) < self.min_length:
+            raise EncodeError(f"Encoded value for MinMaxLengthType "
+                              f"must be at least {self.min_length} bytes long. "
```
kayoub5, line 86: * Suggestion: use odxassert instead.
* min_length => is this in bytes or in characters?

odxtools/minmaxlengthtype.py
```diff
@@ -34,113 +34,136 @@ def __post_init__(self):
...
+        termination_seq = self.__termination_sequence()
 
-        # If no termination char is found, this is the next byte after the parameter.
-        max_termination_byte = len(coded_message)
+        max_terminator_pos = len(coded_message)
         if self.max_length is not None:
-            max_termination_byte = min(max_termination_byte, cursor_position + self.max_length)
+            max_terminator_pos = min(max_terminator_pos, cursor_pos + self.max_length)
 
         if self.termination != "END-OF-PDU":
```
kayoub5, line 107: What's the default value for `self.termination` if not specified in the XML?

odxtools/minmaxlengthtype.py
```diff
@@ -34,113 +34,136 @@ def __post_init__(self):
...
+                terminator_pos = coded_message.find(termination_seq, terminator_pos,
+                                                    max_terminator_pos)
+                if terminator_pos < 0:
+                    # termination sequence was not found, i.e., we
+                    # are terminated by either the end of the PDU or
+                    # our maximum size. (whatever is the smaller
+                    # value.)
+                    terminator_pos = len(coded_message)
+                    if self.max_length is not None:
+                        terminator_pos = min(cursor_pos + self.max_length, terminator_pos)
```
kayoub5, line 124: Why not just `terminator_pos = max_terminator_pos` ?

odxtools/minmaxlengthtype.py
andlaus: I like the current code slightly better, even though it is a bit longer: I don't like doing arithmetic on non-numeric variables. (I'm less opposed to doing this to literals a la `[0]*num_bytes`)

kayoub5: `termination_sequence = termination_sequence + termination_sequence`

odxtools/minmaxlengthtype.py
andlaus: good point: if there is an error in the ODX file and the termination of such a DOP is set to END-OF-PDU while the corresponding parameter is *not* at the end of the PDU, we get this division by zero. I've added an odxassert: fe6d83870ced635ea36c66337d2974b9ecc59b4a

odxtools/minmaxlengthtype.py
andlaus: > Suggestion: use odxassert instead.

I think that at the other occasions where we catch an illegal parameter value, we raise an EncodeError as well?

> min_length => is this in bytes or in characters?

the spec says bytes. (Note that I think that this is a bad idea for anything other than ASCII or UTF-8 strings. IMO, they should have restricted this kind of DOP to strings, removed the non-null termination options and called it something like "CStringDOP"...)

> No f"(Is: {len(value_bytes)} bytes.)" ?

I guess that could be useful to find a value which works (how to determine the the length of an encoded from a physical value is not obvious for me...)

odxtools/minmaxlengthtype.py
andlaus: the XSD says it is a required attribute...

odxtools/minmaxlengthtype.py
andlaus: good point.  fixed: f7edde6386ce41ef08d5a16caecad4a224c58860

kayoub5:
Current remaining open topics are optional, current code is good as is.

kayoub5 (approved):

odxtools/minmaxlengthtype.py
andlaus: better, but I still slightly prefer the jungle of `if`s. That said, if you have more than just a cursory preference, I'll change it...

kayoub5: This a proposal, you can ignore

andlaus: I only want to ignore it if you deem it to be unimportant :)

kayoub5: already did, PR approved

andlaus:
alright, let's get this done!

PR #212, opened 2023-09-27
Add type annotations, part 2

andlaus:
This is the second part of the quest for comprehensive type annotations. Besides the on-topic changes, it also contains a minor cleanup of the README, and the removal of the completely unused but confusingly named `Parameter.get_coded_value()` method.

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,6 +39,7 @@ send to/received from ECUs in an pythonic manner.
 - [Installation](#installation)
 - [Usage Examples](#usage-examples)
   - [Python snippets](#python-snippets)
+- [Using the non-strict mode](#using-the-non-strict-mode)
 - [Interactive Usage](#interactive-usage)
   - [Python REPL](#python-repl)
 - [Command line usage](#command-line-usage)
@@ -608,10 +609,6 @@ project, please read the [contributing guide](https://github.com/mercedes-benz/o
 
 Please read our [Code of Conduct](https://github.com/mercedes-benz/daimler-foss/blob/master/CODE_OF_CONDUCT.md)
 as it is our base for interaction.
-
-## License
-
-This project is licensed under the [MIT LICENSE](https://github.com/mercedes-benz/odxtools/blob/main/LICENSE).
 
 ## Provider Information
 
--- a/README.md
+++ b/README.md
@@ -1,4 +1,8 @@
 <!-- SPDX-License-Identifier: MIT -->
+[![PyPi - Version](https://img.shields.io/pypi/v/odxtools)](https://pypi.org/project/odxtools)
+[![PyPI - License](https://img.shields.io/pypi/l/odxtools)](LICENSE)
+[![CI Status](https://github.com/mercedes-benz/odxtools/actions/workflows/tests.yml/badge.svg?branch=main)](https://github.com/mercedes-benz/odxtools/actions?query=branch%3Amain)
+
 # odxtools
 
 `odxtools` is a set of utilities for working with diagnostic
--- a/odxtools/compumethods/texttablecompumethod.py
+++ b/odxtools/compumethods/texttablecompumethod.py
@@ -3,7 +3,7 @@ from dataclasses import dataclass
 from typing import List, Optional
 
 from ..exceptions import DecodeError, EncodeError, odxassert
-from ..odxtypes import DataType
+from ..odxtypes import AtomicOdxType, DataType
 from .compumethod import CompuMethod, CompuMethodCategory
 from .compuscale import CompuScale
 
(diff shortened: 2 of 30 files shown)


README.md
```diff
@@ -609,10 +610,6 @@ project, please read the [contributing guide](https://github.com/mercedes-benz/o
 Please read our [Code of Conduct](https://github.com/mercedes-benz/daimler-foss/blob/master/CODE_OF_CONDUCT.md)
 as it is our base for interaction.
 
-## License
```
kayoub5, line 612: Remove license from table of content

README.md
```diff
@@ -609,10 +610,6 @@ project, please read the [contributing guide](https://github.com/mercedes-benz/o
...
-
-This project is licensed under the [MIT LICENSE](https://github.com/mercedes-benz/odxtools/blob/main/LICENSE).
```
kayoub5, line 614: TIP: add a license badge

PyPI - License

```markdown
![PyPI - License](https://img.shields.io/pypi/l/odxtools)

```

odxtools/compumethods/texttablecompumethod.py
```diff
@@ -28,25 +28,30 @@ def __post_init__(self) -> None:
...
+    def convert_physical_to_internal(self, physical_value: AtomicOdxType) -> AtomicOdxType:
+        scales = [x for x in self._get_scales() if x.compu_const == physical_value]
+        if scales:
+            inverse_value: Optional[AtomicOdxType] = scales[0].compu_inverse_value
+            if inverse_value is not None:
+                res = inverse_value
+            elif scales[0].lower_limit is not None:
+                res = scales[0].lower_limit.value
+            else:
+                res = None
```
kayoub5, line 47: Maybe try the other scales? not just scales[0]

odxtools/compumethods/texttablecompumethod.py
```diff
@@ -28,25 +28,30 @@ def __post_init__(self) -> None:
...
+
             odxassert(self.internal_type.isinstance(res))
-            return res
-        raise EncodeError(f"Texttable compu method could not encode '{physical_value}'.")
+            return res  # type: ignore[return-value]
```
kayoub5, line 50: Why are you ignoring return value type?

odxtools/isotp_state_machine.py
```diff
@@ -112,12 +113,12 @@ async def read_telegrams(self, bus):
         :param bus: Input file or socket of can bus to read the can frames
         """
 
-        if hasattr(bus, "set_filters"):
+        if isinstance(bus, can.Bus):
             # this is a bit of a hack: passing any object which
             # exhibits a set_filters() method is assumed to be a can
             # bus object.
```
kayoub5, line 119: comment needed anymore?

README.md
andlaus: There still is a License section. This one was just the duplicate...

README.md
andlaus: good suggestion! done: b3359adeeb67bf4b9932beec69f790c360dfee85

odxtools/compumethods/texttablecompumethod.py
andlaus: mypy produces the following error:
```
odxtools/compumethods/texttablecompumethod.py:50: error: Incompatible return value type (got "str | int | float | bytes | None", expected "str | int | float | bytes")  [return-value]
Found 1 error in 1 file (checked 10 source files)
```

since line 49 ensures that res is not `None`, it is a false positive... (I don't think there's a practical way of properly conferring this fact to mypy.)

odxtools/compumethods/texttablecompumethod.py
andlaus: because of line 39, there should be at most one match (or else the same integer gets mapped to different values). The `scales` variable name is a bit confusing, though...

kayoub5: There could be in theory, multiple matching_scales, is there an ODX requirement that says multiple scales can't share the same physical value?

odxtools/compumethods/texttablecompumethod.py
kayoub5:
```suggestion
            return return odxrequire(res)
```

odxtools/compumethods/texttablecompumethod.py
andlaus: if I understood this correctly, that does not make sense for text tables: it would mean that e.g. the value `1` could be mapped to `"ACCELERATE"` as well as to `"BRAKE"`

kayoub5: It's the other way around, I know it's kind of an edge case, but if it can happen, it will eventually happen.
```
NO Lower, Upper 1: INVALID
2: ACCELERATE
3: BRAKE
Lower: 4, Upper: 15: INVALID
```

odxtools/compumethods/texttablecompumethod.py
kayoub5: problem caused by line 47, `res = None`

andlaus: nice catch. fixed, thanks!

odxtools/compumethods/texttablecompumethod.py
```diff
@@ -28,25 +28,30 @@ def __post_init__(self) -> None:
...
+        if len(matching_scales) == 1:
+            inverse_value: Optional[AtomicOdxType] = matching_scales[0].compu_inverse_value
+            if inverse_value is not None:
+                res = inverse_value
+            elif matching_scales[0].lower_limit is not None:
+                res = matching_scales[0].lower_limit.value
+            else:
+                res = None
+
             odxassert(self.internal_type.isinstance(res))
```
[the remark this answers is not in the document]
andlaus: okay, I removed it. (I don't really think that it hurt, though.)

odxtools/compumethods/texttablecompumethod.py
andlaus: okay, I see. fixed: 9ca3ea5acbbcc39c8e8bd76bcb504e04eb0f1c6a

(besides going over the whole list, I also included a check for the upper limit. (i.e., in theory it should now always bail out on the first scale?) )

kayoub5 (approved):

PR #213, opened 2023-09-28
Add type annotations, part 3

andlaus:
This is the third part of the quest for comprehensive type annotations. It is mainly concerned about the parameter classes, but as usual, it also contains a few drive-by fixes, namely the removal of the confusingly named and functionally unused method `DataObjectProperty.get_valid_physical_values()` and the removal of the unused `.as_dict()` methods of various classes.

### DIFF
--- a/odxtools/basicstructure.py
+++ b/odxtools/basicstructure.py
@@ -107,18 +107,18 @@ class BasicStructure(DopBase):
         print(parameter_info(self.free_parameters), end="")
 
     def convert_physical_to_internal(self,
-                                     param_values: ParameterValue,
+                                     param_value: ParameterValue,
                                      triggering_coded_request: Optional[bytes],
                                      is_end_of_pdu: bool = True) -> bytes:
 
-        if not isinstance(param_values, dict):
+        if not isinstance(param_value, dict):
             raise EncodeError(
                 f"Expected a dictionary for the values of structure {self.short_name}, "
-                f"got {type(param_values)}")
+                f"got {type(param_value)}")
 
         encode_state = EncodeState(
             b'',
-            dict(param_values),
+            dict(param_value),
             triggering_request=triggering_coded_request,
             is_end_of_pdu=False,
         )
--- a/odxtools/basicstructure.py
+++ b/odxtools/basicstructure.py
@@ -11,7 +11,7 @@ from .encodestate import EncodeState
 from .exceptions import DecodeError, EncodeError, OdxWarning, odxassert
 from .nameditemlist import NamedItemList
 from .odxlink import OdxLinkDatabase
-from .odxtypes import ParameterDict, ParameterValueDict
+from .odxtypes import ParameterDict, ParameterValue
 from .parameters.codedconstparameter import CodedConstParameter
 from .parameters.lengthkeyparameter import LengthKeyParameter
 from .parameters.matchingrequestparameter import MatchingRequestParameter
@@ -107,9 +107,14 @@ class BasicStructure(DopBase):
         print(parameter_info(self.free_parameters), end="")
 
     def convert_physical_to_internal(self,
-                                     param_values: ParameterValueDict,
+                                     param_values: ParameterValue,
                                      triggering_coded_request: Optional[bytes],
                                      is_end_of_pdu: bool = True) -> bytes:
+
+        if not isinstance(param_values, dict):
+            raise EncodeError(
+                f"Expected a dictionary for the values of structure {self.short_name}, "
+                f"got {type(param_values)}")
 
         encode_state = EncodeState(
             b'',
@@ -181,9 +186,13 @@ class BasicStructure(DopBase):
                    "\n".join(self.__message_format_lines()))
 
     def convert_physical_to_bytes(self,
-                                  param_values: ParameterValueDict,
+                                  param_values: ParameterValue,
                                   encode_state: EncodeState,
                                   bit_position: int = 0) -> bytes:
+        if not isinstance(param_values, dict):
+            raise EncodeError(
+                f"Expected a dictionary for the values of structure {self.short_name}, "
+                f"got {type(param_values)}")
         if bit_position != 0:
             raise EncodeError("Structures must be aligned, i.e. bit_position=0, but "
                               f"{self.short_name} was passed the bit position {bit_position}")
@@ -195,7 +204,7 @@ class BasicStructure(DopBase):
 
     def convert_bytes_to_physical(self,
                                   decode_state: DecodeState,
-                                  bit_position: int = 0) -> Tuple[ParameterValueDict, int]:
+                                  bit_position: int = 0) -> Tuple[ParameterValue, int]:
         if bit_position != 0:
             raise DecodeError("Structures must be aligned, i.e. bit_position=0, but "
                               f"{self.short_name} was passed the bit position {bit_position}")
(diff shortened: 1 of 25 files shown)


odxtools/basicstructure.py
```diff
@@ -107,10 +107,15 @@ def print_free_parameters_info(self) -> None:
         print(parameter_info(self.free_parameters), end="")
 
     def convert_physical_to_internal(self,
-                                     param_values: ParameterValueDict,
+                                     param_values: ParameterValue,
```
kayoub5, line 110: param_values or param_value ?

odxtools/basicstructure.py
```diff
@@ -350,8 +359,8 @@ def param_sort_key(param: Parameter) -> Tuple[int, int]:
                     # END-OF-PDU fields do not exhibit a fixed bit
                     # length, so they need special treatment here
                     dct = None
-                    if hasattr(params[param_idx], "dop"):
-                        dop = params[param_idx].dop  # type: ignore
+                    if hasattr(params[param_idx], "_dop"):
+                        dop = params[param_idx]._dop  # type: ignore[attr-defined]
```
kayoub5, line 363: Why?

odxtools/basicstructure.py
```diff
@@ -302,8 +311,8 @@ def param_sort_key(param: Parameter) -> Tuple[int, int]:
         # replace structure parameters by their sub parameters
         params: List[Parameter] = []
         for p in sorted_params:
-            if isinstance(p, ValueParameter) and isinstance(p.dop, BasicStructure):
-                params += p.dop.parameters
+            if isinstance(p, ValueParameter) and isinstance(p._dop, BasicStructure):
+                params += p._dop.parameters
```
kayoub5, line 315: Why?

odxtools/dataobjectproperty.py
```diff
@@ -174,6 +174,3 @@ def convert_bytes_to_physical(self,
 
     def is_valid_physical_value(self, physical_value):
         return self.compu_method.is_valid_physical_value(physical_value)
-
-    def get_valid_physical_values(self):
```
kayoub5, line 178: I personally use `get_valid_physical_values`, so it would be nicer to keep it.

odxtools/dopbase.py
```diff
@@ -52,11 +53,13 @@ def _resolve_snrefs(self, diag_layer: "DiagLayer") -> None:
     def is_visible(self) -> bool:
         return self.is_visible_raw in (None, True)
 
-    def convert_physical_to_bytes(self, physical_value, encode_state: EncodeState,
+    def convert_physical_to_bytes(self, physical_value: ParameterValue, encode_state: EncodeState,
```
kayoub5, line 56: convert_physical_to_bytes have a different argument name in each class, physical_value, param_values...

odxtools/parameters/parameterwithdop.py
```diff
@@ -54,15 +54,19 @@ def _resolve_snrefs(self, diag_layer: "DiagLayer") -> None:
                 spec.env_datas.get(self.dop_snref))
 
     @property
-    def dop(self) -> Optional[DopBase]:
+    def dop(self) -> DopBase:
         """may be a DataObjectProperty, a Structure or None"""
 
+        if self._dop is None:
+            odxraise("Specifying a data object property is mandatory but it could "
+                     "not be resolved")
```
kayoub5, line 62: looks like a use case of odxrequire

odxtools/parameters/reservedparameter.py
```diff
@@ -32,11 +34,11 @@ def bit_length(self) -> int:
         # need to take the "bit_length_raw" detour...
         return self.bit_length_raw
 
-    def get_coded_value_as_bytes(self, encode_state):
+    def get_coded_value_as_bytes(self, encode_state: EncodeState) -> bytes:
         bit_position_int = self.bit_position if self.bit_position is not None else 0
         return (0).to_bytes((self.bit_length + bit_position_int + 7) // 8, "big")
 
-    def decode_from_pdu(self, decode_state: DecodeState):
+    def decode_from_pdu(self, decode_state: DecodeState) -> Tuple[Any, int]:
```
kayoub5, line 41: "Any" ?

odxtools/parameters/tableentryparameter.py
```diff
@@ -22,8 +25,8 @@ def is_required(self) -> bool:
     def is_settable(self) -> bool:
         raise NotImplementedError("TableKeyParameter.is_settable is not implemented yet.")
 
-    def get_coded_value_as_bytes(self):
+    def get_coded_value_as_bytes(self, encode_state: EncodeState) -> bytes:
         raise NotImplementedError("Encoding a TableKeyParameter is not implemented yet.")
 
-    def decode_from_pdu(self, coded_message, default_byte_position=None):
+    def decode_from_pdu(self, decode_state: DecodeState) -> Tuple[Any, int]:
```
kayoub5, line 31:
```suggestion
    def decode_from_pdu(self, decode_state: DecodeState) -> Tuple[ParameterValue, int]:
```

odxtools/basicstructure.py
andlaus: both, but this is a bit twisted so let me explain: `ParameterValue` represents the value which can be expressed by any parameter while `ParameterValueDict` is a name->value dictionary for a set of parameters. That said, there are composite parameters (structures) which expect a name-to-value dictionary as their value, so a `ParameterValue` object is nested and can thus be a `ParameterValueDict`, which is exactly what is expected here.

unfortunately (or maybe luckily), we cannot restrict the signature of methods of child classes relative to their parents, so we need to use `ParameterValue` as the type annotation and ensure at runtime that it is the right kind of object, i.e., a dictionary...

odxtools/basicstructure.py
andlaus: it is slightly faster because `.dop` requires a function call...

odxtools/dataobjectproperty.py
andlaus: okay, but maybe it should be moved out of the base class because it is not possible to define this for all DOPs. (what are the valid values of a structure?)

odxtools/dopbase.py
andlaus: yes, and if it lacks proper type annotation it gets confusing rather quickly. (I think there are even instances of functions with the same name doing something slightly different in unrelated classes. This is what I mean with the de-/encoding code being quite messy at the moment.) Anyway, I'd prefer not unifying the parameter names as part of this PR series...

odxtools/parameters/reservedparameter.py
andlaus: Should be `ParameterValue`. I wonder why mypy does not complain. fixed: 83d105ce0e80d59618dec29b5002e9fc6f75fb89

odxtools/basicstructure.py
andlaus: correction: I have a dataset where the DOP cannot be resolved (in this case, because the dataset is faulty referencing a non-existant object), and for this I'll get an exception in `.dop` while I just want it to be ignored (for now)...

odxtools/basicstructure.py
andlaus: okay, I switched it to singular.

odxtools/basicstructure.py
kayoub5: disable strict mode?

odxtools/dataobjectproperty.py
kayoub5: It's fine to return an empty list in that case

andlaus: okay, the function is now still gone, but for texttables you can retrieve the same data using the compu-method's `get_scales()` method:
```python
choices = [ s.compu_const for s in my_texttable_compumethod.get_scales() ]
```

kayoub5 (approved):

PR #214, opened 2023-10-04
iso statemachine make it work for CAN FD

hacxo:
Make it work without a PDX-file

Christian Hackenbeck &lt;christian.hackenbeck@mercedes-benz.com&gt;, on behalf of [MBition GmbH](https://mbition.io/).

### DIFF
--- a/odxtools/isotp_state_machine.py
+++ b/odxtools/isotp_state_machine.py
@@ -5,6 +5,7 @@ import asyncio
 import re
 import sys
 from enum import IntEnum
+from io import TextIOWrapper
 from typing import AsyncGenerator, Iterable, List, Optional, TextIO, Tuple, Union
 
 import bitstruct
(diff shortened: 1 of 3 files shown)


andlaus (approved):
LGTM, @kayoub5

odxtools/isotp_state_machine.py
```diff
@@ -128,7 +131,7 @@ async def read_telegrams(self, bus: Union[can.Bus,
                 for tmp in self.decode_rx_frame(msg.arbitration_id, msg.data):
                     yield tmp
         else:
-            assert isinstance(bus, TextIO)
+            assert isinstance(bus, (TextIO, TextIOWrapper))
```
kayoub5, line 134: @andlaus from I have seen, TextIO is not runtime checkable, or at least there is nothing in the official docs to indicate it is

If you want to do runtime check, you most likely need to check on `io.TextIOBase`

hacxo: Thanks for your input. I added `io.TextIOBase` instead of `typing.TextIO` in the last commit.

odxtools/isotp_state_machine.py
```diff
@@ -128,7 +131,7 @@ async def read_telegrams(self, bus: Union[can.Bus,
                 for tmp in self.decode_rx_frame(msg.arbitration_id, msg.data):
                     yield tmp
         else:
-            assert isinstance(bus, TextIO)
+            assert isinstance(bus, (TextIOBase, TextIOWrapper))
```
andlaus, line 134: I suppose that this can become
```suggestion
            assert isinstance(bus, TextIOBase)
```

because `TextIOWrapper` derives from `TextIOBase`...

andlaus: (note that, to satisfy the CI, you also need to remove the now unused import of `TextIOWrapper` above...)

hacxo: I agree. `TextIOWrapper` is indeed derived from `TextIOBase` (src: https://docs.python.org/3/library/io.html#io.TextIOWrapper)

kayoub5 (approved):

PR #215, opened 2023-10-05
Add type annotations, part 4

andlaus:
The next episode of the type annotation quest. This one should be pretty light on drive-by fixes...

### DIFF
--- a/odxtools/basicstructure.py
+++ b/odxtools/basicstructure.py
@@ -225,7 +225,7 @@ class BasicStructure(DopBase):
 
         return inner_decode_state.parameter_values, decode_state.cursor_position + inner_decode_state.cursor_position
 
-    def encode(self, coded_request: Optional[bytes] = None, **params: Any) -> bytes:
+    def encode(self, coded_request: Optional[bytes] = None, **params: ParameterValue) -> bytes:
         """
         Composes an UDS message as bytes for this service.
         Parameters:
--- a/odxtools/basicstructure.py
+++ b/odxtools/basicstructure.py
@@ -2,7 +2,7 @@
 import math
 import warnings
 from dataclasses import dataclass
-from typing import TYPE_CHECKING, List, Optional, Tuple
+from typing import TYPE_CHECKING, Any, Dict, List, Optional, Tuple
 
 from .dataobjectproperty import DataObjectProperty
 from .decodestate import DecodeState
@@ -10,7 +10,7 @@ from .dopbase import DopBase
 from .encodestate import EncodeState
 from .exceptions import DecodeError, EncodeError, OdxWarning, odxassert
 from .nameditemlist import NamedItemList
-from .odxlink import OdxLinkDatabase
+from .odxlink import OdxLinkDatabase, OdxLinkId
 from .odxtypes import ParameterDict, ParameterValue
 from .parameters.codedconstparameter import CodedConstParameter
 from .parameters.lengthkeyparameter import LengthKeyParameter
@@ -225,7 +225,7 @@ class BasicStructure(DopBase):
 
         return inner_decode_state.parameter_values, decode_state.cursor_position + inner_decode_state.cursor_position
 
-    def encode(self, coded_request: Optional[bytes] = None, **params) -> bytes:
+    def encode(self, coded_request: Optional[bytes] = None, **params: Any) -> bytes:
         """
         Composes an UDS message as bytes for this service.
         Parameters:
@@ -276,7 +276,7 @@ class BasicStructure(DopBase):
         })
         return param_dict
 
-    def _build_odxlinks(self):
+    def _build_odxlinks(self) -> Dict[OdxLinkId, Any]:
         result = super()._build_odxlinks()
 
         for p in self.parameters:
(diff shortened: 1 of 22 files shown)


odxtools/basicstructure.py
```diff
@@ -66,9 +68,8 @@ def coded_const_prefix(self, request_prefix: bytes = b'') -> bytes:
         prefix = b''
         encode_state = EncodeState(prefix, parameter_values={}, triggering_request=request_prefix)
         for p in self.parameters:
-            if isinstance(p, CodedConstParameter) and p.bit_length % 8 == 0:
-                encode_state.coded_message = p.encode_into_pdu(encode_state)
-            elif isinstance(p, MatchingRequestParameter):
+            if isinstance(p, (CodedConstParameter, NrcConstParameter, MatchingRequestParameter,
+                              PhysicalConstantParameter)):
```
kayoub5, line 72: Use `p.is_settable` instead?

odxtools/basicstructure.py
```diff
@@ -224,7 +225,7 @@ def convert_bytes_to_physical(self,
 
         return inner_decode_state.parameter_values, decode_state.cursor_position + inner_decode_state.cursor_position
 
-    def encode(self, coded_request: Optional[bytes] = None, **params) -> bytes:
+    def encode(self, coded_request: Optional[bytes] = None, **params: Any) -> bytes:
```
kayoub5, line 228: Sure about Any?

odxtools/minmaxlengthtype.py
```diff
@@ -51,15 +51,28 @@ def __termination_sequence(self) -> bytes:
...
-        bit_length = 8 * self.max_length if self.max_length is not None else 8 * len(internal_value)
-        bit_length = min(8 * len(internal_value), bit_length)
+        if not isinstance(internal_value, (bytes, str)):
+            odxraise("MinMaxLengthType is currently only implemented for strings and byte arrays",
+                     EncodeError)
+
+        # explictly tell mypy that the internal value is a bytes
+        # object or a string. (for whatever reason, the condition
+        # above does not suffice...)
+        if TYPE_CHECKING:
```
kayoub5, line 63: If possible, remove the if check

odxtools/basicstructure.py
andlaus: that would be not correct from a logical POV: what is required here is to determine the "constant prefix" of a structure (i.e., the part which allows to identify it), i.e., all parameters which are constant -- meaning that they are defined statically or only depending on previous communication (`MatchingRequestParameter`). The problem is that some parameters parameters are not settable, but can still not be used for identification (e.g. `ReservedParameter`)

maybe we can introduce a new property ('.is_prefixable`?) for this purpose, but I don't think that the complexity costs of this would be currently worth it...

kayoub5: The way I see it, a `ReservedParameter` should also be part of the prefix, any other examples?

odxtools/minmaxlengthtype.py
andlaus: it worked! yay!

odxtools/basicstructure.py
andlaus: That was a quick and dirty way to get around mypy errors: for the `kwargs` dict, mypy seems to assume that the type of the annotation is used for *all* keyword arguments, so just setting this to `Any` always works.

In this place, using `ParameterValue` would be better, though (it is almost the same as `Any` by now, though). fixed, thanks: 9c7ecc0818b4cb4b9eedde9b13f3eb8599f78eae

odxtools/basicstructure.py
andlaus: `ReservedParameter` cannot be made part of the prefix because its purpose is to specify that the data it covers that is send at runtime is undefined. (possibly, we can make it settable, but that would be rather inelegant..)

another example is (maybe) `SystemParameter`, but this one is currently a stub and throws `NotImplementedError` upon accessing `.is_settable` (as do a few others...)

kayoub5: I have seen cases where `Suppress Positive Response Message Indication Bit` was implemented as a `ReservedParameter`.

I don't see a reason to explicitly exclude it.

kayoub5: If that's the case, then the logic would be fine for this PR. 

As a side note, the logic you have now in ReservedParameter raise an error for non zero values when decoding, that doesn't match the statement of "The file format does not specify any value for such parameters"

andlaus: hm, that got me scratching my head, and I thus looked up the spec. (cf p. 78 and 79): It says that all bits shall be ignored, but it also says that all its bits shall be set to zero, which IMO is a contradiction. I suppose that for the purpose of encoding reserved parameters, all its bits should be set to 0 (because what else can we do?) whilst during decoding these parameters ought to be ignored. I think that the words of the spec can be interpreted that way, but it is formulated quite confusingly. Agree?

(In the mean time, I have added a patch to the series which does just that: bd853f5c91504cf1bf7105020fc29fbeccd4b626)

kayoub5 (approved):

PR #216, opened 2023-10-06
Drive by fixes #2

### LINKED ISSUE #2: Better API documentation
The API documentation is generated from the python docstrings and it can be generated and inspected via
```bash
cd $ODXTOOLS_SRC_DIR/doc
make html
firefox _build/html/index.html
```

It IMO already looks decent, but it is sorely lacking in completeness and does not feature any "tutorial style" introductions. this should be changed in the medium to long term.

andlaus:
This PR consists of fixes for issues I've noticed while adding type annotations up to [#215](https://github.com/mercedes-benz/odxtools/pull/215). By far the biggest change is refactoring the `.bit_length` property of DOPs, DiagCodedTypes, and parameters to a `.get_static_bit_length()` method in order to make it clearer that this information might be calculated (and thus a call might be relative expensive), and that it is not about the coded size of an object required at run time (this can only be computed given the concrete values of all required parameters). 


odxtools/basicstructure.py
```diff
@@ -31,38 +30,28 @@ class BasicStructure(DopBase):
...
+        length = 0
+        for param in self.parameters:
+            param_bit_length = param.get_static_bit_length()
+            if param_bit_length is None:
+                # We were not able to calculate a static bit length
+                return None
+            elif param.byte_position is not None:
+                bit_pos = param.bit_position or 0
+                byte_pos = param.byte_position or 0
+                cursor = byte_pos * 8 + bit_pos
```
kayoub5, line 48: if existing cursor value is bigger, keep it (use max function)

kayoub5, line 48: please double check the spec on this, if I recall correctly, if no byte offset is specified, the max (and not the last) of previous params position+length is used

odxtools/determinenumberofitems.py
```diff
@@ -1,18 +1,47 @@
...
+from .exceptions import odxrequire
+from .odxlink import OdxDocFragment, OdxLinkDatabase, OdxLinkId, OdxLinkRef
+
+if TYPE_CHECKING:
+    from ..diaglayer import DiagLayer
 
 
 @dataclass
-class DetermineNumberOfItems(PositionedDataObjectProperty):
+class DetermineNumberOfItems:
```
kayoub5, line 14: why?

odxtools/field.py
```diff
@@ -66,9 +66,8 @@ def structure(self) -> BasicStructure:
         """may be a Structure or a env-data-desc"""
         return odxrequire(self._structure, "EnvironmentDataDescription is not supported")
 
-    @property
-    def bit_length(self):
-        return self.structure.bit_length
+    def get_static_bit_length(self) -> Optional[int]:
```
kayoub5, line 69: sure about this change?

odxtools/parameters/nrcconstparameter.py
```diff
@@ -47,8 +47,7 @@ def _resolve_odxlinks(self, odxlinks: OdxLinkDatabase) -> None:
     def _resolve_snrefs(self, diag_layer: "DiagLayer") -> None:
         super()._resolve_snrefs(diag_layer)
 
-    @property
-    def bit_length(self) -> Optional[int]:
+    def get_static_bit_length(self) -> Optional[int]:
         return getattr(self.diag_coded_type, "bit_length", None)
```
kayoub5, line 51: bit_length?

odxtools/positioneddataobjectproperty.py
```diff
@@ -1,78 +0,0 @@
...
-from .exceptions import odxrequire
-from .odxlink import OdxDocFragment, OdxLinkDatabase, OdxLinkId, OdxLinkRef
-from .odxtypes import ParameterValue
-
-if TYPE_CHECKING:
-    from .diaglayer import DiagLayer
-
-
-@dataclass
-class PositionedDataObjectProperty:
```
kayoub5, line 18: why? why? why?

odxtools/basicstructure.py
andlaus: the max is on line 51 (`cursor` is the position of the first bit where the parameter would be encoded)

odxtools/field.py
andlaus: pretty much: to me, `.bit_length` gives the impression that it is just some data from the XML (which it sometimes indeed is) and not something which usual is calculated or not available at all. I'm open for better solutions/names, though.

odxtools/parameters/matchingrequestparameter.py
```diff
@@ -41,8 +40,8 @@ def get_coded_value_as_bytes(self, encode_state: EncodeState) -> bytes:
     def decode_from_pdu(self, decode_state: DecodeState) -> Tuple[ParameterValue, int]:
         byte_position = (
             self.byte_position if self.byte_position is not None else decode_state.cursor_position)
-        bit_position = self.bit_position if self.bit_position is not None else 0
-        byte_length = (self.bit_length + bit_position + 7) // 8
+        bit_position = self.bit_position or 0
+        byte_length = (odxrequire(self.get_static_bit_length()) + bit_position + 7) // 8
```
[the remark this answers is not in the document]
andlaus: good point. fixed: f631778184791e0cea8a91890029bddd223510f7

odxtools/parameters/nrcconstparameter.py
andlaus: right. fixed: bf17811b63e71fcff6f3ef5ef54d402f48093d44

odxtools/positioneddataobjectproperty.py
andlaus: because `PositionedDataObjectProperty` is not a `DataObjectProperty`, it does not appear in the XSD, and it is only used in two places where for each occurrence about of 10 lines of code are required to replace it. (I have concluded that it is best for our sanity to match the hierarchy of python classes as closely with the one defined by the XSD as possible. Work in progress...)

odxtools/basicstructure.py
kayoub5: Take this example:
* P1 (len = 4, pos = 2)
* P2 (len = 1, pos = 1)
* P3 (len = 2, pos = None)
The spec says P3 should be encoded from position 6, current code encode it from position 2
Thus total struct size would end up being 4 instead of 8.

odxtools/determinenumberofitems.py
kayoub5: 40 lines is not "small"

odxtools/field.py
kayoub5: I am asking about returning None instead of `self.structure.bit_length`

odxtools/basicstructure.py
andlaus: okay, I looked it up: the spec says "The parameter starts then at the byte edge next to the end of the last parameter."

now it is a matter of interpretation what "last parameter" is supposed to mean. I can see three possibilities:
1) the parameter which has its bytes located furthest back
2) the parameter which we've encountered at that point that has its last coded bit located most to the east
3) the parameter immediately before the currently considered one

the first option does IMO not make any sense at all, the second is what you remember and the third option is the current implementation. I think that the third option is the most likely one, but I would not be too surprised if some vendors went with 2) while others took option 3).

kayoub5: In this case, it's fine to choose either, but make sure both encoding logic and bit length calculation logic are using the same logic.

PS: double check the 2.0 spec, it could have different/better wording.

odxtools/determinenumberofitems.py
andlaus: it's 38 lines less than `positioneddataobjectproperty.py` ;)

odxtools/field.py
andlaus: this is the base class for all fields, and as far as I can see, a static size can only be specified for `STATIC-FIELD` if it uses a structure with a static size. (Dynamically sized fields have a -- ehem -- dynamic length and also, `Structure.bit_length` does not exist anymore after this patch.) Since this is the base class for all fields, we should use the same approach as e.g. `DopBase` and return `None` whilst the derived classes override it if it makes sense...

kayoub5 (approved):

odxtools/basicstructure.py
andlaus: > PS: double check the 2.0 spec, it could have different/better wording.

okay, did that. unfortunately this part was copy-and-pasted from 2.0 to 2.2: `Sometimes, the position of a parameter cannot be determined until runtime, [...] the parameter starts then at the byte edge next to the end of the last parameter.`

andlaus:
let's merge this and get on to greener patures!

PR #217, opened 2023-10-09
Add type annotations, part 5

andlaus:
After this, `mypy --disallow-untyped-defs` does not complain about anything of the "core" odxtools anymore. To do: The CLI tools, unit tests and examples.

### DIFF
--- a/tests/test_diag_data_dictionary_spec.py
+++ b/tests/test_diag_data_dictionary_spec.py
@@ -293,11 +293,6 @@ class TestDiagDataDictionarySpec(unittest.TestCase):
         self.assertEqual(len(ddds.structures), 0)
         self.assertEqual(len(ddds.end_of_pdu_fields), 0)
 
-        self.assertEqual({x.short_name
-                          for x in ddds.all_data_object_properties},
-                         {x.short_name
-                          for x in (dtc_dop, dop_1, dop_2, table)})
-
 
 if __name__ == "__main__":
     unittest.main()
--- a/tests/test_odxtools.py
+++ b/tests/test_odxtools.py
@@ -12,7 +12,7 @@ odxdb = load_pdx_file("./examples/somersault.pdx")
 
 # use the diag layer container's document fragments as the default for
 # resolving references
-container_doc_frags = odxdb.diag_layer_containers.somersault.odx_id.doc_fragments
+container_doc_frags = odxdb.diag_layer_containers["somersault"].odx_id.doc_fragments
 
 
 class TestStrictMode(unittest.TestCase):
--- a/examples/somersaultecu.py
+++ b/examples/somersaultecu.py
@@ -1247,7 +1247,7 @@ somersault_env_datas = {
             is_visible_raw=None,
             byte_size=None,
             dtc_values=[],
-            parameters=[
+            parameters=NamedItemList([
                 ValueParameter(
                     short_name="flip_speed",
                     long_name="Flip Speed",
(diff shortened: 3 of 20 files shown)


tests/test_odxtools.py
```diff
@@ -12,7 +12,7 @@
 
 # use the diag layer container's document fragments as the default for
 # resolving references
-container_doc_frags = odxdb.diag_layer_containers.somersault.odx_id.doc_fragments
+container_doc_frags = odxdb.diag_layer_containers["somersault"].odx_id.doc_fragments
```
kayoub5, line 15: If you can't do `odxdb.diag_layer_containers.somersault.odx_id.doc_fragments` then `NamedItemList` type hinting is not correct

tests/test_diag_data_dictionary_spec.py
```diff
@@ -293,11 +293,6 @@ def test_initialization(self):
         self.assertEqual(len(ddds.structures), 0)
         self.assertEqual(len(ddds.end_of_pdu_fields), 0)
 
-        self.assertEqual({x.short_name
-                          for x in ddds.all_data_object_properties},
-                         {x.short_name
-                          for x in (dtc_dop, dop_1, dop_2, table)})
```
kayoub5, line 299: Why?

tests/test_odxtools.py
andlaus: the problem is that these attributes are all dynamically added at runtime (by extending `__dict__` of the named item list) and there is no way that mypy can be aware of this. (I also cannot come up with a "clean" way to tell `mypy` short of modifying the `mypy` source code...)

kayoub5: according to mypy issue tracker, it is supported. 
https://github.com/python/mypy/pull/544
https://github.com/python/mypy/issues/520

tests/test_diag_data_dictionary_spec.py
andlaus: This is a drive-by fix which I failed to catch when pruning this series for the PR. (Sorry for that. I can move it to a separate PR if you want.)

The reason for this is that I came to the conclusion that the `all_data_object_properties` attribute does more harm than good, i.e., it is terribly easy to forget adapting it to changes elsewhere and there are no real users within the codebase. (the only valid use case which I can come up with out of my head is resolving SNREFs to DOPs, but even for this that attribute would only be useful if such SNREFs encompass *all* DOP objects, not just structures, simple DOPs, environment datas, etc and I'm not sure if this is even possible because the short names of these object categories might be allowed to collide.)

tests/test_odxtools.py
andlaus: interesting, I didn't know about `__getattr__()` (it still feels a bit like voodoo to me). Anyway, I've adapted the implementation of `ItemList`: 17221463c280b9d8d95201fd14c0f173c415a513 (this is a drive-by patch but I hope this is okay for you in this instance...)

odxtools/nameditemlist.py
```diff
@@ -34,8 +35,8 @@ class ItemList(Generic[T]):
     """
 
     def __init__(self, input_list: Optional[Iterable[T]] = None) -> None:
-        self._names: List[str] = []
-        self._values: List[T] = []
+        self._item_dict: OrderedDict[str, T] = OrderedDict()
+        self._item_list: List[T] = []
```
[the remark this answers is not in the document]
andlaus: I haven't found a good way of accessing OrderedDict by integer index  and with slicing supported...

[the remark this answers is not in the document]
andlaus: (in principle, this should be relatively easy to implement this by OrderedDict, but it seems like it simply does not expose this functionality...)

kayoub5 (approved):

PR #218, opened 2023-10-10
Add type annotations, part 6

andlaus:
This time, the command line tools are annotated. It turned out that this was surprisingly difficult...

### DIFF
--- a/odxtools/cli/browse.py
+++ b/odxtools/cli/browse.py
@@ -7,12 +7,15 @@ from typing import List, Optional, Union
 import PyInquirer.prompt as PI_prompt
 
 from ..database import Database
+from ..dataobjectproperty import DataObjectProperty
 from ..diaglayer import DiagLayer
 from ..diagservice import DiagService
-from ..odxtypes import DataType, ParameterValueDict
+from ..exceptions import odxraise, odxrequire
+from ..odxtypes import AtomicOdxType, DataType, ParameterValueDict
 from ..parameters.matchingrequestparameter import MatchingRequestParameter
 from ..parameters.parameter import Parameter
 from ..parameters.parameterwithdop import ParameterWithDOP
+from ..parameters.valueparameter import ValueParameter
 from ..request import Request
 from ..response import Response
 from . import _parser_utils
@@ -21,7 +24,7 @@ from . import _parser_utils
 _odxtools_tool_name_ = "browse"
 
 
-def _convert_string_to_odx_type(string_value: str, odx_type: DataType):
+def _convert_string_to_odx_type(string_value: str, odx_type: DataType) -> AtomicOdxType:
     """Similar to odx_type.from_string(string_value) but more relaxed to parse user input"""
     if odx_type == DataType.A_UINT32:
         return int(string_value, 0)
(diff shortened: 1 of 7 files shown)


odxtools/cli/browse.py
```diff
@@ -31,28 +34,38 @@ def _convert_string_to_odx_type(string_value: str, odx_type: DataType):
...
         try:
-            val = _convert_string_to_odx_type(input, parameter.physical_type.base_data_type)
+            phys_type = odxrequire(parameter.physical_type)
+            val = _convert_string_to_odx_type(input, phys_type.base_data_type)
         except:  # noqa: E722
             return False
-        return parameter.dop.is_valid_physical_value(val)
+        dop = parameter.dop
+        if isinstance(dop, DataObjectProperty):
+            return dop.is_valid_physical_value(val)
```
kayoub5, line 55: is_valid_physical_value, is defined in base dop, right? so it's the implementation of dop sub class that should raise the not implemented error

odxtools/cli/browse.py
```diff
@@ -101,12 +115,12 @@ def encode_message_interactively(sub_service, ask_user_confirmation=False):
     for param_or_dict in param_dict.values():
         if isinstance(param_or_dict, dict):
             for param in param_or_dict.values():
-                if param.is_settable:
+                if not isinstance(param_or_dict, dict) and param.is_settable:
```
kayoub5, line 118: why are you testing on dict?

odxtools/cli/snoop.py
```diff
@@ -61,29 +62,30 @@ def handle_telegram(telegram_id, payload):
...
-              f"({payload}, {len(payload)} bytes)")
+              f"({payload!r}, {len(payload)} bytes)")
 
 
-def init_verbose_state_machine(BaseClass, *args, **kwargs):
+def init_verbose_state_machine(BaseClass: Type[IsoTpStateMachine], *args: Any,
+                               **kwargs: Any) -> IsoTpStateMachine:
 
-    class InformativeIsoTpDecoder(BaseClass):
+    class InformativeIsoTpDecoder(BaseClass):  # type: ignore[valid-type, misc]
```
kayoub5, line 71: why?

odxtools/cli/browse.py
andlaus: Just checked: it only available for simple DOPs (i.e., `DataObjectProperty`, not `DopBase`) but you're right: it should probably me moved (I'll deal with it on Monday...)

odxtools/cli/browse.py
andlaus: param is of type `ParameterValue`, in this particular location we want to ask the user for a value since querying structures is not (yet?) implemented...

odxtools/cli/snoop.py
andlaus: mypy has trouble understanding the class ought to be derived from the class type which is passed as a function argument. unfortunately, I couldn't find a good solution...

odxtools/cli/browse.py
andlaus: okay, fixed: 7e4a516cec685a4bb25b1029cfed62070c8bf863. thanks for noticing!

kayoub5:
@andlaus could you publish a release after merging this? you have almost a 100 commits in main branch since last release

andlaus:
yes, I plan to do so after the last type annotations PR, and I intend to declare this to be a major release. (I've already have these patches finished locally...)

PR #222, opened 2023-10-17
NamedItemList: make tab completion for NamedItemList work again

andlaus:
it turns out that an *explicit* list with all possibilities is required for the REPL tab completion code to show the list of all possibilities (duh!). (The `__getattr__()` method is obviously implicit, but we still need it to keep mypy happy...)

note that this change doesn't have an effect on scripts because tab completion is a purely interactive thing.

### DIFF
--- a/odxtools/nameditemlist.py
+++ b/odxtools/nameditemlist.py
@@ -2,8 +2,8 @@
 import abc
 from collections import OrderedDict
 from keyword import iskeyword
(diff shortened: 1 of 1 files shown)


andlaus:
(also note that calling `dir(name_item_list)` is quite slow and memory intensive. That should not matter though, as these objects are not supposed to contain more than a few thousand items...)

kayoub5 (approved):

PR #223, opened 2023-10-18
Minor cleanups

andlaus:
This PR contains a few cleanups for stuff which I stumbled over yesterday:

- Removal of `is_functional` for communication parameters
- Make `DiagLayer.protocols` a cached property


kayoub5:
> Removal of is_functional for communication parameters

Why?

odxtools/diaglayerraw.py
```diff
@@ -163,7 +163,7 @@ def from_et(et_element: ElementTree.Element, doc_frags: List[OdxDocFragment]) ->
         ]
 
         communication_parameters = [
-            CommunicationParameterRef.from_et(el, doc_frags, variant_type)
+            CommunicationParameterRef.from_et(el, doc_frags)
```
kayoub5, line 166: variant_type is not used outside of here, maybe remove it?

andlaus:
> > Removal of is_functional for communication parameters
> 
> Why?

mainly because DIAG-SERVICE tags may have a COMPARAM-REFS sub-element  and it is completely impossible to determine if the parameter is supposed to be "functional" or not. besides that, I'm not sure what it even means for a communication parameter to be functional, and finally, there were no in-tree users or tests...

odxtools/diaglayerraw.py
andlaus: ok, I tried this, but *actually* it is used in a few places, e.g. https://github.com/mercedes-benz/odxtools/blob/main/odxtools/diaglayer.py#L589-L590

kayoub5: I am talking about the local variable used only in this function (declared in line 74), not about all occurrences of `variant_type`

kayoub5:
> > > Removal of is_functional for communication parameters
> > 
> > 
> > Why?
> 
> mainly because DIAG-SERVICE tags may have a COMPARAM-REFS sub-element and it is completely impossible to determine if the parameter is supposed to be "functional" or not. besides that, I'm not sure what it even means for a communication parameter to be functional, and finally, there were no in-tree users or tests...

@dmholtz was the one who introduced the `is_functional`, maybe he can shed some light on the topic?

odxtools/diaglayer.py
```diff
@@ -720,16 +717,13 @@ def get_doip_logical_ecu_address(self, protocol_name: Optional[str] = None) -> O
         return int(ecu_addr)
 
     def get_doip_logical_gateway_address(self,
-                                         protocol_name: Optional[str] = None,
-                                         is_functional: Optional[bool] = False) -> Optional[int]:
```
dmholtz, line 724: Removing the `is_functional` parameter in any of these methods is a breaking change and should result in a new major release of odxtools if the removal is really necessary.

odxtools/communicationparameterref.py
```diff
@@ -41,7 +39,6 @@ def from_et(et_element: ElementTree.Element, doc_frags: List[OdxDocFragment],
         else:
             value = create_complex_value_from_et(odxrequire(et_element.find("COMPLEX-VALUE")))
 
-        is_functional = dl_type == DiagLayerType.FUNCTIONAL_GROUP
```
dmholtz, line 44: This is the critical line of code where the parameter is set.

I used the DiagLayer's type to determine if a communication parameter is used for functional addressing or not. The reason for this heuristic is that all of the pdx files I worked with last year provided two almost indistinguishable communication parameters, one for functional addressing and one for non-functional addressing. Apart the surrounding DiagLayer and its type there is afair no other indication to what type of addressing the parameter refers to.

I suggest not to remove this heuristic unless a reasonable alternative is found, because some of our current projects use the query methods in the `DiagLayer` class, which depend on this parameter.

dmholtz (changes requested):

odxtools/communicationparameterref.py
andlaus: the problem is that DiagServices may also contain referencces to communication parameters and this heuristic is not available because diagnostic services are inherited, i.e., does the commparam change its "allegiance" if it is inherited by an ECU-VARIANT that is derived from a FUNCTIONAL-GROUP?

kayoub5: @andlaus the alternative to `is_functional` would be to expose to what `DiagLayerType` the comm param belongs to.

dmholtz: > I suggest not to remove this heuristic unless a reasonable alternative is found, because some of our current projects use the query methods in the DiagLayer class, which depend on this parameter.

As this seems to be an issue with our PDX parameter naming conventions, I will forward this to my colleagues, who will eventually check if we can fix this internally without stopping the development of odxtools

odxtools/diaglayerraw.py
andlaus: That one is passed to `DiagLayerRaw` on line 181...

odxtools/communicationparameterref.py
dmholtz: We have revisted our usage of this part of the API and found that there exists a fix for the behavior described above (at least) in the most recent version of odxtools. In case two communcation parameters have the same name, they can be distinguished by simply also providing the protocol name to the query.

The `is_functional` parameter / attribute is indeed no longer required and can be removed.

odxtools/diaglayer.py
dmholtz: This is still true IMO.

dmholtz (approved):

andlaus:
cool! let's get this merged then!

PR #225, opened 2023-10-20
`dataclass_fields_asdict()`: do not use internals of the dataclass implementation

andlaus:
instead, use `inspect.signature()`. This should slightly reduce the long-term maintainance burden because even if the internals of dataclasses change over time, `inspect.signature()` should continue to work...

### DIFF
--- a/odxtools/utils.py
+++ b/odxtools/utils.py
@@ -1,5 +1,6 @@
 # SPDX-License-Identifier: MIT
 import re
+from inspect import signature
 from typing import Any, Dict, Optional
 from xml.etree import ElementTree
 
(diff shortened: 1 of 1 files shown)


odxtools/utils.py
```diff
@@ -34,7 +35,7 @@ def dataclass_fields_asdict(obj: Any) -> Dict[str, Any]:
     This makes hierarchies of dataclasses possible while initializing
     the base class using common code.
     """
-    return {x.name: getattr(obj, x.name) for x in obj.__dataclass_fields__.values()}
+    return {x: getattr(obj, x) for x in signature(type(obj)).parameters}
```
kayoub5, line 38: inspect is usually slow, are you sure there are no performance side effects?

kayoub5: Maybe use `fields` ? https://docs.python.org/3/library/dataclasses.html#dataclasses.fields

andlaus: good point. I got a 30% performance regression when listing one of my bigger files (which is dominated by loading the database). Using your suggestion leaves peformance virtually unchanged. Thanks for the proposal, fixed: 54f928968ea3e76c2ef30d594c9dcbbd33b29cd6 .

(in this commit I also extended the docstring with the reason why we need `dataclass_fields_asdict()` in the first place...)

kayoub5 (approved):

PR #226, opened 2023-10-20
diag_layer: allow to access services according to their UDS service group (SID)

andlaus:
This is a convenience functionality to quickly find the diagnostic services one is interested in for ECUs offering lots of services. The simplest way to use it is using the `.print_info()` method:

```python
ecu.uds_service_groups.print_info()
```

That said, bins can also be programmatically processed:

```python
session_services = ecu.uds_service_groups[0x10]
print("The services concerning diagnostic sessions are:")
for session in session_services:
    print(f"  {service.short_name}")
```

### DIFF
--- a/odxtools/parameters/codedconstparameter.py
+++ b/odxtools/parameters/codedconstparameter.py
@@ -40,7 +40,7 @@ class CodedConstParameter(Parameter):
         super()._resolve_snrefs(diag_layer)
 
     def get_static_bit_length(self) -> Optional[int]:
-        return getattr(self.diag_coded_type, "bit_length", None)
+        return self.diag_coded_type.get_static_bit_length()
 
     @property
     def internal_data_type(self) -> DataType:
--- a/odxtools/udsbinner.py
+++ b/odxtools/udsbinner.py
@@ -1,0 +1,123 @@
+# SPDX-License-Identifier: MIT
+from typing import Dict, Iterable, Optional
+
+from .diagservice import DiagService
+from .nameditemlist import NamedItemList
+from .parameters.codedconstparameter import CodedConstParameter
+
+
+class UdsBinner:
+    """Class to categorize a long list of services into the service
+    groups defined by the UDS standard.
+
+    This class is supposed to be used like this:
+
+    db = odxtools.load_file("my_cool_diagnostics_db.pdx")
+    ...
+    uds_bins = UdsBinner(db.ecus.my_ecu)
+    uds_bins.print_info()
+    """
+
+    def __init__(self, services: Iterable[DiagService]):
+        service_groups: Dict[Optional[int], NamedItemList[DiagService]] = {}
+        for service in services:
+            SID = self.__extract_sid(service)
+
+            if SID not in service_groups:
+                service_groups[SID] = NamedItemList()
+
+            service_groups[SID].append(service)
+
+        self.service_groups = service_groups
+
+    def __extract_sid(self, service: DiagService) -> Optional[int]:
+        # diagnostic services without requests are possible; just like
+        # aircraft without wings...
+        if service.request is None:
+            return None
+
+        prefix = 0  # prefix of constant parameters
+        cursor = 0  # bit position of the next parameter
+        for param in service.request.parameters:
+            if not isinstance(param, CodedConstParameter):
+                # we *need* at least the first byte of a request to be statically defined!
+                return None
+
+            param_len = param.get_static_bit_length()
+            if param_len is None:
+                return None
+
+            if not isinstance(param.coded_value, int):
+                return None
+
+            prefix <<= param_len
+            prefix |= param.coded_value & ((1 << param_len) - 1)
+            cursor += param_len
+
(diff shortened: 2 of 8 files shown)


odxtools/parameters/codedconstparameter.py
```diff
@@ -40,7 +40,7 @@ def _resolve_snrefs(self, diag_layer: "DiagLayer") -> None:
         super()._resolve_snrefs(diag_layer)
 
     def get_static_bit_length(self) -> Optional[int]:
-        return getattr(self.diag_coded_type, "bit_length", None)
+        return self.diag_coded_type.get_static_bit_length()
```
kayoub5, line 43: Drive by?

odxtools/udsbinner.py
```diff
@@ -0,0 +1,123 @@
...
+                print(f"Non standard services:")
+            for service in service_list:
+                print(f"  {service.short_name}")
+
+    #: map from the ID of a service group to its human-readable name
+    #: as defined by the UDS standard
+    sid_to_name = {
+        # diagnostic services that cannot be categorized in terms of
+        # the UDS standard get `None` as their SID...
+        None: "Not Applicable",
```
kayoub5, line 86:
```suggestion
        None: "Unknown",
```

odxtools/udsbinner.py
```diff
@@ -0,0 +1,123 @@
...
+        0x37: "Request Transfer Exit",
+        0x38: "Request File Transfer",
+    }
+
+    def __getitem__(self, sid: Optional[int]) -> NamedItemList[DiagService]:
+        if sid is None:
+            return self.service_groups.get(sid, NamedItemList())
+
+        if sid < 0 or sid > 255:
+            raise AttributeError(f"SIDs must be in range 0x00 to 0xff")
```
kayoub5, line 121: Index or Attribute Error?

odxtools/udsbinner.py
```diff
@@ -0,0 +1,123 @@
...
+
+            if cursor >= 8:
+                # we have a prefix that is at least 8 bits
+                # long. return its most significant byte.
+                prefix >>= cursor - 8
+                return prefix & 0xff
+
+        return None
+
+    def print_info(self) -> None:
```
kayoub5, line 65: return the string, let user do the printing
```suggestion
    def __str__(self) -> None:
```

andlaus: great suggestion! implemented: 29a305a4c7c12757381e91938c8c6bf6e199e5f1

odxtools/udsbinner.py
andlaus: Index. fixed: 23513093ccc79254596265bc197fd40808273a37

odxtools/udsbinner.py
andlaus: IMO, in this case, "not applicable" is -- err -- more applicable: This is meant for all services which do not conform to the UDS specification at all, i.e., the first byte of their request cannot be statically determined. "Unknown" SIDs are SIDs which have not (yet?) been specified by UDS (or ones which I missed to add...)

odxtools/parameters/codedconstparameter.py
andlaus: yes. this is too small to factor out, though?

kayoub5: include in changelog, important bug fix

odxtools/udsbinner.py
```diff
@@ -0,0 +1,128 @@
...
+        return None
+
+    def __str__(self) -> str:
+        """Return an informative string about which of the diagnostic
+        services are part of which UDS service group.
+
+        """
+        result = StringIO()
+        for sid, service_list in self.service_groups.items():
+            sid_name = self.sid_to_name.get(sid)
```
kayoub5, line 73: will throw for unknown service,

```suggestion
            sid_name = self.sid_to_name.get(sid, hex(sid))
```

odxtools/udsbinner.py
kayoub5: - there are other standard diag services other than UDS, for example OBD II (range 1 to 0xa)
- assigning all services that you don't recognize the same name is not nice, instead return sid as hex

odxtools/parameters/codedconstparameter.py
andlaus: I put it in a separate PR. I don't think that it currently changes much behavior-wise because the only diag coded type that matters which has the `.bit_length` property/implements `get_static_bit_length()` is `StandardLengthType`...

odxtools/udsbinner.py
andlaus: I refactored this code with the goals of including the OBD services and using the existing `.uds` and `.obd` sub-packages...

odxtools/udsbinner.py
andlaus: I refactored the code a bit. I hope these issues are solved with that...

kayoub5 (approved):

PR #227, opened 2023-10-21
Split find into find and decode

floroks:
Fix issue with find command requiring data, by splitting the decode and find (by service name) functionalities
into separate commands.

### DIFF
--- a/odxtools/cli/decode.py
+++ b/odxtools/cli/decode.py
@@ -1,0 +1,130 @@
+# SPDX-License-Identifier: MIT
+import argparse
+import re
+from typing import Dict, List, Optional
+
+from ..database import Database
+from ..diagservice import DiagService
+from ..exceptions import odxraise
+from ..odxtypes import ParameterValue
+from ..singleecujob import SingleEcuJob
+from . import _parser_utils
+
+# name of the tool
+_odxtools_tool_name_ = "decode"
+
+
+def get_display_value(v: ParameterValue) -> str:
+    if isinstance(v, bytes):
+        return v.hex(" ")
+    elif isinstance(v, int):
+        return f"{v} (0x{v:x})"
+    else:
+        return str(v)
+
+
+def print_summary(
+    odxdb: Database,
+    ecu_variants: Optional[List[str]] = None,
+    data: bytes = b'',
+    decode: bool = False,
+) -> None:
+    ecu_names = ecu_variants if ecu_variants else [ecu.short_name for ecu in odxdb.ecus]
+    services: Dict[DiagService, List[str]] = {}
+    for ecu_name in ecu_names:
+        ecu = odxdb.ecus[ecu_name]
+        if not ecu:
+            print(f"The ecu variant '{ecu_name}' could not be found!")
+            continue
+        if data:
+            found_services = ecu._find_services_for_uds(data)
+            for found_service in found_services:
+                ecu_names = services.get(found_service, [])
+                ecu_names.append(ecu_name)
+                services[found_service] = ecu_names
+
+    print(f"Binary data: {data.hex(' ')}")
+    for service, ecu_names in services.items():
+        if isinstance(service, DiagService):
+            print(
+                f"Decoded by service '{service.short_name}' (decoding ECUs: {', '.join(ecu_names)})"
+            )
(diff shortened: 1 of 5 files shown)


odxtools/cli/decode.py
```diff
@@ -0,0 +1,128 @@
...
+
+def add_subparser(subparsers: "argparse._SubParsersAction") -> None:
+    parser = subparsers.add_parser(
+        "decode",
+        description="\n".join([
+            "Decode request by hex-data",
+            "",
+            "Examples:",
+            "  For displaying the service associated with the request 10 01 & decoding it:",
+            "    odxtools decode ./path/to/database.pdx -d 10 01",
```
andlaus, line 77:
```suggestion
            "    odxtools decode ./path/to/database.pdx -d '10 01'",
```

odxtools/cli/decode.py
```diff
@@ -0,0 +1,128 @@
...
+            "  For more information use:",
+            "    odxtools decode -h",
+        ]),
+        help="Find & print services by hex-data, or name. Can also decode the request.",
```
andlaus, line 81: that seems to be not correct after this PR anymore?

tests/test_cli.py
```diff
@@ -68,14 +77,23 @@ def test_list_tool(self) -> None:
...
-        UtilFunctions.run_find_tool(data="3E 00")
-        UtilFunctions.run_find_tool(data="3E 00", ecu_variants=["somersault_lazy"])
-        UtilFunctions.run_find_tool(data="3E 00", decode=True)
-        UtilFunctions.run_find_tool(data="3E 00", decode=True, relaxed_output=True)
+    def test_find_tool(self) -> None:
+        UtilFunctions.run_find_tool(service_names=["headstand"])
+        UtilFunctions.run_find_tool(service_names=["headstand"], allow_unknown_bit_lengths=True)
+        UtilFunctions.run_find_tool(
+            service_names=["headstand"], allow_unknown_bit_lengths=True, no_details=True)
+        # UtilFunctions.run_find_tool(service_names="3e00")
```
andlaus, line 93: avoid commented-out code if there's not a very good reason for it...

odxtools/cli/decode.py
floroks: oops, yes, thanks, i fixed it to work again

tests/test_cli.py
floroks: true, I removed it

odxtools/cli/decode.py
floroks: oops, yes, thanks, i fixed it to work with -d 10 01 again. -D is a modifier to additionally decode it, although maybe it might be better to just differentiate between -d 10 01 and -D 10 01 - thoughts?

odxtools/cli/decode.py
```diff
@@ -0,0 +1,130 @@
...
+        metavar="VARIANT",
+        required=False,
+        help="Specifies which ecu variants should be included.",
+        default="all",
+    )
+
+    parser.add_argument(
+        "-d",
+        "--data",
+        nargs="*",
```
andlaus, line 101: I'd vastly prefer to make this a simple string, because right now specifying positional arguments after -D makes things quite iffy (e.g. `odxtools decode -D 01 01 my_file.pdx` will fail while `odxtools decode -D "01 01" my_file.pdx` will work just fine).
```suggestion
```

odxtools/cli/decode.py
andlaus: allowing the data to be specified as multiple arguments is a bad idea IMO, see below...

odxtools/cli/decode.py
floroks: changed it

andlaus:
okay, looks good to me now. A small thing which I've noticed while playing with it: if the data is specified using an illegal format (e.g. `-d "3e:00"`), `odxtools decode` only bails out after it has parsed the PDX file, which can take a few seconds. It would probably be good to reverse this order because parsing the data is basically instantaneous...

floroks:
fixed, I also remove all non-hex-chars from the string, to be more fault-tolerant to the input

andlaus:
ok, thanks! merging.

PR #228, opened 2023-10-21
Fix an issue, in which overlapping parameters were shown separately

floroks:
Fix an issue, in which overlapping parameters were shown separately
Overlapping parameters happen when you define parameters with different bit-lengths at the same byte-position.

Example:
Define parameter X at position 3 with bit-length 24
Define parameter Y at position 3 with bit-length 8
Define parameter Z at position 4 with bit-length 8
Define parameter A at position 5 with bit-length 8

### DIFF
--- a/odxtools/basicstructure.py
+++ b/odxtools/basicstructure.py
@@ -287,10 +287,13 @@ class BasicStructure(DopBase):
         # sort parameters
         sorted_params: list = list(self.parameters)  # copy list
 
-        def param_sort_key(param: Parameter) -> Tuple[int, int]:
+        def param_sort_key(param: Parameter) -> Tuple[int, int, int]:
             byte_position_int = param.byte_position if param.byte_position is not None else 0
             bit_position_int = param.bit_position if param.bit_position is not None else 0
-            return (byte_position_int, 8 - bit_position_int)
+            param_bit_length_int = param.get_static_bit_length() or 0
+            # take bit-length into account for overlapping parameters. 1000000 is an arbitrary number chosen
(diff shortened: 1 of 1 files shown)


odxtools/basicstructure.py
```diff
@@ -287,10 +287,13 @@ def _message_format_lines(self, allow_unknown_lengths: bool = False) -> List[str
         # sort parameters
         sorted_params: list = list(self.parameters)  # copy list
 
-        def param_sort_key(param: Parameter) -> Tuple[int, int]:
+        def param_sort_key(param: Parameter) -> Tuple[int, int, int]:
```
kayoub5, line 290: @andlaus why are parameters being sorted here? does the spec indicate that as a pre-condition?

odxtools/basicstructure.py
```diff
@@ -411,6 +422,17 @@ def param_sort_key(param: Parameter) -> Tuple[int, int]:
             formatted_lines.append(next_line)
             next_line = ""
 
+            if 0 < param_idx < len(params) - 1 and hasattr(
```
andlaus, line 425: I prefer range checks to be oldschool:
```suggestion
            if 0 < param_idx and param_idx < len(params) - 1 and hasattr(
```

(while `a < x < b`  works with python it feels too magical to me because `(a < x) < b` will *not* work in all cases)...

floroks: fixed

andlaus:
ok, thanks; merging.

(note to myself: `BasicStructure._message_format_lines()` is *way* too long, too complicated and should probably not be part of the class in the first place.)

PR #231, opened 2023-10-27
add DiagComm as a base class for all diagnostic communication objects

andlaus:
diagnostic communication objects are diagnostic services and single ecu jobs.

besides bringing the class hierarchy in line with the ODX specification, this PR also properly internalizes all sub-elements and attributes of DIAG-COMM, DIAG-SERVICE and SINGLE-ECU-JOB. (i.e., all diag comm objects should now be complete)


odxtools/diagservice.py
```diff
@@ -1,50 +1,66 @@
...
     neg_response_refs: List[OdxLinkRef]
-    admin_data: Optional[AdminData]
-    semantic: Optional[str]
-    audience: Optional[Audience]
-    functional_class_refs: Iterable[OdxLinkRef]
-    pre_condition_state_refs: Iterable[OdxLinkRef]
-    state_transition_refs: Iterable[OdxLinkRef]
-    sdgs: List[SpecialDataGroup]
+
+    #pos_response_suppressable: Optional[PosResponseSuppressable]
```
kayoub5, line 47: Commented out code?

andlaus: this is meant as a nagging TODO item for anyone who stumbles over this. I prefixed it with "TODO": f5397ea7297731d89757e62c8c75705ed1fc25f8, thanks for the catch!

tests/test_ecu_variant_matching.py
```diff
@@ -105,6 +117,17 @@ def supplier_service(monkeypatch: pytest.MonkeyPatch, dummy_response: Response)
         description=None,
         semantic=None,
         admin_data=None,
+        protocol_snrefs=[],
+        related_diag_comm_refs_raw=[],
+        diagnostic_class=None,
+        is_mandatory_raw=None,
+        is_executable_raw=None,
+        is_final_raw=None,
+        comparam_refs=NamedItemList[CommunicationParameterRef](),
```
kayoub5, line 126: refs as named list?

odxtools/cli/decode.py
```diff
@@ -30,7 +30,8 @@ def print_summary(
     decode: bool = False,
 ) -> None:
     ecu_names = ecu_variants if ecu_variants else [ecu.short_name for ecu in odxdb.ecus]
-    services: Dict[DiagService, List[str]] = {}
+    service_db: Dict[str, DiagService] = {}
```
kayoub5, line 33: why not use a named list?

odxtools/diagcomm.py
```diff
@@ -0,0 +1,241 @@
...
+    from .diaglayer import DiagLayer
+
+
+class DiagClassType(Enum):
+    STARTCOMM = "STARTCOMM"
+    STOPCOMM = "STOPCOMM"
+    VARIANTIDENTIFICATION = "VARIANTIDENTIFICATION"
+    READ_DYN_DEFINED_MESSAGE = "READ-DYN-DEFINED-MESSAGE"
+    DYN_DEF_MESSAGE = "DYN-DEF-MESSAGE"
+    CLEAR_DYN_DEF_MESSAGE = "CLEAR-DYN-DEF-MESSAGE"
```
kayoub5, line 31: from spec?

odxtools/diagcomm.py
```diff
@@ -0,0 +1,241 @@
...
+
+
+@dataclass
+class RelatedDiagCommRef:
+    diag_comm_ref: OdxLinkRef
```
kayoub5, line 36: why not sub-class OdxLinkRef?

odxtools/diagcomm.py
```diff
@@ -0,0 +1,241 @@
...
+    def protocols(self) -> NamedItemList["DiagLayer"]:
+        return self._protocols
+
+    @property
+    def related_diag_comm_refs(self) -> NamedItemList[RelatedDiagCommRef]:
+        return self._related_diag_comm_refs
+
+    @property
+    def pre_condition_states(self) -> NamedItemList[State]:
+        return self._pre_condition_states
```
kayoub5, line 171: why the property if there is no logic?

odxtools/diaglayer.py
```diff
@@ -298,7 +299,7 @@ def ecu_variant_patterns(self) -> List[EcuVariantPattern]:
     # <stuff subject to value inheritance>
     #######
     @property
-    def diag_comms(self) -> NamedItemList[Union[DiagService, SingleEcuJob]]:
+    def diag_comms(self) -> NamedItemList[DiagComm]:
```
kayoub5, line 302: breaking change?

tests/test_decoding.py
```diff
@@ -262,7 +285,18 @@ def test_decode_request_coded_const(self) -> None:
...
+            diagnostic_class=None,
+            is_mandatory_raw=None,
+            is_executable_raw=None,
+            is_final_raw=None,
             semantic=None,
+            comparam_refs=NamedItemList[CommunicationParameterRef](),
+            is_cyclic_raw=None,
+            is_multiple_raw=None,
+            addressing_raw=None,
+            transmission_mode_raw=None,
```
kayoub5, line 299: the number of arguments that needs to be passed is getting ridiculous

tests/test_ecu_variant_matching.py
andlaus: In contrast to the XSD, `CommunicationParameterRef` is not derived from `OdxLinkRef`. This mainly because it contains plenty of additional information, such as the value for the respective communication parameter but also because `OdxLinkRef` is currently a frozen dataclass in order for it to be hashable. (its `short_name` property currently is a forward to the short name of the referenced comparam spec: https://github.com/mercedes-benz/odxtools/blob/main/odxtools/communicationparameterref.py#L135 ) I don't have an elegant solution here, and I'm not even sure if one exists...

odxtools/diagcomm.py
andlaus: mainly because `OdxLinkRef` is currently frozen. That said, references do not really need to be hashable, so maybe this should be changed. Any preference?

odxtools/diagcomm.py
andlaus: to make this attribute unsettable (by normal means)

odxtools/diaglayer.py
andlaus: not really: The only two concrete subclasses of DIAG-COMM are DIAG-SERVICE and SINGLE-ECU-JOB...

tests/test_decoding.py
andlaus: true. maybe we should instantiate these classes using their `.from_et()` constructors...

odxtools/cli/decode.py
andlaus: for the `service_db` this would indeed be possible, albeit it would change the entry creation code and I cannot see how item access could benefit either. Also, the ` service_ecus` variable cannot easily be changed, because there is no easy way of associating a name to an entry (which is a list of ecu names that feature that service).

The reason why this was necessary in the first place is that I figured that it is not a terribly good idea for `DiagService` objects to be hashable...

odxtools/diagcomm.py
kayoub5: why? most attributes in other classes are settable

odxtools/diagcomm.py
andlaus: yes. This class was just moved over from `singleecujob.py` (`DiagService` would have needed it, too, but it did not internalize the `DIAGNOSTIC-CLASS` sub-element...)

odxtools/diagcomm.py
andlaus: only the parts that can be defined using the XML. computed stuff like resolved references is not supposed to be user settable. (there are plenty examples: [1](https://github.com/mercedes-benz/odxtools/blob/main/odxtools/tablerow.py#L140), [2](https://github.com/mercedes-benz/odxtools/blob/main/odxtools/diaglayer.py#L350), [3](https://github.com/mercedes-benz/odxtools/blob/main/odxtools/odxtools/database.py#L147), ...)

odxtools/diagcomm.py
andlaus: okay, I "unfroze" `OdxLinkRef`: 1d1b2012fbb6fc2ca9454e01e7492b1a97d7e242

tests/test_ecu_variant_matching.py
kayoub5: My suggestion in this case, is to use a different suffix, for example `Reference` instead of `Ref`, so that the two types don't get mixed up

andlaus: Okay: I have refactored the comparam classes, but it seems like the comparam stuff in the ODX spec is quite quirky, so I decided to deviate from the ODX specification's structure a bit: I renamed the "*Comparam" to "*ComparamSpec" and "CommunicationParameterRef" to "Comparam", moved the physical default value for complex comparam specs into the class itself instead of keeping it weirdly external as mandated by ODX, plus  a few other things.

since this a rather big endeavor, I'd like to do this in a later pull request...

kayoub5 (approved):

Issue #236, opened 2023-11-07
ComplexComparam `from_et()` does not behave properly and skips over subparams


willzhang05:
I noticed the recent PR https://github.com/mercedes-benz/odxtools/pull/233 introduced a regression in how the `ComplexComparam` class is initialized in the `from_et()` function, causing the generated XML for ComplexComparams to be incorrect.

I have a couple findings:

1. I was noticing that in my output XML, sub-parameters for `COMPLEX-COMPARAM` tags were missing. The source XML that was being parsed came from Vector software, and not all of the sub-parameters have associated `COMPLEX-PHYSICAL-DEFAULT-VALUE` tags.

   I wasn't quite clear on the meaning of the comment in the code: https://github.com/mercedes-benz/odxtools/blob/6865fca3b059fabf8a0672fe129d5280ed337573/odxtools/complexcomparam.py#L64-L70 so I took a look at the ISO 22901-1 (2008) document. On page 351 of the ISO document, the XML schema suggests to me that the `COMPLEX-PHYSICAL-DEFAULT-VALUE` tag is not required to be there (minimum occurrence of 0):

   ```xml
   <xsd:element maxOccurs="1" minOccurs="0" type="COMPLEX-PHYSICAL-DEFAULT-VALUE" name="COMPLEX-
   PHYSICAL-DEFAULT-VALUE"/>
   ```
   so based on this, this seems like it might be wrong:
https://github.com/mercedes-benz/odxtools/blob/6865fca3b059fabf8a0672fe129d5280ed337573/odxtools/complexcomparam.py#L71-L80 

   If this is the last sub-parameter, the loop will break without appending the sub-parameter to the `subparams` list. 

   In addition, the conditional may be inverted, since if it isn't a complex physical default value, it calls `create_complex_value_from_et()` and then continues the loop without appending the sub-parameter to the `subparams` list.

   Both of these issues cause the generated XML to be missing the sub-parameters inside `COMPLEX-COMPARAM` tags.

2. The use of iterators doesn't seem to result in desired behavior:
https://github.com/mercedes-benz/odxtools/blob/6865fca3b059fabf8a0672fe129d5280ed337573/odxtools/complexcomparam.py#L71 
[...]

PR #237, opened 2023-11-07
fix ComplexComparam `from_et` function

willzhang05:
Addresses https://github.com/mercedes-benz/odxtools/issues/236

### DIFF
--- a/odxtools/complexcomparam.py
+++ b/odxtools/complexcomparam.py
@@ -49,18 +49,14 @@ class ComplexComparam(BaseComparam):
         # default values. Due to the quirky way this is defined by the
         # ODX specification, this is a *major* pain in the butt!
         subparams: NamedItemList[BaseComparam] = NamedItemList()
-        elem_it = iter(et_element)
-        while True:
-            try:
-                elem = next(elem_it)
-            except StopIteration:
-                break
-
-            if elem.tag not in ("COMPARAM", "COMPLEX-COMPARAM"):
+        elems = list(et_element)
+        i = 0
+        while i < len(elems):
+            if elems[i].tag not in ("COMPARAM", "COMPLEX-COMPARAM"):
+                i += 1
                 continue
 
-            subparam = create_any_comparam_from_et(elem, doc_frags)
-
+            subparam = create_any_comparam_from_et(elems[i], doc_frags)
             # the next element in the list *may* hold the physical
             # default value for the sub-parameter. if it is not the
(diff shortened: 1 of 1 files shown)


odxtools/complexcomparam.py
```diff
@@ -49,37 +49,30 @@ def from_et(et_element: ElementTree.Element,
         # default values. Due to the quirky way this is defined by the
         # ODX specification, this is a *major* pain in the butt!
         subparams: NamedItemList[BaseComparam] = NamedItemList()
-        elem_it = iter(et_element)
+        elems = list(iter(et_element))
```
andlaus, line 52: I believe the `iter()` is unnecessary here:
```suggestion
        elems = list(et_element)
````

File: odxtools/complexcomparam.py [hunk 2]
```diff
@@ -49,37 +49,30 @@ def from_et(et_element: ElementTree.Element,
...
+        i = 0
         while True:
-            try:
-                elem = next(elem_it)
-            except StopIteration:
+            if i >= len(elems):
```
andlaus, line 55: IMO it is better to use a while with a condition here
```python
        while i < len(elems):
```

andlaus (commented):
Thanks for the patch. besides my usual nitpicks it looks good!

Before it can be merged, you need to sign the [CLA](https://github.com/mercedes-benz/odxtools/blob/main/CONTRIBUTORS_LICENSE_AGREEMENT.md). (please put me into the CC of the email: andreas.lauser@mercedes-benz.com)

odxtools/complexcomparam.py
```diff
@@ -49,37 +49,30 @@ def from_et(et_element: ElementTree.Element,
...
-                elem = next(elem_it)
-            except StopIteration:
-                break
-
-            if elem.tag != "COMPLEX-PHYSICAL-DEFAULT-VALUE":
-                subparam.physical_default_value = create_complex_value_from_et(elem)
-                elem_it = tmp_it
-                continue
+            if i + 1 < len(elems):
+                if elems[i + 1].tag == "COMPLEX-PHYSICAL-DEFAULT-VALUE":
```
andlaus, line 70:
```suggestion
            if i + 1 < len(elems) and elems[i + 1].tag == "COMPLEX-PHYSICAL-DEFAULT-VALUE":
```

andlaus:
okay, I've received the CLA. Let's get this merged. Thanks for the contribution :)

PR #240, opened 2023-11-17
clean up the initialization code for DOPs

andlaus:
after this, everything DOP-BASE related is initialized where it is supposed to be and all its attributes and subtags should be handled for every tag extending it. (except the OID attribute...)


odxtools/dopbase.py
```diff
@@ -20,18 +24,21 @@ class DopBase(IdentifiableElement):
     Any class that a parameter can reference via a DOP-REF should inherit from this class.
     """
 
-    is_visible_raw: Optional[bool]
```
kayoub5, line 23: Why was `is_visible_raw` removed?

examples/somersaultecu.py
```diff
@@ -1301,8 +1302,8 @@ class SomersaultSID(IntEnum):
             short_name="start_session",
             long_name="Start the diagnostic session & do some mischief",
             description=None,
+            admin_data=None,
             sdgs=[],
-            is_visible_raw=None,
```
kayoub5, line 1305: Are you sure about this?

odxtools/dopbase.py
andlaus: because DOP-BASE does not feature such an attribute...

kayoub5 (approved):

PR #241, opened 2023-11-17
added IsoTpConnection

hacxo:
This class provides an async IsoTpConnection that supports CAN-FD.

Christian Hackenbeck &lt;christian.hackenbeck@mercedes-benz.com&gt;, on behalf of [MBition GmbH](https://mbition.io/).

### DIFF
--- a/odxtools/isotp_connection.py
+++ b/odxtools/isotp_connection.py
@@ -1,0 +1,95 @@
+# SPDX-License-Identifier: MIT
+
+import asyncio
+from typing import Optional
+
+import bitstruct
+import can
+
+from .isotp_state_machine import IsoTp, IsoTpStateMachine
+
+
+class IsoTpConnection(IsoTpStateMachine):
+    """
+    This class extends the IsoTpStateMachine with the capability of
+    sending out CAN-FD messages and reading ISOTP messages.
+    """
+
+    def __init__(self,
+                 *,
+                 can_bus: can.BusABC,
+                 can_tx_id: int,
+                 can_rx_id: int,
+                 padding_value: int = 0xAA,
+                 is_fd: bool = False):
+        super().__init__(can_rx_ids=[can_rx_id])
+        self._can_bus = can_bus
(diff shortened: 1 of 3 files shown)


odxtools/isotp_connection.py
```diff
@@ -0,0 +1,95 @@
...
+        self._can_bus = can_bus
+        self._can_tx_id = can_tx_id
+        self._frame_size = 64 if is_fd else 8
+        self._padding_value = bytes([padding_value])
+        self._is_fd = is_fd
+
+    async def read_telegram(self) -> bytes:
+        async for x in self.read_telegrams(bus=self._can_bus):
+            return x[1]
+        raise Exception
```
kayoub5, line 35: Why are you raising here?

odxtools/isotp_connection.py
```diff
@@ -0,0 +1,95 @@
...
+
+    async def send_telegram(self, payload: bytes) -> None:
```
kayoub5, line 37: I would never recommend such re-implementation of the ISO-TP protocol, It does not account for block size, separation time, extended addressing, [CAN-FD extensions to ISO-TP](https://en.wikipedia.org/wiki/CAN_FD#CAN_&_CAN_FD_TP_Headers), etc.

It would be better to re-use existing implementations such as https://github.com/pylessard/python-udsoncan/blob/7f7448bd16f3ac78c6568f6202eb5ee449479692/udsoncan/connections.py#L237

Or https://github.com/pylessard/python-can-isotp

odxtools/isotp_connection.py
hacxo: MyPy wanted some return / raise.

odxtools/isotp_connection.py
hacxo: Thanks for the input! I agree that not all features are implemented in the solution. It was rather a quick-and-dirty solution. I will look into alternatives and close this pull request for now.

PR #242, opened 2023-11-20
Fix init of structures

andlaus:
This cleans up the initialization code of structures and some of its derived classes (in particular, `Response`). Note that requests and responses are *not* supposed to be structures, so there will likely be another PR which takes care of that soon. (This will probably be part of a refactoring of the de- and encoding machinery...)


odxtools/response.py
```diff
@@ -1,15 +1,42 @@
...
+from .odxlink import OdxDocFragment
 from .odxtypes import ParameterValue
 from .parameters.matchingrequestparameter import MatchingRequestParameter
+from .utils import dataclass_fields_asdict
 
 
+class ResponseType(Enum):
+    POS_RESPONSE = "POS-RESPONSE"
+    NEG_RESPONSE = "NEG-RESPONSE"
+    GLOBAL_NEG_RESPONSE = "GLOBAL-NEG-RESPONSE"
```
kayoub5, line 18: `_RESPONSE` suffix really needed for all?

PR #243, opened 2023-11-21
fix parsing and writing of COMPARAM-SPEC

andlaus:
It turns out that files containing "COMPARAM-SPEC" root elements were ignored for PDX files using ODX 2.2. (also, the writing code got slightly broken recently.)

I'm not 100% sure if this still works with ODX 2.0 files, though...

### DIFF
--- a/odxtools/database.py
+++ b/odxtools/database.py
@@ -9,6 +9,7 @@ from .comparamspec import ComparamSpec
 from .comparamsubset import ComparamSubset
 from .diaglayer import DiagLayer
 from .diaglayercontainer import DiagLayerContainer
+from .exceptions import odxraise
 from .nameditemlist import NamedItemList, short_name_as_key
 from .odxlink import OdxLinkDatabase
 
@@ -27,6 +28,7 @@ class Database:
                  *,
                  pdx_zip: Optional[ZipFile] = None,
                  odx_d_file_name: Optional[str] = None) -> None:
+        self.model_version = None
 
         if pdx_zip is None and odx_d_file_name is None:
             # create an empty database object
@@ -60,6 +62,11 @@ class Database:
         comparam_specs: List[ComparamSpec] = []
         for root in documents:
             # ODX spec version
+            model_version = version(root.attrib.get("MODEL-VERSION", "2.0"))
+            if self.model_version is not None and self.model_version != model_version:
+                odxraise(f"Different ODX versions used in the same file (ODX {model_version} "
+                         f"and ODX {self.model_version}")
+            self.model_version = model_version
             dlc = root.find("DIAG-LAYER-CONTAINER")
             if dlc is not None:
                 dlcs.append(DiagLayerContainer.from_et(dlc, []))
--- a/odxtools/database.py
+++ b/odxtools/database.py
@@ -81,7 +81,10 @@ class Database:
 
             cp_spec = root.find("COMPARAM-SPEC")
             if cp_spec is not None:
-                comparam_specs.append(ComparamSpec.from_et(cp_spec, []))
+                if model_version == "2.0":
+                    comparam_subsets.append(ComparamSubset.from_et(cp_spec, []))
+                else:  # odx >= 2.2
+                    comparam_specs.append(ComparamSpec.from_et(cp_spec, []))
 
         self._diag_layer_containers = NamedItemList(dlcs)
         self._diag_layer_containers.sort(key=short_name_as_key)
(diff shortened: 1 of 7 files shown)


odxtools/database.py
```diff
@@ -44,10 +43,11 @@ def __init__(self,
             names = list(pdx_zip.namelist())
             names.sort()
             for zip_member in names:
-                # file name can end with .odx, .odx-d, .odx-c, .odx-cs, .odx-e, .odx-f, .odx-fd, .odx-m, .odx-v
-                # We could test for all that, or just make sure suffix starts with .odx
+                # file name can end with .odx, .odx-d, .odx-c,
+                # .odx-cs, .odx-e, .odx-f, .odx-fd, .odx-m, .odx-v We
+                # could test for all that, or just make sure suffix
```
kayoub5, line 48:
```suggestion
                # We could test for all that, or just make sure suffix
```

odxtools/database.py
```diff
@@ -44,10 +43,11 @@ def __init__(self,
             names = list(pdx_zip.namelist())
             names.sort()
             for zip_member in names:
-                # file name can end with .odx, .odx-d, .odx-c, .odx-cs, .odx-e, .odx-f, .odx-fd, .odx-m, .odx-v
-                # We could test for all that, or just make sure suffix starts with .odx
+                # file name can end with .odx, .odx-d, .odx-c,
+                # .odx-cs, .odx-e, .odx-f, .odx-fd, .odx-m, .odx-v We
```
kayoub5, line 47:
```suggestion
                # .odx-cs, .odx-e, .odx-f, .odx-fd, .odx-m, .odx-v
```

odxtools/database.py
```diff
@@ -60,22 +60,21 @@ def __init__(self,
...
             # ODX spec version
-            model_version = version(root.attrib.get("MODEL-VERSION", "2.0"))
             dlc = root.find("DIAG-LAYER-CONTAINER")
             if dlc is not None:
                 dlcs.append(DiagLayerContainer.from_et(dlc, []))
-            # In ODX 2.0 there was only COMPARAM-SPEC
-            # In ODX 2.2 content of COMPARAM-SPEC was renamed to COMPARAM-SUBSET
-            # and COMPARAM-SPEC becomes a container for PROT-STACKS
-            # and a PROT-STACK references a list of COMPARAM-SUBSET
-            if model_version >= version("2.2"):
```
kayoub5, line 71: The test on the model version is required, otherwise the wrong parser will be used

andlaus: why? since COMPARAM-SUBSET is not present in ODX 2.0, the first `if` branch should never be taken?

(maybe COMPARAM-SPEC must be made aware of ODX 2.0, though...)

kayoub5: > maybe COMPARAM-SPEC must be made aware of ODX 2.0, though...

Exactly.

andlaus: ok, fixed: 4985121279460073e65d8c931f05d890f75540ad

(Leaving that `if` alone would not have changed anything, though...)

odxtools/database.py
```diff
@@ -61,21 +63,25 @@ def __init__(self,
...
+            # content of COMPARAM-SPEC was moved to COMPARAM-SUBSET
+            # and COMPARAM-SPEC became a container for PROT-STACKS and
+            # a PROT-STACK references a list of COMPARAM-SUBSET
+            cp_subset = root.find("COMPARAM-SUBSET")
+            if cp_subset is not None:
+                comparam_subsets.append(ComparamSubset.from_et(cp_subset, []))
+
+            cp_spec = root.find("COMPARAM-SPEC")
+            if cp_spec is not None:
+                comparam_specs.append(ComparamSpec.from_et(cp_spec, []))
```
kayoub5, line 84: This may look strange at first, but between 2.0 and 2.2 only the XML tag changed, the content of the COMPARAM-SPEC in 2.0 is basically a COMPARAM-SUBSET in 2.2
This way the code needed for 2.0 support remains minimal
```suggestion
                comparam_specs.append(ComparamSubset.from_et(cp_spec, []))
```

andlaus: ah, I see. it's better to append these objects to `comparam_subsets` then, though? (This would also make conversion of datasets from odx 2.0 to 2.2 easier.) I've implemented it this way for now: adf3dfb946eb35519f866282df0f90b369bd0c97

kayoub5 (approved):

PR #244, opened 2023-11-22
Add some convenience for communication parameters

andlaus:
This adds a few communication parameter related convenience functions (`get_can_mtu()`, `use_can_fd()`, `get_can_baudrate()` and `get_can_fd_baudrate()`) and it allows to specify the protocol to be used by either passing the name of the protocol as a string or using the whole protocol diaglayer. The latter is useful to specify the protocol like `ecu.get_can_receive_address(ecu.protocols.OBD_CAN)` instead of `ecu.get_can_receive_address(ecu.protocols.OBD_CAN.short_name)`.

### DIFF
--- a/odxtools/diaglayer.py
+++ b/odxtools/diaglayer.py
@@ -1,4 +1,5 @@
 # SPDX-License-Identifier: MIT
+import re
 import warnings
 from dataclasses import dataclass
 from functools import cached_property
@@ -635,6 +636,68 @@ class DiagLayer:
             return None
 
         return cps[0]
+
+    def get_can_mtu(self, protocol_name: Optional[str] = None) -> Optional[int]:
+        """Return the maximum size of CAN frames that can be
+        transmitted in bytes.
+
+        For classic CAN busses, this is basically always 8. CAN-FD can
+        send up to 64 bytes per frame.
+        """
+        com_param = self.get_comparam("CP_CANFDTxMaxDataLength", protocol_name=protocol_name)
+        if com_param is None:
+            rx_id = self.get_can_receive_id(protocol_name=protocol_name)
+            if rx_id is not None:
+                # bus is CAN. We also use 8 bytes for CAN-FD if this
+                # parameter was not specified.
+                return 8
+
+            # bus is not CAN
+            return None
+
+        val = com_param.value
+        if not isinstance(val, str):
+            return None
+
+        m = re.search("TX_DL *= *([0-9]*)", val)
+        if m:
+            return int(m.group(1))
+
+        # unexpected format of parameter value
+        return 8
+
+    def use_can_fd(self, protocol_name: Optional[str] = None) -> bool:
+        return (self.get_can_mtu(protocol_name=protocol_name) or 0) > 8
+
+    def get_can_baudrate(self, protocol_name: Optional[str] = None) -> Optional[int]:
+        """Baudrate of the CAN bus which is used by the ECU"""
+        com_param = self.get_comparam("CP_Baudrate", protocol_name=protocol_name)
+        if com_param is None:
+            return None
+
+        val = com_param.value
+        if not isinstance(val, str):
+            return None
+
+        return int(val)
+
+    def get_can_fd_baudrate(self, protocol_name: Optional[str] = None) -> Optional[int]:
+        """Data baudrate of the CAN bus which is used by the ECU
+
+        If the ECU is not using CAN-FD, None is returned.
+        """
+        if not self.use_can_fd(protocol_name=protocol_name):
+            return None
+
+        com_param = self.get_comparam("CP_CANFDBaudrate", protocol_name=protocol_name)
+        if com_param is None:
+            return None
+
+        val = com_param.value
+        if not isinstance(val, str):
+            return None
+
+        return int(val)
 
     def get_can_receive_id(self, protocol_name: Optional[str] = None) -> Optional[int]:
         """CAN ID to which the ECU listens for diagnostic messages"""
--- a/odxtools/diaglayer.py
+++ b/odxtools/diaglayer.py
@@ -614,11 +614,17 @@ class DiagLayer:
         self,
         cp_short_name: str,
         *,
-        protocol_name: Optional[str] = None,
+        protocol: Optional[Union[str, "DiagLayer"]] = None,
     ) -> Optional[ComparamInstance]:
         """Find a specific communication parameter according to some criteria.
 
         Setting a given parameter to `None` means "don't care"."""
+
+        protocol_name: Optional[str]
+        if isinstance(protocol, DiagLayer):
+            protocol_name = protocol.short_name
+        else:
+            protocol_name = protocol
 
         # determine the set of applicable communication parameters
         cps = [cp for cp in self.comparams if cp.short_name == cp_short_name]
@@ -637,16 +643,16 @@ class DiagLayer:
 
         return cps[0]
 
-    def get_can_mtu(self, protocol_name: Optional[str] = None) -> Optional[int]:
+    def get_can_mtu(self, protocol: Optional[Union[str, "DiagLayer"]] = None) -> Optional[int]:
         """Return the maximum size of CAN frames that can be
         transmitted in bytes.
 
         For classic CAN busses, this is basically always 8. CAN-FD can
         send up to 64 bytes per frame.
         """
-        com_param = self.get_comparam("CP_CANFDTxMaxDataLength", protocol_name=protocol_name)
+        com_param = self.get_comparam("CP_CANFDTxMaxDataLength", protocol=protocol)
         if com_param is None:
-            rx_id = self.get_can_receive_id(protocol_name=protocol_name)
+            rx_id = self.get_can_receive_id(protocol=protocol)
             if rx_id is not None:
                 # bus is CAN. We also use 8 bytes for CAN-FD if this
                 # parameter was not specified.
@@ -666,12 +672,12 @@ class DiagLayer:
         # unexpected format of parameter value
         return 8
 
-    def use_can_fd(self, protocol_name: Optional[str] = None) -> bool:
-        return (self.get_can_mtu(protocol_name=protocol_name) or 0) > 8
+    def use_can_fd(self, protocol: Optional[Union[str, "DiagLayer"]] = None) -> bool:
+        return (self.get_can_mtu(protocol=protocol) or 0) > 8
 
-    def get_can_baudrate(self, protocol_name: Optional[str] = None) -> Optional[int]:
+    def get_can_baudrate(self, protocol: Optional[Union[str, "DiagLayer"]] = None) -> Optional[int]:
         """Baudrate of the CAN bus which is used by the ECU"""
-        com_param = self.get_comparam("CP_Baudrate", protocol_name=protocol_name)
+        com_param = self.get_comparam("CP_Baudrate", protocol=protocol)
         if com_param is None:
             return None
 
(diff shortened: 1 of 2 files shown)


odxtools/diaglayer.py
```diff
@@ -636,9 +643,73 @@ def get_comparam(
...
+        if m:
+            return int(m.group(1))
+
+        # unexpected format of parameter value
+        return 8
+
+    def use_can_fd(self, protocol: Optional[Union[str, "DiagLayer"]] = None) -> bool:
+        return (self.get_can_mtu(protocol=protocol) or 0) > 8
+
+    def get_can_baudrate(self, protocol: Optional[Union[str, "DiagLayer"]] = None) -> Optional[int]:
```
kayoub5, line 678: You are putting too much logic into a class that's already too big, maybe move those to helper class? `diag_layer.com.get_can_baudrate()`

odxtools/diaglayer.py
```diff
@@ -827,7 +906,9 @@ def get_doip_routing_activation_type(self,
 
         return int(result)
 
-    def get_tester_present_time(self, protocol_name: Optional[str] = None) -> Optional[float]:
+    def get_tester_present_time(self,
+                                protocol: Optional[Union[str,
+                                                         "DiagLayer"]] = None) -> Optional[float]:
```
kayoub5, line 911: The more I look at the current API, the more I think something like {{diag_layer.com(protocol).tester_present_time}} would make more sense, where `com` is a function that returns an object that encapsulate all the CP lookup logic.

andlaus: such an approach makes implementing fallbacks more complicated (i.e., "find a COMPARAM for a given protocol, if none exits, look up for one which does not specify a protocol"). It might be a good idea to put these methods into a separate "comparam lookup convenience" class, but IMO that's a topic for a separate PR...

odxtools/diaglayer.py
kayoub5: Python allows for multiple inheritance, maybe make use of that to split the large file into smaller ones?

kayoub5: My main concern is that `odxtools/diaglayer.py` have +1000 lines of code, we barely scratched the surface of `convenience for communication parameters`, there are like dozens of those com parameters in the standard

andlaus:
@kayoub5: Do you agree to merging this with the understanding that the comparam convenience code should be improved/refactored in the future is acceptable?

odxtools/diaglayer.py
```diff
@@ -636,9 +643,73 @@ def get_comparam(
...
+        For classic CAN busses, this is basically always 8. CAN-FD can
+        send up to 64 bytes per frame.
+        """
+        com_param = self.get_comparam("CP_CANFDTxMaxDataLength", protocol=protocol)
+        if com_param is None:
+            rx_id = self.get_can_receive_id(protocol=protocol)
+            if rx_id is not None:
+                # bus is CAN. We also use 8 bytes for CAN-FD if this
+                # parameter was not specified.
+                return 8
```
kayoub5, line 659: This is misleading, if the PDX did not explicitly specify an MTU, no MTU should be inferred from other properties

odxtools/diaglayer.py
```diff
@@ -636,9 +643,73 @@ def get_comparam(
...
+            return None
+
+        m = re.search("TX_DL *= *([0-9]*)", val)
+        if m:
+            return int(m.group(1))
+
+        # unexpected format of parameter value
+        return 8
+
+    def use_can_fd(self, protocol: Optional[Union[str, "DiagLayer"]] = None) -> bool:
```
kayoub5, line 675: Maybe return None if inconclusive?
```suggestion
    def use_can_fd(self, protocol: Optional[Union[str, "DiagLayer"]] = None) -> Optional[bool]:
```

odxtools/diaglayer.py
```diff
@@ -636,9 +643,73 @@ def get_comparam(
...
+        if com_param is None:
+            return None
+
+        val = com_param.value
+        if not isinstance(val, str):
+            return None
+
+        return int(val)
+
+    def get_can_fd_baudrate(self,
```
kayoub5, line 690: Mostly duplicates `get_can_baudrate`, maybe add a `_get_baudrate(protocol, param_name)` helper?

andlaus: this is the signalling rate for the payload portion of CAN-FD frames. You still need the CAN baudrate for the rest of the frame. (i.e., CAN-FD busses exhibit two bitrates. this is to enable to continue to use non-FD devices on FD busses...)

kayoub5: I am proposing this
```python
def get_can_baudrate(self, ....):
    self._get_baudrate(self, "CP_Baudrate", ...)

def get_can_fd_baudrate(self, ....):
    self._get_baudrate(self, "CP_CANFDBaudrate", ...)
```

odxtools/diaglayer.py
```diff
@@ -636,9 +643,73 @@ def get_comparam(
...
+        """Baudrate of the CAN bus which is used by the ECU"""
+        com_param = self.get_comparam("CP_Baudrate", protocol=protocol)
+        if com_param is None:
+            return None
+
+        val = com_param.value
+        if not isinstance(val, str):
+            return None
+
+        return int(val)
```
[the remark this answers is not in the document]
kayoub5: Or is that not allowed by spec?

odxtools/diaglayer.py
```diff
@@ -636,9 +643,73 @@ def get_comparam(
...
+            return int(m.group(1))
+
+        # unexpected format of parameter value
+        return 8
+
+    def use_can_fd(self, protocol: Optional[Union[str, "DiagLayer"]] = None) -> bool:
+        return (self.get_can_mtu(protocol=protocol) or 0) > 8
+
+    def get_can_baudrate(self, protocol: Optional[Union[str, "DiagLayer"]] = None) -> Optional[int]:
+        """Baudrate of the CAN bus which is used by the ECU"""
```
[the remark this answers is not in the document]
andlaus: what's the unit of baud? (probably 1/s, but I'm not sure if this is very helpful...)

odxtools/diaglayer.py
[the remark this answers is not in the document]
andlaus: AFAIK, comparams are not exhaustively covered by the spec. for the ECUs which I have access to, the comparam specifies this to be an UINT32:
```
>>> ecu.comparams.CP_Baudrate.spec.dop_ref
OdxLinkRef(ref_id='IDENTICAL_UINT32_Baud', ref_docs=[OdxDocFragment(doc_name='ISO_11898_2_DWCAN', doc_type='COMPARAM-SUBSET')])
```

odxtools/diaglayer.py
andlaus: what would be the difference from the user's POV? maybe it is better to call this method `get_can_fd_databaudrate()`, though?

andlaus: (also, I do not believe that this would make things easier to understand or even save a significant number of lines of code...)

andlaus: a more generic approach would be
```python
def get_simple_comparam_value(self, comparam_name, conversion_fn)
```
so that most of these convenience function would reduce to
```python
def get_can_baudrate(self):
    return self.get_simple_comparam_value("CP_Baudrate", int)
```

(not sure if this a good idea either...)

odxtools/diaglayer.py
andlaus: what does `None` mean in this context? "I don't know"? I'd expect to go for classical CAN if CAN FD is not specified explicitly. The thing I'm not 100% sure is if other manufacturers specify using CAN FD the same way as Mercedes does. (I assume that they are, mainly because the number of available commercial tools is rather limited AFAICS. If there are other ways to specify CAN FD in the wild: patches welcome ;))

odxtools/diaglayer.py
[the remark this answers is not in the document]
andlaus: done: 6f59aa68d428cb0b70c0970ca1c0ec91096b3494

odxtools/diaglayer.py
kayoub5: `None` always means "Missing" or "Undefined", that's the default fallback when we don't know what the value is.

odxtools/diaglayer.py
kayoub5: You can keep the `get_simple_comparam_value` private, that way you can change it if you want later

andlaus: I tried implementing this approach: It didn't work because I haven't found a good way to make it play nicely with type checking: The issue is that the return type of that function would be the return type of the callable passed as the second parameter...

odxtools/diaglayer.py
```diff
@@ -636,9 +643,73 @@ def get_comparam(
...
+                return 8
+
+            # bus is not CAN
+            return None
+
+        val = com_param.value
+        if not isinstance(val, str):
+            return None
+
+        m = re.search("TX_DL *= *([0-9]*)", val)
```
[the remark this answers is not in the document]
andlaus: hm, okay. that looks similar to my files, except that some of mine do not seem to specify the part after the comma, which makes determining whether whether to use CAN-FD even more awkward.

question: does the file which sets `CP_CANFDTxMaxDataLength` to `TX_DL = 8, CANFD` also specify a data bit rate (the `CP_CANFDBaudrate` comparam)? (if not, `CP_CANFDTxMaxDataLength` might be incorrect there...)

odxtools/diaglayer.py
andlaus: `None` may also know "not applicable", right? if we know it's CAN but not CAN-FD the maximum frame payload size is always 8 bytes. I guess this is what > 90% of people would expect? (I admit that determining whether to use CAN FD is currently quite sketchy, though...)

odxtools/diaglayer.py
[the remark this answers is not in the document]
andlaus: okay, I misremembered (haven't found such a dataset). I then looked at the description of the `CP_CANFDTxMaxDataLength` parameter and it seems to imply that CAN-FD shall only be used if "CANFD" is mentioned in the value of `CP_CANFDTxMaxDataLength`...

(changed: 2cc565762ac6fe07ae7a0bb37dbcee034585c49b)

odxtools/diaglayer.py
```diff
@@ -636,9 +643,101 @@ def get_comparam(
...
+        if m:
+            return int(m.group(1))
+
+        # unexpected format of parameter value
+        return 8
+
+    def use_can(self, protocol: Optional[Union[str, "DiagLayer"]] = None) -> bool:
```
kayoub5, line 675: Prefixing the method with "use" means that it will do some kind of state change, and it's not a getter, which does not reflect its implementation, same for use_can_fd
```suggestion
    def get_can_used(self, protocol: Optional[Union[str, "DiagLayer"]] = None) -> bool:
```

odxtools/diaglayer.py
```diff
@@ -636,9 +643,101 @@ def get_comparam(
 
         return cps[0]
 
-    def get_can_receive_id(self, protocol_name: Optional[str] = None) -> Optional[int]:
+    def get_can_mtu(self, protocol: Optional[Union[str, "DiagLayer"]] = None) -> Optional[int]:
+        """Return the maximum size of CAN frames that can be
```
kayoub5, line 647:
```suggestion
        """Return the maximum size of a CAN frame payload that can be
```

kayoub5: In Linux Socket CAN, CAN MTU includes the CAN header, maybe stick to the standing naming of "Max Data Length"

odxtools/diaglayer.py
andlaus: I don't agree with you 100%, but I see your point (to me the verb can also be a question, i.e., it does not necessarily need to be an imperative). I renamed this to [`uses_can()`](e6459950662b002641d540ffcc471db993dd3e4b). (note that I think this would better be a property, but properties cannot take arguments...)

odxtools/diaglayer.py
andlaus: changed: e6459950662b002641d540ffcc471db993dd3e4b

kayoub5 (approved):

PR #245, opened 2023-12-01
Add support for InternalConstr

willzhang05:
Implemented the classes based on the schema in ISO 22901-1:2008 pg. 376-377, 394

### DIFF
--- a/odxtools/internalconstr.py
+++ b/odxtools/internalconstr.py
@@ -1,59 +1,19 @@
 # SPDX-License-Identifier: MIT
 from dataclasses import dataclass
-from enum import Enum
 from typing import List, Optional
 from xml.etree import ElementTree
 
 from .compumethods.limit import Limit
-from .exceptions import odxraise, odxrequire
 from .odxtypes import DataType
-from .utils import create_description_from_et
-
-
-class Validity(Enum):
-    VALID = "VALID"
-    NOT_VALID = "NOT-VALID"
-    NOT_DEFINED = "NOT-DEFINED"
-    NOT_AVAILABLE = "NOT-AVAILABLE"
-
-
-@dataclass
-class ScaleConstr:
-    """This class represents a SCALE-CONSTR.
-    """
-
-    short_label: Optional[str]
-    description: Optional[str]
-    lower_limit: Optional[Limit]
-    upper_limit: Optional[Limit]
-    validity: Validity
-
-    @staticmethod
-    def from_et(et_element: ElementTree.Element, internal_type: DataType) -> "ScaleConstr":
-        short_label = et_element.findtext("SHORT-LABEL")
-        description = create_description_from_et(et_element.find("DESC"))
-
-        lower_limit = Limit.from_et(et_element.find("LOWER-LIMIT"), internal_type=internal_type)
-        upper_limit = Limit.from_et(et_element.find("UPPER-LIMIT"), internal_type=internal_type)
-
-        validity_str = odxrequire(et_element.get("VALIDITY"))
-        try:
-            validity = Validity(validity_str)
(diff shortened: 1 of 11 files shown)


andlaus:
@willzhang05: what's the status here?

willzhang05:
@andlaus Sorry, been busy with some other things, will try to address these later this week.

andlaus:
> @andlaus Sorry, been busy with some other things, will try to address these later this week.

Great to read :) (I was just wondering if this was still on your radar...)

odxtools/internalconstr.py
```diff
@@ -0,0 +1,34 @@
...
+from xml.etree import ElementTree
+
+from .compumethods.limit import Limit
+from .odxtypes import DataType
+from .scaleconstr import ScaleConstr
+
+
+@dataclass
+class InternalConstr:
+    """This class represents a INTERNAL-CONSTR.
```
kayoub5, line 13: this class now represents both internal and physical constr, so maybe update clasd name and dovs?

odxtools/scaleconstr.py
```diff
@@ -0,0 +1,50 @@
...
+    NOT_DEFINED = "NOT-DEFINED"
+    NOT_AVAILABLE = "NOT-AVAILABLE"
+
+
+@dataclass
+class ScaleConstr:
+    """This class represents a SCALE-CONSTR.
+    """
+
+    short_label: Optional[str]
```
kayoub5, line 25: there is a short label, but no long label?

odxtools/write_pdx_file.py
```diff
@@ -119,6 +119,8 @@ def write_pdx_file(
         jinja_env.globals["make_bool_xml_attrib"] = make_bool_xml_attrib
 
         vars: Dict[str, Any] = {}
+        # set the index.xml short name tag to the filename without the file extension.
+        vars["ecu_name"] = os.path.splitext(os.path.basename(output_file_name))[0].replace(".", "_")
```
kayoub5, line 123: those changes look unrelated to the rest of the PR, maybe move them to their own PR?

odxtools/scaleconstr.py
uerg: yes, the XSD says only short label
image

odxtools/internalconstr.py
andlaus: this is exactly the same as in the XSD, so I think this is okay. (I agree that this should be done differently by the spec...)

odxtools/write_pdx_file.py
andlaus: also, this seems to be incorrect from a logical point of view: `output_file_name` is set by the `for` loop on line 100, so whatever it contains is just the first item of the last entry of `auxiliary_content_specifiers` (note that it also might not exist at all...)

andlaus: further, the `ecu_name` variable does not seem to be used by the writing code, so these two lines should probably be removed...

odxtools/internalconstr.py
kayoub5: Class name wise, that's fine due to the spec, but the docs can still be updated, right?

andlaus: true
```suggestion
    """This class represents a INTERNAL-CONSTR objects.

In DOPs, INTERNAL-CONSTR objects can be used to specify validity conditions for internal representations of quantities 
(INTERNAL-CONSTR) and for quantities in the physical domain (PHYS-CONSTR).
```

kayoub5 (approved):

andlaus:
looks good; let's merge. thanks everybody!

PR #246, opened 2023-12-07
Refactor `DiagCodedType` coding

andlaus:
This PR refactors the de- and encoding logic of `DiagCodedType`. The aim is to make the code easier understand, speed it up a bit and to fix a few minor correctness issues...

### DIFF
--- a/odxtools/diagcodedtype.py
+++ b/odxtools/diagcodedtype.py
@@ -3,7 +3,10 @@ import abc
 from dataclasses import dataclass
 from typing import TYPE_CHECKING, Any, Dict, Literal, Optional, Tuple, Union
 
-import bitstruct
+try:
+    import bitstruct.c as bitstruct
+except ImportError:
+    import bitstruct
 
 from . import exceptions
 from .decodestate import DecodeState
@@ -23,8 +26,8 @@ ODX_TYPE_TO_FORMAT_LETTER = {
     DataType.A_FLOAT64: "f",
     DataType.A_BYTEFIELD: "r",
     DataType.A_UNICODE2STRING: "r",  # UTF-16 strings must be converted explicitly
-    DataType.A_ASCIISTRING: "t",
-    DataType.A_UTF8STRING: "t",
+    DataType.A_ASCIISTRING: "r",
+    DataType.A_UTF8STRING: "r",
 }
 
 # Allowed diag-coded types
@@ -103,9 +106,12 @@ class DiagCodedType(abc.ABC):
         ]:
             extracted_bytes = extracted_bytes[::-1]
 
-        format_letter = ODX_TYPE_TO_FORMAT_LETTER[base_data_type]
         padding = (8 - (bit_length + bit_position) % 8) % 8
-        text_encoding = 'utf-8'
+        internal_value, = bitstruct.unpack_from(
+            f"{ODX_TYPE_TO_FORMAT_LETTER[base_data_type]}{bit_length}",
+            extracted_bytes,
+            offset=padding)
+
         text_errors = 'strict' if exceptions.strict_mode else 'replace'
         if base_data_type == DataType.A_ASCIISTRING:
             # The spec says ASCII, meaning only byte values 0-127.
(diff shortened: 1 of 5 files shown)


odxtools/diagcodedtype.py
```diff
@@ -3,7 +3,10 @@
 from dataclasses import dataclass
 from typing import TYPE_CHECKING, Any, Dict, Literal, Optional, Tuple, Union
 
-import bitstruct
+try:
+    import bitstruct.c as bitstruct
```
kayoub5, line 7: Are you sure you want to use `bitstruct.c` ? There are limitations
* Integers and booleans must be 64 bits or less. => ODX limit integers to 64 bits?
* Text and raw must be a multiple of 8 bits. => should be fine
* Bit endianness and byte order are not yet supported. => matters in ODX

See https://github.com/qchateau/cbitstruct as alternative, limitations there sound a bit more realistic for ODX use cases

odxtools/diagcodedtype.py
```diff
@@ -23,8 +26,8 @@
     DataType.A_FLOAT64: "f",
     DataType.A_BYTEFIELD: "r",
     DataType.A_UNICODE2STRING: "r",  # UTF-16 strings must be converted explicitly
-    DataType.A_ASCIISTRING: "t",
-    DataType.A_UTF8STRING: "t",
+    DataType.A_ASCIISTRING: "r",
+    DataType.A_UTF8STRING: "r",
```
kayoub5, line 30: Why?

odxtools/diagcodedtype.py
```diff
@@ -89,46 +94,45 @@ def _extract_internal(
...
-        # Apply byteorder
-        if not is_highlow_byte_order and base_data_type not in [
-                DataType.A_UNICODE2STRING,
-                DataType.A_BYTEFIELD,
-                DataType.A_ASCIISTRING,
-                DataType.A_UTF8STRING,
+        # Apply byteorder for numerical objects. Note that doing this
+        # here might lead to garbage data being included in the result
+        # if the data to be extracted is not byte aligned and crosses
+        # byte boundaries, but it is what the specification says.
```
kayoub5, line 100: What part of the spec?

odxtools/diagcodedtype.py
```diff
@@ -89,46 +94,45 @@ def _extract_internal(
...
             extracted_bytes = extracted_bytes[::-1]
 
-        format_letter = ODX_TYPE_TO_FORMAT_LETTER[base_data_type]
-        padding = 8 * byte_length - (bit_length + bit_position)
-        text_encoding = 'utf-8'
-        text_errors = 'strict'
-        if not exceptions.strict_mode:
-            text_errors = 'replace'
+        padding = (8 - (bit_length + bit_position) % 8) % 8
+        internal_value, = bitstruct.unpack_from(
```
kayoub5, line 110: Consider compiling the bitstruct format using `bitstruct.compile` and cache the result in the codedtype itself, for a significant performance boost

odxtools/diagcodedtype.py
andlaus: because `bitstruct.c` does not support en- and decoding string objects. Also, the previous error handling code was wrong for UTF-8 strings containing multi-byte characters (`len(s)` is the number of characters, not the number if bytes.).

odxtools/diagcodedtype.py
andlaus: I played with this thought, but concluded that it is not worth the effort: it would imply that the arguments of the `_extract_internal_value()` function must be baked into the `CompiledFormat` object (i.e., a major redesign of code that uses this function would be required), and second, for `bitstruct.c` the unpack operation with these simple formats should be about the same speed using compiled objects as for the uncompiled ones because I suspect that performance is dominated by the overhead of the python interpreter not by the execution of the function itself (I haven't benchmarked it, though...)

odxtools/diagcodedtype.py
andlaus: > Are you sure you want to use bitstruct.c ?

pretty much.

> Integers and booleans must be 64 bits or less. => ODX limit integers to 64 bits?

I think it even limits it to 32 bits, see section 7.1.3 (ASAM version of the document).

> Bit endianness and byte order are not yet supported.

we do (and always did) the byte order handling manually. (byte order specifiers are unsupported, i.e., everything i assumed to be big endian by `bitstruct.c`, IIRC.)

odxtools/diagcodedtype.py
andlaus: okay, I benchmarked it a bit: using `bitstruct.c` instead of `bitstruct` to unpack the format `"p4u8"` comes with a speedup of about factor 30; using a precompiled format object with `bitstruct.c` yields another speedup of roughly factor `1.5`. That said, I am pretty certain that the coding functions are not dominated by the packing and unpacking routines when using `bitstruct.c`...

odxtools/diagcodedtype.py
andlaus: > See https://github.com/qchateau/cbitstruct as alternative, limitations there sound a bit more realistic for ODX use cases

it seems like this does not support byte fields longer than 64 bits. (if the README is correct...)

PR #248, opened 2023-12-08
Fix ODX 2.0 compatibility

kayoub5:
ODX file had `2.0.1` as model version


andlaus:
thanks. my exposure to ODX 2.0 datasets is rather limited (read: non-existant), so I suppoose you'll continue to have the "honor" of being the guardian of 2.0 compatibility ;)

PR #250, opened 2023-12-19
List improvements

kakoeh:
PR is spin-off of PR#249

Katja Köhler &lt;katja.koehler@mercedes-benz.com&gt;, on behalf of [MBition GmbH](https://mbition.io/).

### DIFF
--- a/odxtools/cli/_print_utils.py
+++ b/odxtools/cli/_print_utils.py
@@ -19,14 +19,13 @@ def format_desc(desc: str, ident: int = 0) -> str:
     return desc
 
 
-def print_diagnostic_service(
-    service: DiagService,
-    print_params: bool = False,
-    print_pre_condition_states: bool = False,
-    print_state_transitions: bool = False,
-    print_audiences: bool = False,
-    allow_unknown_bit_lengths: bool = False,
-) -> None:
+def print_diagnostic_service(service: DiagService,
+                             print_params: bool = False,
+                             print_pre_condition_states: bool = False,
+                             print_state_transitions: bool = False,
+                             print_audiences: bool = False,
+                             allow_unknown_bit_lengths: bool = False,
+                             plumbing_output: bool = False) -> None:
     print(f" {service.short_name} <ID: {service.odx_id}>")
 
     if service.description:
--- a/odxtools/cli/_print_utils.py
+++ b/odxtools/cli/_print_utils.py
@@ -1,9 +1,22 @@
 # SPDX-License-Identifier: MIT
 import re
+from typing import Any, Dict, List, Optional, Union
 
 import markdownify
+from rich import console, print
+from tabulate import tabulate
 
+from ..diaglayer import DiagLayer
 from ..diagservice import DiagService
+from ..parameters.codedconstparameter import CodedConstParameter
+from ..parameters.nrcconstparameter import NrcConstParameter
+from ..parameters.parameter import Parameter
+from ..parameters.physicalconstantparameter import PhysicalConstantParameter
+from ..parameters.systemparameter import SystemParameter
+from ..parameters.valueparameter import ValueParameter
+from ..singleecujob import SingleEcuJob
+
+terminal = console.Console()
 
 
 def format_desc(desc: str, ident: int = 0) -> str:
@@ -26,7 +39,7 @@ def print_diagnostic_service(service: DiagService,
                              print_audiences: bool = False,
                              allow_unknown_bit_lengths: bool = False,
                              plumbing_output: bool = False) -> None:
-    print(f" {service.short_name} <ID: {service.odx_id}>")
+    print(f" [cyan]{service.short_name}[/cyan] <ID: {service.odx_id}>")
 
     if service.description:
         desc = format_desc(service.description, ident=3)
@@ -52,27 +65,71 @@ def print_diagnostic_service(service: DiagService,
         print(f"  Enabled-Audiences: {', '.join(enabled_audiences_short_names)}")
 
     if print_params:
-        assert service.request is not None
-        assert service.positive_responses is not None
-        assert service.negative_responses is not None
+        print_service_parameters(
+            service,
+            allow_unknown_bit_lengths=allow_unknown_bit_lengths,
+            plumbing_output=plumbing_output)
 
-        print(f"  Message format of a request:")
-        service.request.print_message_format(
-            indent=3, allow_unknown_lengths=allow_unknown_bit_lengths)
 
-        print(f"  Number of positive responses: {len(service.positive_responses)}")
-        if len(service.positive_responses) == 1:
-            resp = service.positive_responses[0]
+def print_service_parameters(service: DiagService,
+                             allow_unknown_bit_lengths: bool = False,
+                             plumbing_output: bool = False) -> None:
+    # prints parameter details of request, posivite response and negative response of diagnostic service
 
-            print(f"  Message format of a positive response:")
-            resp.print_message_format(indent=3, allow_unknown_lengths=allow_unknown_bit_lengths)
+    assert service.request is not None
+    assert service.positive_responses is not None
+    assert service.negative_responses is not None
 
-        print(f"  Number of negative responses: {len(service.negative_responses)}")
-        if len(service.negative_responses) == 1:
-            resp = service.negative_responses[0]
+    # Request
+    print(f"\n  [yellow]Request Properties[/yellow]:")
+    print(f"   Request Name: {service.request.short_name}")
 
-            print(f"  Message format of a negative response:")
-            resp.print_message_format(indent=3, allow_unknown_lengths=allow_unknown_bit_lengths)
+    if service.request and not service.request.required_parameters:
+        ba = f"   Byte-Array: {service()!r}"
+        hs = f"   Hex-String: 0x{str(service().hex().upper())}"
+        terminal.print(ba, overflow="ellipsis", soft_wrap=True)
+        terminal.print(hs, overflow="ellipsis", soft_wrap=True)
+    else:
+        print(f"   Byte-Array: ---\n   Hex-String: ---")
+
+    print(f"   Service Parameters: {service.request.parameters}\n")
+    table = extract_parameter_tabulation_data(list(service.request.parameters))
+    print(tabulate(table, headers='keys', tablefmt='presto'))
+    print(f"\n   Message format of the request:")
+    service.request.print_message_format(
+        indent=0, allow_unknown_lengths=allow_unknown_bit_lengths, plumbing_output=plumbing_output)
+
+    # Positive Response
+    print(f"\n  [yellow]Positive Response Properties[/yellow]:")
+    print(f"   Number of Positive Responses: {len(service.positive_responses)}")
+    print(f"   Positive Responses: {service.positive_responses}")
+    if len(service.positive_responses) == 1:
+        resp = service.positive_responses[0]
+        print(f"   Service Parameters: {resp.parameters}\n")
+        table = extract_parameter_tabulation_data(list(resp.parameters))
+        print(tabulate(table, headers='keys', tablefmt='presto'))
+        print(f"\n   Message format of the positive response:")
+        resp.print_message_format(
+            indent=0,
+            allow_unknown_lengths=allow_unknown_bit_lengths,
+            plumbing_output=plumbing_output)
+
+    # Negative Response
+    print(f"\n  [yellow]Negative Response Properties[/yellow]:")
+    print(f"   Number of Negative Responses: {len(service.negative_responses)}")
+    print(f"   Negative Responses: {service.negative_responses}")
+    if len(service.negative_responses) == 1:
+        resp = service.negative_responses[0]
+        print(f"   Service Parameters: {resp.parameters}\n")
+        table = extract_parameter_tabulation_data(list(resp.parameters))
+        print(tabulate(table, headers='keys', tablefmt='presto'))
+        print(f"\n   Message format of a negative response:")
+        resp.print_message_format(
+            indent=0,
+            allow_unknown_lengths=allow_unknown_bit_lengths,
+            plumbing_output=plumbing_output)
+
+    print("\n")
 
     if (service.positive_responses and
             len(service.positive_responses) > 1) or (service.negative_responses and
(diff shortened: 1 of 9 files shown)


odxtools/cli/_print_utils.py
```diff
@@ -1,9 +1,29 @@
...
+from ..singleecujob import SingleEcuJob
+
+
+class Style:
+    # class with variables and functions to display the result of the comparison
+
+    ## text style variables
+    bold = '\033[1m'
+    underline = '\033[4m'
+    italic = '033[3m'  # doesn't work with Windows CMD and PowerShell
```
kayoub5, line 25: Normally I would recommend using already established python modules that handle coloring, detect if console does not support color, etc.

See https://pypi.org/project/termcolor/ or https://pypi.org/project/rich/

odxtools/cli/_print_utils.py
```diff
@@ -19,14 +39,13 @@ def format_desc(desc: str, ident: int = 0) -> str:
...
 
-def print_diagnostic_service(
-    service: DiagService,
-    print_params: bool = False,
-    print_pre_condition_states: bool = False,
-    print_state_transitions: bool = False,
-    print_audiences: bool = False,
-    allow_unknown_bit_lengths: bool = False,
-) -> None:
+def print_diagnostic_service(service: DiagService,
```
kayoub5, line 42: original code style of keeping first argument in its own lime is what I personally prefer, but the final say on such topics go to @andlaus

odxtools/cli/_print_utils.py
```diff
@@ -53,31 +72,208 @@ def print_diagnostic_service(
...
 
     if (service.positive_responses and
             len(service.positive_responses) > 1) or (service.negative_responses and
                                                      len(service.negative_responses) > 1):
         # Does this ever happen?
         raise NotImplementedError(
             f"The diagnostic service {service.odx_id} offers more than one response!")
+
+
+def generatetable(style: str, list_of_items: Union[List[DiagService],
```
kayoub5, line 149: you are reinventing rich tables https://rich.readthedocs.io/en/stable/tables.html

odxtools/diagcodedtype.py
```diff
@@ -142,6 +142,8 @@ def _encode_internal_value(
         """Convert the internal_value to bytes."""
         # Check that bytes and strings actually fit into the bit length
         if base_data_type == DataType.A_BYTEFIELD:
+            if isinstance(internal_value, bytearray):
```
kayoub5, line 145: @andlaus are we accepting bytearray again?

odxtools/cli/_print_utils.py
andlaus: I agree, but I think this was imposed by yapf or else the workflow would complain. (the point of enforcing yapf's style is to avoid loosing to much time with discussions about coding style.) That said, maybe we should have a hard look at yapf's options in the not too distant future...

odxtools/cli/_print_utils.py
```diff
@@ -53,31 +72,208 @@ def print_diagnostic_service(
...
+def print_service_parameters(service: DiagService,
+                             allow_unknown_bit_lengths: bool = False,
+                             print_objects: bool = False) -> None:
+    # prints parameter details of request, posivite response and negative response of diagnostic service
 
-            print(f"  Message format of a positive response:")
-            resp.print_message_format(indent=3, allow_unknown_lengths=allow_unknown_bit_lengths)
+    assert service.request is not None
+    assert service.positive_responses is not None
+    assert service.negative_responses is not None
```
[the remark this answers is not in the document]
andlaus: they fix some mypy errors. Given that this is not in the core library but "only" in the CLI utilities, it is IMO okay if they fail for pathological datasets...

odxtools/cli/list.py
```diff
@@ -196,9 +224,9 @@ def service_filter(s: DiagComm) -> bool:
         print_dops=args.all or args.dops,
         variants=None if variants == "all" else variants,
         print_params=args.all or args.params,
-        print_comparams=args.all,
+        print_com_params=args.all,
```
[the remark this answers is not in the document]
andlaus: good question; `comparam`? I don't care too much if this is inconsistent, though...

odxtools/diagcodedtype.py
andlaus: it would be nice, because mypy allows `bytearray` where `bytes` is specified. the problem is that `bitstruct.c` currently only accepts `bytearray` for the `unpack()` function, not for `pack()`. I suppose I should fix that...

odxtools/cli/list.py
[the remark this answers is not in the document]
kakoeh: Done

odxtools/cli/_print_utils.py
andlaus: as far as I can see, this function is a bit of a misnomer: it does not produce the ASCII art itself, but only extracts the data expected by [`tabulate.tabulate()`](https://pypi.org/project/tabulate/) (which indeed seems to be a simpler version of the "rich" module). 

```suggestion
def extract_tabulation_data(style: str, list_of_items: Union[List[DiagService],
```

Besides that, what I'd like to see is to make the `style` argument an enum, so that there is less room for error, i.e.,
```suggestion
class TabulationMode(Enum):
    SERVICE_DETAILS = "service-details"
    PARAMETER_DETAILS = "parameter-details"

def extract_tabulation_data(mode: TabulationMode, list_of_items: Union[List[DiagService],
```

Even better (?), the function could be split:
```python
def extract_service_tabulation_data(services: List[DiagService]) -> Dict[str, Any]:
    ...

def extract_parameter_tabulation_data(parameters: List[Parameter]) -> Dict[str, Any]:
    ...
```

odxtools/cli/_print_utils.py
kakoeh: text formatting in the terminal is done via the python module `rich` now

odxtools/cli/_print_utils.py
kakoeh: The functionality of `generatetable` was split into `extract_service_tabulation_data` and `extract_parameter_tabulation_data` as suggested. It might be reasonable to switch from tabulate to rich tables in the future as the rich module is already used for formatting the text output in the terminal.

odxtools/cli/_print_utils.py
kayoub5: @kakoeh did you change the argument position to be on the same line as the function, or did yapf did that on its own?

odxtools/cli/_print_utils.py
```diff
@@ -53,31 +63,197 @@ def print_diagnostic_service(
...
+
+def extract_service_tabulation_data(services: List[DiagService]) -> Dict[str, Any]:
+    # extracts data of diagnostic services into Dictionary which can be printed by tabulate module
+
+    name = []
+    semantic = []
+    request: List[Optional[str]] = []
+
+    for service in services:
+        assert isinstance(service, DiagService)
```
kayoub5, line 144: Why this check is here?

odxtools/cli/_print_utils.py
```diff
@@ -53,31 +63,197 @@ def print_diagnostic_service(
...
+        name.append(service.short_name)
+        semantic.append(service.semantic)
+        if not getattr(service.request, "required_parameters", None):
```
kayoub5, line 147: `required_parameters` is not always present as a property?

odxtools/cli/_print_utils.py
```diff
@@ -53,31 +63,197 @@ def print_diagnostic_service(
...
-        print(f"  Number of negative responses: {len(service.negative_responses)}")
-        if len(service.negative_responses) == 1:
-            resp = service.negative_responses[0]
+    # Request
+    print(f"\n  [yellow]Request Properties[/yellow]:")
+    print(f"   Request Name: {service.request.short_name}")
+    if not getattr(service.request, "required_parameters", None):
+        if len(service()) > 24:
+            print(
+                f"   Byte-Array: {service()[:24]!r}...\tHex-String: 0x{service().hex().upper()[:24]}..."
```
kayoub5, line 87: Why 24 ?

odxtools/cli/_print_utils.py
```diff
@@ -53,31 +63,197 @@ def print_diagnostic_service(
...
+    assert service.negative_responses is not None
 
-        print(f"  Number of negative responses: {len(service.negative_responses)}")
-        if len(service.negative_responses) == 1:
-            resp = service.negative_responses[0]
+    # Request
+    print(f"\n  [yellow]Request Properties[/yellow]:")
+    print(f"   Request Name: {service.request.short_name}")
+    if not getattr(service.request, "required_parameters", None):
+        if len(service()) > 24:
```
kayoub5, line 85: Nitpick: maybe cache the return value of `service()`

odxtools/cli/_print_utils.py
```diff
@@ -53,31 +63,197 @@ def print_diagnostic_service(
...
+    semantic = []
+    request: List[Optional[str]] = []
+
+    for service in services:
+        assert isinstance(service, DiagService)
+        name.append(service.short_name)
+        semantic.append(service.semantic)
+        if not getattr(service.request, "required_parameters", None):
+            if len(service().hex().upper()) > 24:
+                request.append('0x' + service().hex().upper()[:24] + '...')
```
kayoub5, line 149: Maybe a function that takes bytes as input, and return truncated data or full data as string? same logic in print_service_parameters

odxtools/cli/_print_utils.py
```diff
@@ -53,31 +63,197 @@ def print_diagnostic_service(
...
+    bit_length: List[Optional[int]] = []
+    semantic = []
+    param_type = []
+    value: List[Optional[str]] = []
+    value_type: List[Optional[str]] = []
+    data_type: List[Optional[str]] = []
+    dop: List[Optional[str]] = []
+
+    for param in parameters:
+        assert isinstance(param, Parameter)
```
kayoub5, line 172: What else would it be?

odxtools/cli/list.py
```diff
@@ -2,72 +2,93 @@
...
+            diag_layers.append([x for x in odxdb.diag_layers if x.short_name == name][0])
+
+        else:
+            print(f"The variant '{name}' could not be found!")
+            return
+
+    if diag_layers:
+        print("\n")
+        print(f"Overview of diagnostic layers: ")
+        print_ecu_variant_metrics(diag_layers)
```
kayoub5, line 50: Naming: ecu variants or diag layers?

odxtools/cli/list.py
```diff
@@ -2,72 +2,93 @@
...
+
+    for dl in diag_layers:
+        print("\n")
+        print(f"[green]Diagnostic layer:[/green] '[bold white]{dl.short_name}[/bold white]'")
+        print(f" [blue]Variant Type[/blue]: {dl.variant_type.value}")
```
kayoub5, line 55:
```suggestion
        print(f"  [blue]Variant Type[/blue]: {dl.variant_type.value}")
```

odxtools/cli/list.py
```diff
@@ -2,72 +2,93 @@
...
+                    print(f" {gnr}")
+                else:
+                    print(f" {gnr.short_name}")
 
         if print_services and len(all_services) > 0:
             services = [s for s in all_services if service_filter(s)]
             if len(services) > 0:
-                print(f"The services of '{dl.short_name}' are: ")
+                print("\n")
+                print(f"The [blue]services[/blue] of '{dl.short_name}' are: ")
```
kayoub5, line 91: Feels like the "The" prefix is added/removed at random

odxtools/cli/list.py
```diff
@@ -77,23 +98,28 @@ def print_summary(
...
 
-        if print_comparams and len(comparams) > 0:
+        if print_comparam and len(comparam) > 0:
+            print("\n")
             print(
-                f"The communication parameters of the {dl.variant_type.value} '{dl.short_name}' are: "
+                f"The [blue]communication parameters[/blue] of the {dl.variant_type.value} '{dl.short_name}' are: "
             )
-            for com_param in comparams:
+            for com_param in comparam:
```
kayoub5, line 122:
```suggestion
            for com_param in comparams:
```

pyproject.toml
```diff
@@ -42,6 +42,8 @@ dependencies = [
     "markdownify >= 0.11",
     "deprecation >= 2.1",
     "packaging",
+    "tabulate >= 0.9.0",
+    "rich >= 13.7.0",
```
kayoub5, line 46: Having both rich and tabulate feels like an overkill, fairly certain they won't play nice with each other (use same colors etc)

odxtools/cli/list.py
```diff
@@ -196,9 +230,9 @@ def service_filter(s: DiagComm) -> bool:
         print_dops=args.all or args.dops,
         variants=None if variants == "all" else variants,
         print_params=args.all or args.params,
-        print_comparams=args.all,
+        print_comparam=args.all,
```
kayoub5, line 233:
```suggestion
        print_comparams=args.all,
```

odxtools/cli/list.py
```diff
@@ -2,72 +2,93 @@
...
         all_services: List[DiagComm] = sorted(dl.services, key=lambda x: x.short_name)
 
         data_object_properties = dl.diag_data_dictionary_spec.data_object_props
-        comparams = dl.comparams
-
-        print(f"{dl.variant_type} '{dl.short_name}'")
-        print(
-            f" num services: {len(all_services)}, num DOPs: {len(data_object_properties)}, num communication parameters: {len(comparams)}."
-        )
+        comparam = dl.comparams
```
kayoub5, line 61:
```suggestion
        comparams = dl.comparams
```

odxtools/cli/list.py
kayoub5: Not every diag layer is an ECU variant

odxtools/cli/_print_utils.py
```diff
@@ -53,31 +63,197 @@ def print_diagnostic_service(
...
+                data_type.append(None)
+            if isinstance(param, PhysicalConstantParameter):
+                if isinstance(param.physical_constant_value, int):
+                    value.append("0x" + f"{int(param.physical_constant_value):0{length}X}")
+                else:
+                    value.append(f"{param.physical_constant_value!r}")
+                value_type.append('constant value')
+            elif isinstance(param, ValueParameter) and param.physical_default_value is not None:
+                if isinstance(param.physical_default_value, int):
+                    value.append("0x" + f"{int(param.physical_default_value):0{length}X}")
```
kayoub5, line 209: Why force all ints to be hex?

odxtools/cli/_print_utils.py
kakoeh: I didn't change it, this was performed by yapf

odxtools/cli/_print_utils.py
kakoeh: `required_parameters` is a function that returns a list of all parameters where the property `is_required` is set to true. This list can be empty.

odxtools/cli/_print_utils.py
kakoeh: 24 is the compromise of displaying the request string in one line (truncate output for very long request strings to avoid line breaks) and display as much information as possible. As the return value of service() is a bytearray, it seemed reasonable to me to truncate the data to a multiple of 8.

odxtools/cli/_print_utils.py
kayoub5: that still does not explain the need for getattr

odxtools/cli/_print_utils.py
kayoub5, line 87: tip: use ellipse character instead of three dots

https://www.compart.com/en/unicode/U+2026

andlaus: I guess three dots are better; it keeps the output ASCII as long as no non-ascii characters are featured by in the dataset...

odxtools/cli/_print_utils.py
andlaus: That said, that piece of code should be
```suggestion
        if getattr(service.request, "required_parameters", None) is None:
```

odxtools/cli/_print_utils.py
andlaus: (seems to be an artifact of splitting the original function :))

odxtools/cli/_print_utils.py
andlaus: also, better use
```suggestion
                    value.append(f"0x{int(param.physical_default_value):0{length}X}")
```
instead of concatinating strings...

odxtools/cli/_print_utils.py
kakoeh: Leftover from old function `generatetable`, was removed now.

odxtools/cli/_print_utils.py
kakoeh: Indeed, check was removed now.

odxtools/cli/_print_utils.py
kakoeh: new function `format_request_string` was introduced.

odxtools/cli/_print_utils.py
kakoeh: new function `format_request_string` was introduced.

odxtools/cli/list.py
kakoeh: renamed into `print_dl_metrics`

odxtools/cli/list.py
kakoeh: Renamed

odxtools/cli/list.py
kakoeh: Renamed

odxtools/cli/list.py
kakoeh: Renamed

pyproject.toml
kakoeh: Seems to work fine:
image
Nevertheless, I'm open to get rid of tabulate and switch to rich tables in the future

odxtools/cli/list.py
kakoeh: The old output `print(f"{dl.variant_type} '{dl.short_name}'")` was split into
`print(f"[green]Diagnostic layer:[/green] '[bold white]{dl.short_name}[/bold white]'")` and
 `print(f" [blue]Variant Type[/blue]: {dl.variant_type.value}")`
 Indentation was chosen to be the same as in `print(f" Description: " + desc)`

odxtools/cli/find.py
```diff
@@ -120,6 +121,14 @@ def add_subparser(subparsers: "argparse._SubParsersAction") -> None:
         help="Relax output formatting rules (allow unknown bitlengths for ascii representation)",
     )
 
+    parser.add_argument(
+        "-od",
+        "--object-details",
+        action="store_true",
+        required=False,
+        help="Print all object details instead of just the name",
```
andlaus, line 129: hm, I do not really like the name of this parameter because, if I understood this correctly, there are more attributes considered in the printed tables?  maybe rename it to `--print-plumbing` (in reference to [git](https://git-scm.com/book/en/v2/Git-Internals-Plumbing-and-Porcelain))?

odxtools/cli/list.py
kakoeh: As stated above, I changed the output of 
```
print(f"{dl.variant_type} '{dl.short_name}'")
print(f" num services: {len(all_services)}, num DOPs: {len(data_object_properties)}, num communication parameters: {len(comparams)}.")
```
For the rest of the function I didn't change the text but only formats.

odxtools/cli/_print_utils.py
kakoeh: > Why force all ints to be hex?

Doesn't is make sense to display the values as hex as they're sent this way? I don't mind displaying ints as ints if it makes more sense though.

odxtools/cli/find.py
kakoeh: A table with selected attributes is only displayed for request and response parameters. The global negative responses and DOPs are displayed by their attribute `short_name`. 
I renamed the argument to `--plumbing-output`. When specified, the full object is printed in addition to the table (for request and response parameters) or instead of the name (for global negative responses and DOPs).

odxtools/cli/_print_utils.py
kayoub5: the values are sent as binary, not hex, hex is just a human interpretation of numeric values and binary data.

it would not make sense to show a variable that represents a physical quantity (time, speed, distance, temperature) as hex.

kayoub5: as a rule of thumb, if it has a unit, hex makes no sense.

odxtools/cli/_print_utils.py
kayoub5: you are reinventing the overflow argument of rich print function

https://rich.readthedocs.io/en/stable/console.html#overflow

odxtools/cli/find.py
```diff
@@ -131,4 +140,4 @@ def run(args: argparse.Namespace) -> None:
         service_names=args.service_names,
         print_params=not args.no_details,
         allow_unknown_bit_lengths=args.relaxed_output,
-    )
+        print_objects=args.plumbing_output)
```
kayoub5, line 143: keep same variable name across codebase
```suggestion
        plumbing_output=args.plumbing_output)
```

odxtools/cli/_print_utils.py
kayoub5: in that case condition should be `if service.request and service.request.required_parameters:`

odxtools/cli/list.py
kayoub5: @andlaus maybe a clean-up for another PR?

odxtools/cli/_print_utils.py
kakoeh: rich print function
```
ba = f"   Byte-Array: {service()!r}"
hs = f"\n   Hex-String: 0x{str(service().hex().upper())}"
console.print(ba, overflow="ellipsis")
console.print(hs, overflow="ellipsis")
```
resulted in
```
   Byte-Array:
bytearray(b".\x??\x??\x??\x??\x??\x??\x??\x??\x??\x??\x??\x??\x??\x??\x??\x??\x??\x??\x??\x??\x??\x??\x??\x??\x??\…

   Hex-String:
0x????????????????????????????????????????????…
```

I'd prefer old output format without line breaks
```
   Byte-Array: bytearray(b'.\x??\x??\x??\x??\x??\x??\x??\x??')...   Hex-String: 0x????????????????...
```

kayoub5: I think that can be controlled by soft_wrap=True

kayoub5 (approved):

PR #251, opened 2023-12-20
Make `NamedItemList` a list when it comes to type checking

andlaus:
With this applied, whenever a `List` is expected by the type checker, a `NamedItemList` object can be used as well.

Besides this, I cleaned up the `short_name_as_key()` functionality slightly.

### DIFF
--- a/odxtools/nameditemlist.py
+++ b/odxtools/nameditemlist.py
@@ -2,8 +2,8 @@
 import abc
 from collections import OrderedDict
 from keyword import iskeyword
-from typing import (Any, Callable, Collection, Dict, Generic, Iterable, Iterator, List, Optional,
-                    Protocol, Tuple, TypeVar, Union, cast, overload, runtime_checkable)
+from typing import (Any, Callable, Collection, Dict, Iterable, Iterator, List, Optional, Protocol,
+                    SupportsIndex, Tuple, TypeVar, Union, cast, overload, runtime_checkable)
 
 from .exceptions import odxraise
 
@@ -20,7 +20,7 @@ T = TypeVar("T")
 TNamed = TypeVar("TNamed", bound=OdxNamed)
 
 
-class ItemAttributeList(Generic[T]):
+class ItemAttributeList(List[T]):
     """A list that provides direct access to its items as named attributes.
 
     This is a hybrid between a list and a user-defined object: One can
@@ -76,7 +76,7 @@ class ItemAttributeList(Generic[T]):
         self._item_dict[item_name] = item
         self._item_list.append(item)
 
-    def sort(self, key: Optional[Callable[[T], str]] = None, reverse: bool = False) -> None:
+    def sort(self, *, key: Any = None, reverse: bool = False) -> None:
         if key is None:
             self._item_dict = OrderedDict(
                 sorted(self._item_dict.items(), key=lambda x: x[0], reverse=reverse))
@@ -96,7 +96,7 @@ class ItemAttributeList(Generic[T]):
     def items(self) -> Collection[Tuple[str, T]]:
         return self._item_dict.items()
 
-    def __contains__(self, x: T) -> bool:
+    def __contains__(self, x: object) -> bool:
         return x in self._item_list
 
     def __len__(self) -> int:
@@ -108,7 +108,7 @@ class ItemAttributeList(Generic[T]):
         return result
 
     @overload
-    def __getitem__(self, key: int) -> T:
+    def __getitem__(self, key: SupportsIndex) -> T:
         ...
 
     @overload
@@ -119,8 +119,8 @@ class ItemAttributeList(Generic[T]):
     def __getitem__(self, key: slice) -> List[T]:
         ...
 
-    def __getitem__(self, key: Union[int, str, slice]) -> Union[T, List[T]]:
-        if isinstance(key, int):
+    def __getitem__(self, key: Union[SupportsIndex, str, slice]) -> Union[T, List[T]]:
+        if isinstance(key, SupportsIndex):
             return self._item_list[key]
         elif isinstance(key, slice):
             return self._item_list[key]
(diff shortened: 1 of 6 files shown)


odxtools/nameditemlist.py
```diff
@@ -20,7 +20,7 @@ def short_name(self) -> str:
 TNamed = TypeVar("TNamed", bound=OdxNamed)
 
 
-class ItemAttributeList(Generic[T]):
+class ItemAttributeList(List[T]):
```
kayoub5, line 23: `typing.List` is a deprecated alias to `list`, and `list` is not supposed to be extended by user, UserList should be used as base class instead.

Also, the property _item_list makes no sense if the ItemAttributeList is already a list

See https://docs.python.org/3/library/collections.html#collections.UserList

kayoub5: Consider using https://docs.python.org/3/library/collections.abc.html#collections.abc.Collection instead

kayoub5:
The first two commits look OK, if you want to cherry-pick merge them, that would be fine, the last one have issues, see review above

odxtools/nameditemlist.py
```diff
@@ -119,11 +113,11 @@ def __getitem__(self, key: str) -> T:
...
-        if isinstance(key, int):
-            return self._item_list[key]
+    def __getitem__(self, key: Union[SupportsIndex, str, slice]) -> Union[T, List[T]]:
+        if isinstance(key, SupportsIndex):
+            return self.data[key]
         elif isinstance(key, slice):
-            return self._item_list[key]
+            return self.data[key]
         else:
             return self._item_dict[key]
```
kayoub5, line 122: nitpick: function could be simpler by using the divt when the key is str, otherwise fallback to super

odxtools/nameditemlist.py
```diff
@@ -152,7 +146,7 @@ def __eq__(self, other: object) -> bool:
             return self._item_dict == other._item_dict
 
     def __iter__(self) -> Iterator[T]:
-        return iter(self._item_list)
+        return iter(self.data)
```
kayoub5, line 149: useless overload, auto handled by parent class

andlaus: fixed.

odxtools/nameditemlist.py
andlaus: I tried this. this leads to mypy complaints:
```
odxtools/nameditemlist.py:120: error: Incompatible return value type (got "Any | ItemAttributeList[T]", expected "T | list[T]")  [return-value]
odxtools/nameditemlist.py:155: error: Argument 1 of "_get_item_key" is incompatible with supertype "ItemAttributeList"; supertype defines the argument type as "T"  [override]
odxtools/nameditemlist.py:155: note: This violates the Liskov substitution principle
odxtools/nameditemlist.py:155: note: See https://mypy.readthedocs.io/en/stable/common_issues.html#incompatible-overrides
```

I think that these would be fixed if `ItemAttributeList` would directly derive from `UserList[T]`, but this is not allowed in python-3.8...

odxtools/nameditemlist.py
andlaus: hm, unfortunatly deriving from `UserList` does not seem to work as intended: letting mypy loose on
```python
from odxtools.nameditemlist import NamedItemList
from typing import List

class Item:
    def __init__(self, sn: str) -> None:
        self.short_name = sn

def foo(x: List[Item]) -> None:
    print(len(x))

nil = NamedItemList([Item("a"), Item("b")])

foo(nil)
```

produces
```
bla.py:13: error: Argument 1 to "foo" has incompatible type "NamedItemList[Item]"; expected "list[Item]"  [arg-type]
Found 1 error in 1 file (checked 1 source file)
```

any ideas?

kayoub5: try https://docs.python.org/3/library/collections.abc.html#collections.abc.Sequence interface instead

andlaus: the problem with that is that all types would need to be annotated with `Sequence` instead of `List`. I find that rather unintuitive. In the [latest version](https://github.com/mercedes-benz/odxtools/pull/251/commits/67b321891543f359c2ed1090064c7826d447cbdf), I've implemented all public methods of `list`. I'm aware that it's not perfect, but I think it is more than good enough for the intended use case of `NamedItemList`.

kayoub5: list is not an interface, it's a class, you either inherit from it or you don't.

From purely object oriented point of view NamedItemList will not be a List regardless of what functions you implement.

One example where what you are doing will break, is when python 3.13 decides to add new function to list class.

Another one is isinstance check.

andlaus: > list is not an interface, it's a class, you either inherit from it or you don't.

after this PR we do ;)

for now we have to inherit from capital-L list, though because `list` does not yet support generics in python 3.8. (alternatively we can raise the minimum python version to 3.10 and derive from lowercase-l list.)

> From purely object oriented point of view NamedItemList will not be a List regardless of what functions you implement.

but if we inherit from it, it is a list, right?

> One example where what you are doing will break, is when python 3.13 decides to add new function to list class.

as long as this new function would have no effect on the attributes exposed, it would work just fine. This is already the case with some methods, e.g., `sort()` or `reverse()` are not overloaded by `NamedItemList` with this patch.

> Another one is isinstance check.

that is basically the point of this PR: I want `isinstance(named_item_list, list)` to be `True` to drive home the point that this class *is* a list
```
>>> isinstance(ecu.services, list)
True
```
while without it, we get `False`...

andlaus:
@kayoub5: can this be merged? I think a lot of the confusion here stems from the fact that `List[T]` is an (obsolete as of python 3.10) *alias* for `list[T]`, i.e., it is *not* just an interface class for type checking....

kayoub5:
> @kayoub5: can this be merged? I think a lot of the confusion here stems from the fact that `List[T]` is an (obsolete as of python 3.10) _alias_ for `list[T]`, i.e., it is _not_ just an interface class for type checking....

@andlaus I am currently busy with other topics and won't be able to review this at the moment.

andlaus:
okay. I think this PR is basically done from the technical perspective, the issues to be resolved are rather philosophical. (namely, how to implement an extended `list` class.) If you don't  have time, I'll merge the PR some time next week...

andlaus:
just a heads up: if no new showstoppers appear, I intent to merge this and (#253) tomorrow...

kayoub5 (approved):

andlaus:
ok, it's time to take the plunge. @kayoub5: thanks for reviewing!

PR #255, opened 2024-01-19
List tool improvements

andlaus:
This pull request contains various improvements for the CLI tools, in particular for `list`. The motivation for this is that, albeit the new output is pretty, `list` did not work for some of my real-world datasets anymore and -- after I fixed these issues (cf  8b75587d28986323fab102a48259d1463bd50ce2 and a008df688b67eb5c94a63bd8c4142d856447876a) -- it was about an order of magnitude slower than it used to be. Since this is a problem for my smoke testing script (which used to take about 60 seconds to complete and now took about 10 minutes), I added the `--dump-database` option to `list`. This prints everything that gets internalized in an as performant way as possible instead of the printing objects that are handled by the printing helper function in an as intuitive way as possible. (NB that these changes motivated https://github.com/mercedes-benz/odxtools/pull/253 .)


odxtools/cli/list.py
```diff
@@ -61,30 +61,26 @@ def print_summary(odxdb: Database,
...
 
             if (can_tx_id := dl.get_can_send_id(proto.short_name)) is not None:
-                rich.print(
-                    f"  [blue]CAN send ID[/blue] for protocol '{proto.short_name}': 0x{can_tx_id:x}"
-                )
+                rich.print(f"  CAN send ID for protocol '{proto.short_name}': 0x{can_tx_id:x}")
 
         if dl.description:
             desc = format_desc(dl.description, indent=2)
-            rich.print(f" [blue]Description[/blue]: " + desc)
```
kayoub5: Sure about removing colors?

odxtools/nameditemlist.py
```diff
@@ -172,7 +172,7 @@ def __str__(self) -> str:
         return f"[{', '.join( [self._get_item_key(x) for x in self])}]"
 
     def __repr__(self) -> str:
-        return self.__str__()
+        return f"[{', '.join([repr(x) for x in self])}]"
```
kayoub5, line 175: Make sure to include class name
f"NamedItemList[{', '.join([repr(x) for x in self])}]"

odxtools/cli/list.py
andlaus: This is basically a matter of taste, and for my mine the new output of list was a bit too much. (Note that there are still colors, because `rich.print` adds them automatically. this just deals them down a bit.)

odxtools/nameditemlist.py
andlaus: fixed: 36d0ffff36ce29f28f999899de1c93572f877e3e

kayoub5 (approved):

PR #256, opened 2024-01-26
Janitorial work

andlaus:
This PR contains an assortment of smaller fixes, in particular it introduces a "non-strict" mode for `list` and `compare` and cleans up various minor code quality issues in the `compare` tool.

### DIFF
--- a/odxtools/cli/compare.py
+++ b/odxtools/cli/compare.py
@@ -407,45 +407,38 @@ class Comparison(Display):
         # service_dict["changed_name_of_service"][{0 = services, 1 = old service names}][i]
         # service_dict["changed_parameters_of_service"][{0 = services, 1 = changed_parameters, 2 = information_texts}][i]
 
-        # write diagnostic services of each diagnostic layer into a list (type: odxtools.diagservice.DiagService)
-        services_for_dl1 = dl1.services
-        services_for_dl2 = dl2.services
+        dl1_service_names = [service.short_name for service in dl1.services]
 
-        # write request parameters (hex string) refering to Diag-Services into a list (type: bytearray)
-        requests_for_dl1: List[Union[bytes, str]] = []
-        requests_for_dl2: List[Union[bytes, str]] = []
-
-        # method service() fails when request parameters are not coded or have a default value
-        # Error message: odxtools.exceptions.OdxError: The parameters {'foo'} are required but missing!
-        for service in services_for_dl1:
-            assert isinstance(service, DiagService)
-            if not getattr(service.request, "required_parameters", None):
-                requests_for_dl1.append(service())
-            else:
-                requests_for_dl1.append(service.short_name)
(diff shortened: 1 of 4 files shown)


odxtools/cli/compare.py
```diff
@@ -558,6 +560,14 @@ def add_subparser(subparsers: "argparse._SubParsersAction") -> None:
     )
     _parser_utils.add_pdx_argument(parser)
 
+    parser.add_argument(
+        "--no-strict",
+        action="store_true",
+        default=False,
+        required=False,
+        help="Load the dataset in non-strict mode (which is more robust might lead to undefined behavior)",
```
kayoub5, line 568: nitpick: missing "but" between "robust" and "might"

odxtools/cli/compare.py
```diff
@@ -599,6 +609,9 @@ def run(args: argparse.Namespace) -> None:
     task = Comparison()
     task.param_detailed = args.no_details
 
+    if args.no_strict:
+        odxtools.exceptions.strict_mode = False
```
kayoub5, line 613: nitpick: `odxtools.exceptions.strict_mode = not args.no_strict`

odxtools/cli/compare.py
```diff
@@ -718,3 +731,5 @@ def run(args: argparse.Namespace) -> None:
     else:
         # no databases & no variants specified
         print("Please specify either a database or variant for a comparison")
+
+    odxtools.exceptions.strict_mode = True
```
kayoub5, line 735: suggestion: read value then restore it, more fool proof

odxtools/cli/list.py
```diff
@@ -130,6 +132,14 @@ def add_subparser(subparsers: "argparse._SubParsersAction") -> None:
     )
     _parser_utils.add_pdx_argument(parser)
 
+    parser.add_argument(
```
kayoub5, line 135: Add as a global argument, instead of copy pasting in each sub command

odxtools/cli/list.py
```diff
@@ -130,6 +132,14 @@ def add_subparser(subparsers: "argparse._SubParsersAction") -> None:
...
+        "--no-strict",
+        action="store_true",
```
kayoub5, line 137: suggestion: use 
BooleanOptionalAction
see https://stackoverflow.com/a/68837302/1438522

kayoub5 (changes requested):
First three commits look ok and could be merged, last one is more of a feature that merits its own PR

odxtools/cli/list.py
andlaus: as far as I can see this was added in python 3.9, i.e. python 3.8 does not support this? (luckily we're slowly closing in on the EOL of Ubuntu 20.04, when we can raise the minimum python version to 3.10...)

PR #257, opened 2024-01-29
CLI: introduce a non-strict mode to all tools

andlaus:
This is a spin-off of #256: it adds a non-strict mode to all comand line tools whilst the patch that used to be in #256 only dealt with `list` and `compare`. using this mode is more robust regarding incorrect input files, bugs and unimplemented features in odxtools, but it might cause undefined behaviour.

### DIFF
--- a/odxtools/cli/main.py
+++ b/odxtools/cli/main.py
@@ -60,11 +60,10 @@ def start_cli() -> None:
 
     for tool in tool_modules:
         if tool._odxtools_tool_name_ == args.subparser_name:
+            orig_strict = odxtools.exceptions.strict_mode
             odxtools.exceptions.strict_mode = not args.no_strict
             try:
                 tool.run(args)
-            except:
-                raise
             finally:
-                odxtools.exceptions.strict_mode = False
+                odxtools.exceptions.strict_mode = orig_strict
             return
--- a/odxtools/cli/main.py
+++ b/odxtools/cli/main.py
@@ -3,6 +3,8 @@ import argparse
 import importlib
 from typing import Any, List
 
+import odxtools
+
 from ..version import __version__ as odxtools_version
 from .dummy_sub_parser import DummyTool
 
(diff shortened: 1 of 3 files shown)


kayoub5:
suggestion: use 
BooleanOptionalAction
see https://stackoverflow.com/a/68837302/1438522

requires Python 3.9, but we can safely drop support for 3.8 since it reaches end of life October 2024, see https://endoflife.date/python

andlaus:
I'd like to be a bit more conservative here and wait until Ubuntu 20.04 LTS reaches EOL. that should be April/May 2025...

odxtools/cli/main.py
```diff
@@ -50,4 +60,12 @@ def start_cli() -> None:
 
     for tool in tool_modules:
         if tool._odxtools_tool_name_ == args.subparser_name:
-            tool.run(args)
+            orig_strict = odxtools.exceptions.strict_mode
+            odxtools.exceptions.strict_mode = not args.no_strict
+            try:
+                tool.run(args)
+            except:
+                raise
```
kayoub5, line 68: try/finally enough, the except block is not required here

andlaus: that's true if no backtrace ought to be printed if an exception happens to seep out of the tool's `run()` function and if `start_cli()` is guaranteed to never be called from other python code. I think that of these two assertions,  only the latter applies. (I would prefer to be have a fighting chance of finding bugs based on the output of the CLI tools...)

andlaus: (that's due to the fact that so far, no non-standalone callers `start_cli()` exist: `test_cli.py` calls the `run()` functions of the tools directly.)

kayoub5: I meant the code should look like this

```python
            try:
                tool.run(args)
            finally:
                odxtools.exceptions.strict_mode = orig_strict
```

andlaus: okay, my mistake: I somehow assumed that this way exceptions thrown by `tool.run()` get silently ignored whilst in reality they get passed through. (I don't know what made me think that, probably a brain fart :/. anyway: fixed.)

kayoub5 (approved):

Issue #258, opened 2024-01-30
Extra ">" in DIAG-SERVICE jinja


willzhang05:
When the DIAG-SERVICE tag with the TRANSMISSION-MODE attribute is rendered, there appears to be an extraneous ">" character at the end.

![image](https://github.com/mercedes-benz/odxtools/assets/8996428/107024f4-643d-44f5-99c4-a66203d23d2e)

https://github.com/mercedes-benz/odxtools/blob/6a52b4032c497f809ccd391757070bdd46c690d2/odxtools/templates/macros/printService.xml.jinja2#L15-L17

andlaus:
oops, right. can you make a PR?

PR #259, opened 2024-01-30
Remove Extra ">" in DIAG-SERVICE jinja

willzhang05:
resolves https://github.com/mercedes-benz/odxtools/issues/258


andlaus:
thanks for the catch!

Issue #260, opened 2024-01-31
Hardcoded DOCTYPE="CONTAINER" in PARENT-REF tags


willzhang05:
The DOCTYPE="CONTAINER" being hardcoded as an attribute that is included as part of the PARENT-REF tag in the jinja template seems to cause issues with my use case, where I have PARENT-REFs that point to PROTOCOLs.

With this generated XML:
![image](https://github.com/mercedes-benz/odxtools/assets/8996428/6ed9ccfa-150e-4265-a3e7-a095fdbd96e1)

![image](https://github.com/mercedes-benz/odxtools/assets/8996428/581cb397-7ce7-4425-9cc3-2b1897cb32ea)

The inclusion of these two attributes leads to Vector ODXStudio determining that it is a broken link, even if the ID-REF is correct, because I don't believe a PROTOCOL is technically a CONTAINER?

If I eliminate the hardcoded DOCTYPE attribute, I no longer have broken links to the PROTOCOLs.

https://github.com/mercedes-benz/odxtools/blob/6a52b4032c497f809ccd391757070bdd46c690d2/odxtools/templates/macros/printParentRef.xml.jinja2#L8-L9

PR #261, opened 2024-01-31
Address hardcoded DOCTYPE attribute for PARENT-REF

willzhang05:
Addresses https://github.com/mercedes-benz/odxtools/issues/260

I'm not quite sure of the correct way to best handle this, I'm not 100% sure on the spec for PARENT-REF, but looking at the ISO 22901-1 document, I don't see any instances where the DOCTYPE attribute is used with PARENT-REF. Is there somewhere in the spec that lists DOCTYPE as a valid attribute for PARENT-REF?

### DIFF
--- a/odxtools/templates/macros/printParentRef.xml.jinja2
+++ b/odxtools/templates/macros/printParentRef.xml.jinja2
@@ -5,8 +5,6 @@
 
 {%- macro printParentRef(par) -%}
 <PARENT-REF ID-REF="{{par.layer.odx_id.local_id}}"
-            DOCREF="{{par.layer.short_name}}"
-            DOCTYPE="CONTAINER"
             xsi:type="{{par.layer.variant_type.value}}-REF">
 {%- if par.not_inherited_diag_comms %}
  <NOT-INHERITED-DIAG-COMMS>


andlaus:
hm, I don't know about DOCTYPE, but removing the DOCREF will probably make it fail if the referenced parent is in a different DIAG-LAYER-CONTAINER than the variant that uses it (?)

That said, I thought that all variants are contained by `DIAG-LAYER-CONTAINER`s, but maybe this is not the case. what is the nesting of the tags for your protocol? for me, protocols look like
```xml
<ODX>
 <DIAG-LAYER-CONTAINER>
  <PROTOCOLS>
   <PROTOCOL>
    <SHORT-NAME>CAN</SHORT-NAME>
   </PROTOCOL>
  </PROTOCOLS>
 </DIAG-LAYER-CONTAINER>
</ODX>
```

willzhang05:
yes, that's how the tags are nested for me as well. I tried both removing just the DOCREF or just removing the DOCTYPE, but the only one that seemed to be valid for ODXStudio was without both. I'm not sure if this is a Vector quirk, but I'm just having a hard time finding any information about this part of the spec.

andlaus:
in the ODX model, not specifying the DOCREF implies that the referenced object is in the same document fragment (read: `.odx-*` file) as the referencing object. as far as I understand this matter, the DOCREF can be specified even if this is the case, it is just redundant in this case; but maybe I'm wrong here, the wording is a bit ambiguous: 
```
Generally, odxlinks are resolved in two steps:
a) Find the document fragment that contains the link target
b) Retrieve the link target from the document fragment
The first step uses the attributes DOCTYPE and DOCREF:
These two attributes may be defined only jointly. The DOCREF attribute refers to the
SHORT-NAME of the ODX-CATEGORY or the DIAG-LAYER. If DOCREF and DOCTYPE
are not set, the link target exists in the local ODX-CATEGORY where also the link source
exists (internal odxlink).
```

note that this wording neither explicitly allows nor forbits to specify the document fragment for "internal" odxlinks. thus, if candela does  not support this, it might or might not be considered a bug. we could work around this by checking if the referenced diag layer is in the same container as the referencing one and only add these attributes if this is not the case. IMO this would be quite ugly, though...

andlaus:
okay, I played a bit with this code. I noticed that the `DOCTYPE` attribute is not a container, but a layer. does odxstudio accept the file if you change `DOCTYPE="CONTAINER"` to `DOCTYPE="LAYER"`?

willzhang05:
closed in lieu of the solution in https://github.com/mercedes-benz/odxtools/pull/262

PR #262, opened 2024-02-01
use the correct value for the DOCREF in PARENT-REFs

andlaus:
this is supposed to be the short name of the container which contains the referenced layer, not the short name of the layer itself.

this might fix https://github.com/mercedes-benz/odxtools/issues/260. if it does, it is more generic than https://github.com/mercedes-benz/odxtools/pull/261 because with the approach suggested here, the parent of a layer does not necessarily need to be in the same container.

@willzhang05: can you check if the resulting files work with odxstudio for you?


kayoub5 (approved):

willzhang05:
Awesome, this appears have done the job, thanks for the quick turnaround on the fix!
image

PR #264, opened 2024-02-07
Refactor decoding, part 1

andlaus:
This is the first PR of a mini series to refactor the decoding code of odxtools. In this PR, we refactor the `DecodeState` class with the aim of not having to create copies during the decoding process:

- the attributes `cursor_position` and `origin_position` are introduced. The former points to the first undecoded byte whilst the latter points to the byte to which relative positions given by the BYTE-POSITION tag refer to.
- the `parameter_values` property is removed. Instead, the parameter values are kept as local variables in the respective decoding functions.
- `length_keys` and `table_keys` attributes are introduced, to keep track of inter-parameter dependencies

note that the "do not create copies" aim is not yet fully reached after this PR, but it will be at the end of this PR series...

### DIFF
--- a/odxtools/parameters/tablestructparameter.py
+++ b/odxtools/parameters/tablestructparameter.py
@@ -135,13 +135,13 @@ class TableStructParameter(Parameter):
         # find the selected table row
         key_name = self.table_key.short_name
 
-        table = self.table_key.table
-        tr_short_name = decode_state.parameter_values[key_name]
-        candidate_trs = [tr for tr in table.table_rows if tr.short_name == tr_short_name]
-        if len(candidate_trs) != 1:
-            raise EncodeError(f"Could not uniquely resolve a table row named "
-                              f"'{str(tr_short_name)}' in table '{str(table.short_name)}' ")
-        table_row = candidate_trs[0]
+        decode_state.table_keys[key_name]
+        table_row = decode_state.table_keys.get(key_name)
+        if table_row is None:
+            raise odxraise(f"No table key '{key_name}' found when decoding "
+                           f"table struct parameter '{str(self.short_name)}'")
+            dummy_val = cast(str, None), cast(int, None)
+            return dummy_val, decode_state.cursor_position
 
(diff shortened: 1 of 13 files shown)


odxtools/parameters/tablestructparameter.py
```diff
@@ -139,24 +135,23 @@ def decode_from_pdu(self, decode_state: DecodeState) -> Tuple[ParameterValue, in
...
             return (table_row.short_name, val), i
         elif table_row.structure is not None:
-            structure = table_row.structure
-            val, i = structure.convert_bytes_to_physical(decode_state)
+            val, i = table_row.structure.convert_bytes_to_physical(decode_state)
             return (table_row.short_name, val), i
         else:
             # the table row associated with the key neither defines a
-            # DOP not a structure -> ignore it
+            # DOP nor a structure -> ignore it
```
kayoub5, line 156: nor?

tests/test_decoding.py
```diff
@@ -888,7 +888,7 @@ def test_decode_request_end_of_pdu_field(self) -> None:
         diag_layer._resolve_odxlinks(odxlinks)
         diag_layer._finalize_init(odxlinks)
 
-        coded_message = bytes([0x12, 0x34, 0x34])
+        coded_message = bytes([0x12, 0x34, 0x54])
```
kayoub5, line 891: Why?

andlaus: I wanted the two items of the END-OF-PDU fields be different from each other. the second nibble of these bytes is captured by `struct_param_1` which is a MATCHING-REQUEST-PARAM and can not be changed for that reason...

odxtools/parameters/tablestructparameter.py
kayoub5: So it should say "**Neither** DOP nor a a structure"

kayoub5: Just a nitpick here, otherwise LGTM

kayoub5 (approved):

PR #265, opened 2024-02-09
Refactor decoding, part 2

andlaus:
The approach taken in this PR will serve as the template for all other objects which can be decoded (i.e., DOPs and parameters): decoding is done by a `decode_from_pdu()` method which takes a `DecodeState` object as their lone argument and returns the decoded physical value and the passed `DecodeState` is supposed to be changed so that it is set up to decode the next object after calling `decode_from_pdu()`.

IMO, this substantially reduces the mental load required to understand the decoding logic, because before this, parameters, diag coded types, and DOPs all used identical names for the decoding functions  but they behaved slightly different in terms of the parameters which they took and the types they returned...

### DIFF
--- a/odxtools/decodestate.py
+++ b/odxtools/decodestate.py
@@ -17,12 +17,12 @@ class DecodeState:
     #:
     #: i.e., the absolute byte position to which all relative positions
     #: refer to, e.g. the position of the first byte of a structure.
-    origin_position: int = 0
+    origin_byte_position: int = 0
 
     #: Absolute position of the next undecoded byte to be considered
     #:
     #: (if not explicitly specified by the object to be decoded.)
-    cursor_position: int = 0
+    cursor_byte_position: int = 0
 
     #: the bit position [0, 7] where the object to be extracted begins
     #:
--- a/odxtools/decodestate.py
+++ b/odxtools/decodestate.py
@@ -1,6 +1,6 @@
 # SPDX-License-Identifier: MIT
 from dataclasses import dataclass, field
-from typing import TYPE_CHECKING, Dict
+from typing import TYPE_CHECKING, Dict, Optional
 
 if TYPE_CHECKING:
     from .tablerow import TableRow
@@ -23,6 +23,12 @@ class DecodeState:
     #:
     #: (if not explicitly specified by the object to be decoded.)
     cursor_position: int = 0
+
+    #: the bit position [0, 7] where the object to be extracted begins
+    #:
+    #: If bit position is undefined (`None`), the object to be extracted
+    #: starts at bit 0.
+    cursor_bit_position: Optional[int] = None
 
     #: values of the length key parameters decoded so far
     length_keys: Dict[str, int] = field(default_factory=dict)
--- a/odxtools/decodestate.py
+++ b/odxtools/decodestate.py
@@ -1,6 +1,6 @@
 # SPDX-License-Identifier: MIT
 from dataclasses import dataclass, field
-from typing import TYPE_CHECKING, Dict, Optional
+from typing import TYPE_CHECKING, Dict
 
 if TYPE_CHECKING:
     from .tablerow import TableRow
@@ -28,7 +28,7 @@ class DecodeState:
     #:
     #: If bit position is undefined (`None`), the object to be extracted
     #: starts at bit 0.
-    cursor_bit_position: Optional[int] = None
+    cursor_bit_position: int = 0
 
     #: values of the length key parameters decoded so far
     length_keys: Dict[str, int] = field(default_factory=dict)
--- a/odxtools/minmaxlengthtype.py
+++ b/odxtools/minmaxlengthtype.py
@@ -101,11 +101,11 @@ class MinMaxLengthType(DiagCodedType):
         return value_bytes
 
     def decode_from_pdu(self, decode_state: DecodeState, bit_position: int = 0) -> AtomicOdxType:
-        if decode_state.cursor_position + self.min_length > len(decode_state.coded_message):
+        if decode_state.cursor_byte_position + self.min_length > len(decode_state.coded_message):
             raise DecodeError("The PDU ended before minimum length was reached.")
 
         coded_message = decode_state.coded_message
-        cursor_pos = decode_state.cursor_position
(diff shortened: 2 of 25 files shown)


andlaus:
SWF issue: https://issue.swf.i.mercedes-benz.com/servicedesk/customer/portal/101/OSC-803

odxtools/decodestate.py
```diff
@@ -24,6 +24,12 @@ class DecodeState:
     #: (if not explicitly specified by the object to be decoded.)
     cursor_position: int = 0
```
kayoub5, line 25: Maybe add byte there to reduce confusion?
```suggestion
    cursor_byte_position: int = 0
```

odxtools/decodestate.py
```diff
@@ -24,6 +24,12 @@ class DecodeState:
...
 
+    #: the bit position [0, 7] where the object to be extracted begins
+    #:
+    #: If bit position is undefined (`None`), the object to be extracted
+    #: starts at bit 0.
+    cursor_bit_position: Optional[int] = None
```
kayoub5, line 31: I don't see why this can't be 0 by default

odxtools/minmaxlengthtype.py
```diff
@@ -167,9 +168,13 @@ def convert_bytes_to_internal(self,
...
-                bit_position=bit_position,
+                bit_position=0,
                 bit_length=8 * byte_length,
                 base_data_type=self.base_data_type,
                 is_highlow_byte_order=self.is_highlow_byte_order,
             )
-            return value, byte_pos
+
+            decode_state.cursor_position = byte_pos
+            decode_state.cursor_bit_position = None
```
kayoub5, line 178: Maybe add a `goto(byte_pos, bit_pos = 0)` function to `decode_state` ?

tests/test_cli.py
```diff
@@ -130,10 +130,12 @@ def test_compare_tool(self) -> None:
...
         UtilFunctions.run_compare_tool(ecu_variants=["somersault_lazy", "somersault_assiduous"])
 
-    @unittest.skipIf(import_failed, "import of PyInquirer failed")
+    @unittest.skipIf(browse_import_failed, "importing the browse tool failed")
     def test_browse_tool(self) -> None:
         browse_args = Namespace(pdx_file="./examples/somersault.pdx")
-        browse.run(browse_args)
+        with self.assertRaises(SystemError):
+            # browse can only be run interactively
+            browse.run(browse_args)
```
kayoub5, line 138: Maybe mock the interactive part for testing?

odxtools/decodestate.py
andlaus: the idea is that `None` means "not specified" while 0 indicates that it was explicitly set to 0 and thus code where specifying a bit position is not allowed can error out even if 0 was specified. (with this approach, code where bit positions are allowed needs to use `decode_state.cursor_bit_position or 0`, which is not too bad IMO...)

odxtools/decodestate.py
kayoub5: I am just adding `byte` there to be able to tell it apart from the "bit" cursor position, no relation to the `byte_position` attribute is intended.

odxtools/decodestate.py
kayoub5: Any case where None and 0 would have different logic?

andlaus: all parameters which do not accept bit positions, e.g. structures, muxes, value parameters with non-integer values, etc...

odxtools/minmaxlengthtype.py
andlaus: I've tried this as a prototype, but I was not really happy with it because it obfuscated what was happening too much. That said, how do you interpret the following sentence from section 7.3.5.2 of the ASAM version of the spec:  

```
The parameter starts then at the byte edge next to the end of the last parameter.
```

does "last parameter" mean
a) the parameter which is directly before the one currently decoded in the list of parameters
b) the parameter which encompasses the byte of the PDU furthest to the back which has been seen so far
c) the parameter specified last in the list

IMO, (c) does not make sense, we currently use interpretation (b), but the more I think about it, (a) makes more sense (and, as a bonus, is easier to implement)...

odxtools/decodestate.py
andlaus: okay, changed. I'm not happy about the length of the name of this attribute, but I can see your point...

tests/test_cli.py
andlaus: good idea. I tested to implement this, but the assumption that `sys.stdout` is a tty is burried too deeply within PyInquirer. (there are plenty of asserts and probably quite a few implicit assumptions...)

That said, I've noticed that this test stopped to ask for user input when PyInquirer is available and it is run in "bare metal mode" (without using `pytest`), so I mocked `sys.stdout.isatty()` to make it reliably bail out...

odxtools/minmaxlengthtype.py
kayoub5: Both A and B are valid interpretations given how vague the spec is, and would give identical results given sane enough PDX file that does order the parameters by desired position.

That being said, the specs didn't use the term "previous parameter" which is more common, making option B more likely.

odxtools/decodestate.py
kayoub5: All of those either require bit position to be non specified or 0

odxtools/minmaxlengthtype.py
andlaus: okay. would you support switching the interpretation to option (a) in the future anyway? (it simplifies things considerably...)

odxtools/decodestate.py
andlaus: point taken & code changed. I wanted to be able to check if a bit position is set to 0 where it is not allowed to be defined in the decoding logic, but I finally saw the light and concluded that this is the wrong abstraction level to check this...

odxtools/minmaxlengthtype.py
kayoub5: I have not seen a PDX so far, where choosing A or B matters, if I ran into issues I will let you know.

kayoub5 (approved):

PR #267, opened 2024-02-15
Refactor decoding, part 4

andlaus:
This PR is the final part of the decode code refactoring miniseries. It contains the "leftovers" which I've encountered and fixed while implementing the rest of the series and which require the API changes of the rest of the series to be present.

### DIFF
--- a/odxtools/cli/browse.py
+++ b/odxtools/cli/browse.py
@@ -2,7 +2,7 @@
 import argparse
 import logging
 import sys
-from typing import List, Optional, Union
+from typing import List, Optional, Union, cast
 
 from InquirerPy import prompt as PI_prompt
 from tabulate import tabulate  # TODO: switch to rich tables
@@ -99,7 +99,7 @@ def prompt_single_parameter_value(parameter: Parameter) -> Optional[AtomicOdxTyp
         return None
     elif parameter.physical_type.base_data_type is not None:
         return _convert_string_to_odx_type(
-            answer.get(parameter.short_name), parameter.physical_type.base_data_type)
+            cast(str, answer.get(parameter.short_name)), parameter.physical_type.base_data_type)
     else:
         logging.warning(
             f"Parameter {parameter.short_name} does not have a physical data type. Param details: {parameter}"
@@ -148,7 +148,7 @@ def encode_message_interactively(sub_service: Union[Request, Response],
                     lambda input: _convert_string_to_bytes(input),
             }]
             answer = PI_prompt(answered_request_prompt)
-            answered_request = answer.get("request")
+            answered_request = cast(bytes, answer.get("request"))
             print(f"Input interpretation as list: {list(answered_request)}")
 
         # Request values for parameters
@@ -270,7 +270,10 @@ def browse(odxdb: Database) -> None:
         if answer.get("variant") == "[exit]":
             return
 
-        variant = odxdb.diag_layers[answer.get("variant")]
+        variant_name = answer.get("variant")
+        assert isinstance(variant_name, str)
+        variant = odxdb.diag_layers[variant_name]
+        print(f"{type(answer.get('variant'))=}")
         assert isinstance(variant, DiagLayer)
 
         if (rx_id := variant.get_receive_id()) is not None:
@@ -287,7 +290,6 @@ def browse(odxdb: Database) -> None:
             f"{variant.variant_type.value} '{variant.short_name}' (Receive ID: {recv_id}, Send ID: {send_id})"
         )
 
-        service_sn = 0
         while True:
             services: List[DiagService] = [
                 s for s in variant.services if isinstance(s, DiagService)
@@ -307,6 +309,7 @@ def browse(odxdb: Database) -> None:
                 break
 
             service_sn = answer.get("service")
+            assert isinstance(service_sn, str)
 
             service = variant.services[service_sn]
             assert isinstance(service, DiagService)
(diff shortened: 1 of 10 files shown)


odxtools/cli/browse.py
```diff
@@ -99,7 +99,7 @@ def prompt_single_parameter_value(parameter: Parameter) -> Optional[AtomicOdxTyp
         return None
     elif parameter.physical_type.base_data_type is not None:
         return _convert_string_to_odx_type(
-            answer.get(parameter.short_name), parameter.physical_type.base_data_type)
+            cast(str, answer.get(parameter.short_name)), parameter.physical_type.base_data_type)
```
kayoub5, line 102:
```suggestion
            str(answer.get(parameter.short_name)), parameter.physical_type.base_data_type)
```

odxtools/diagcodedtype.py
```diff
@@ -70,7 +70,7 @@ def is_highlow_byte_order(self) -> bool:
         return self.is_highlow_byte_order_raw in [None, True]
 
     @staticmethod
-    def _extract_internal_value(
+    def _extract_atomic_value(
```
kayoub5, line 73: Why?

odxtools/dynamiclengthfield.py
```diff
@@ -56,4 +56,32 @@ def convert_physical_to_bytes(
         raise NotImplementedError()
 
     def decode_from_pdu(self, decode_state: DecodeState) -> ParameterValue:
-        raise NotImplementedError()
+
+        odxassert(not decode_state.cursor_bit_position,
+                  "No bit position can be specified for dynamic length fields!")
+
+        orig_origin = decode_state.origin_byte_position
```
kayoub5, line 63: "DynamicLengthField: implement support for decoding" is a new feature, merits its own PR

odxtools/parameters/matchingrequestparameter.py
```diff
@@ -42,11 +43,13 @@ def decode_from_pdu(self, decode_state: DecodeState) -> ParameterValue:
         if self.byte_position is not None:
             decode_state.cursor_byte_position = decode_state.origin_byte_position + self.byte_position
 
-        byte_position = decode_state.cursor_byte_position
-        bit_position = self.bit_position or 0
-        byte_length = (8 * self.byte_length + bit_position + 7) // 8
-        result = decode_state.coded_message[byte_position:byte_position + byte_length]
-        decode_state.cursor_byte_position += byte_length
+        result, decode_state.cursor_byte_position = DiagCodedType._extract_atomic_value(
```
kayoub5, line 46: Is it just me, or a function that takes `decode_state` and `self` would be easier to understand and re-use?

tests/test_somersault.py
```diff
@@ -259,7 +259,7 @@ def test_code_table_params(self) -> None:
         self.assertEqual(
             decoded_resp_data["last_pos_response"][1]  # type: ignore[index, call-overload]
             ["num_flips_done"],  # type: ignore[index, call-overload]
-            bytes([123]))
+            123)
```
kayoub5, line 262: Why?

odxtools/diagcodedtype.py
andlaus: I wanted a different name: there are/were so many methods with "internal" in their name that I get confused quite  easily. (Also, I think that the new value better conveys the fact that this method returns an `AtomicOdxType` object...)

tests/test_somersault.py
andlaus: mainy to better allow non-byte aligned MatchingRequestParameters, but I think using integers here feels more natural, too... (and since python has arbitrary-sized integers, this does not reduce generality.)

odxtools/parameters/matchingrequestparameter.py
andlaus: this is the lowest level method which IMO should be usable without a `decode_state` and even without a DiagCodedType object. (also, DCTs already have a `decode_from_pdu()` method...)

odxtools/dynamiclengthfield.py
andlaus: yeah, I took the patch out of the PR...

odxtools/cli/browse.py
andlaus: not sure: without the cast mypy has the following complaint:
```
odxtools/cli/browse.py:102: error: Argument 1 to "_convert_string_to_odx_type" has incompatible type "str | bool | list[Any] | None"; expected "str"  [arg-type]
```

I guess it is better to make it shut up and deal with the exception if something goes wrong anyway instead of making it "work" unconditionally?

odxtools/parameters/matchingrequestparameter.py
kayoub5: then why it is in the DiagCodedType class?

andlaus: because usually (all occasions instead of this one), it is called from there.

andlaus: (`_extract_atomic_value()` deals with retrieving a non-byte-aligned object from a blob of binary data, while `decode_from_pdu()` also deals with the "book keeping" of updating `decode_state`.)

kayoub5: maybe put the function in the decode state itself?

kayoub5 (approved):

PR #273, opened 2024-02-29
Refactor limit handling

andlaus:
today, we set out to clean up the limit handling code.  this is surprisingly difficult: the problem is that the LIMIT tag does not specify the data type (integer, float, string, bytefield) which the limit is about but the limit obviously needs to know that type in order to do something. The solution taken here is to pass the type of the limit's value to the `from_et()` method as an extra argument. For multiplexer cases where even the type of the limit's value is not known when the limit must be instantiated, (this is only known after the reference to the switch key has been resolved) the method `Limit.set_value_type()` is added.


odxtools/compumethods/compuscale.py
```diff
@@ -64,4 +74,35 @@ def from_et(et_element: ElementTree.Element, doc_frags: List[OdxDocFragment], *,
...
             compu_inverse_value=compu_inverse_value,
             compu_const=compu_const,
-            compu_rational_coeffs=compu_rational_coeffs)
+            compu_rational_coeffs=compu_rational_coeffs,
+            internal_type=internal_type,
+            physical_type=physical_type)
+
+    def applies(self, internal_value: AtomicOdxType) -> bool:
+
+        if self.lower_limit is None and self.upper_limit is None:
```
kayoub5, line 83: limits are required?

kayoub5:
Does the limit really need to know its type during creation?

The datatype can be simply passed as argument to `complies_to_upper` and `complies_to_lower`

odxtools/compumethods/compuscale.py
andlaus: the XSD says that they are optional for COMPU-SCALE. (this is part of why handling of limits is so awkward...)

andlaus:
> Does the limit really need to know its data type during creation?

not during creation, but as soon as it must check if a value is within the specified limit.

> The datatype can be simply passed as argument to complies_to_upper and complies_to_lower

I considered doing that. It turns out that this implies that the limit value must be converted from string each and every time which a limit is checked (and the code felt quite clunky and "footgunny" because one can easily pass the wrong type to these methods...)

odxtools/compumethods/compuscale.py
kayoub5: this is the compu scale class... am I missing something?

andlaus: ah, okay. I suppose that I misread your point: this is supposed to be True?! (did I ever mention that limit handling is surprisingly difficult? anyway: fixed

andlaus:
@kayoub5: Do you think that this can be merged?

kayoub5 (approved):

PR #274, opened 2024-03-01
Implement static fields

andlaus:
This PR implements static fields, i.e., arrays of structures of fixed size. It includes parsing and writing, en- and decoding as well as a unit test.

Besides this, the PR does some janitorial work for `DiagDataDictionarySpec`: It adds the missing `admin_data` attribute and it reorders all member attributes to the same order as the corresponding subtags in the XSD.


kayoub5:
Please split the two commits in two PRs, it's too much to review changes from both at the same time

odxtools/dynamiclengthfield.py
```diff
@@ -107,11 +108,11 @@ def decode_from_pdu(self, decode_state: DecodeState) -> ParameterValue:
...
-        else:
-            decode_state.cursor_byte_position = decode_state.origin_byte_position + self.offset
-            result: List[ParameterValue] = []
-            for _ in range(n):
-                result.append(self.structure.decode_from_pdu(decode_state))
+            n = 0
+
+        decode_state.cursor_byte_position = decode_state.origin_byte_position + self.offset
+        for _ in range(n):
+            result.append(self.structure.decode_from_pdu(decode_state))
```
kayoub5, line 115: Why is there changes for dynamic length field here?

odxtools/diagdatadictionaryspec.py
```diff
@@ -115,25 +122,13 @@ def from_et(et_element: ElementTree.Element,
         else:
             unit_spec = None
 
-        # TODO: Parse different specs.. Which of them are needed?
-        for (path, name) in [
-            ("STATIC-FIELDS", "static fields"),
-            ("DYNAMIC-LENGTH-FIELDS/DYNAMIC-LENGTH-FIELD", "dynamic length fields"),
-            (
-                "DYNAMIC-ENDMARKER-FIELDS/DYNAMIC-ENDMARKER-FIELD",
```
kayoub5, line 123: `DYNAMIC-ENDMARKER-FIELD` is now supported?

odxtools/dynamiclengthfield.py
andlaus: because I've took the dynamic length field code as a template and I noticed that it was hard to read. (at runtime the new code should be equivalent to the old one.)

odxtools/diagdatadictionaryspec.py
andlaus: nope, but it will be soon and this message will be not seen by too many people?!

andlaus:
> Please split the two commits in two PRs, it's too much to review changes from both at the same time

done: #275. Thanks for the nudge.

kayoub5 (approved):

PR #277, opened 2024-03-04
Make more linters happy

andlaus:
This fixes some complaints brought up by `flake8`, `pytype` and `pyright`. Note that I don't intend to run these linters as part of the CI, because it would make contributing even harder than it already is and there are some issues with `flake8` and `pytype`:

- `flake8` likes to play "style police" and thus requires *a ton* of parameters (`--ignore E124,E125,E126,E131,E501,F541,W503,W504`), which together with the fact that it cannot (easily) be configured via `pyproject.toml`  makes it quite cumbersome.
- `pytype` is glacially slow and produces a few false positives which I could not fix
- `pyright` fights with `mypy` about the import of `InquirerPy.prompt` in the browse tool, so it draws the shorter stick...

### DIFF
--- a/odxtools/cli/browse.py
+++ b/odxtools/cli/browse.py
@@ -4,7 +4,7 @@ import logging
 import sys
 from typing import List, Optional, Union, cast
 
-import InquirerPy.prompt as PI_prompt
+import InquirerPy.prompt as IP_prompt
 from tabulate import tabulate  # TODO: switch to rich tables
 
 from ..database import Database
@@ -20,7 +20,7 @@ from ..parameters.valueparameter import ValueParameter
 from ..request import Request
 from ..response import Response
 from . import _parser_utils
-from ._parser_utils import TSubparsersAction
+from ._parser_utils import SubparsersList
 from ._print_utils import extract_parameter_tabulation_data
 
 # name of the tool
@@ -89,13 +89,15 @@ def prompt_single_parameter_value(parameter: Parameter) -> Optional[AtomicOdxTyp
         # else _convert_string_to_odx_type(x, p.physical_type.base_data_type, param=p) # This does not work because the next parameter to be promted is used (for some reason?)
     }]
 
-    if hasattr(parameter, "dop") and hasattr(parameter.dop, "compu_method") \
-       and hasattr(parameter.dop.compu_method, "get_scales"):
-        scales = parameter.dop.compu_method.get_scales()
+    if (dop := getattr(parameter, "dop", None)) and \
+       (compu_method := getattr(dop, "compu_method", None)):
+        scales = compu_method.internal_to_phys
         choices = [scale.compu_const for scale in scales if scale is not None]
+        if (cdv := compu_method.compu_default_value) is not None:
+            choices.append(cdv.compu_const)
         param_prompt[0]["choices"] = choices
 
-    answer = PI_prompt(param_prompt)
+    answer = IP_prompt(param_prompt)
     if answer.get(parameter.short_name) == "" and not parameter.is_required:
         return None
     elif parameter.physical_type.base_data_type is not None:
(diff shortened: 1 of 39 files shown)


kayoub5:
@andlaus I would recommend dropping flake8 in favor of https://github.com/astral-sh/ruff

odxtools/cli/browse.py
```diff
@@ -4,7 +4,7 @@
 import sys
 from typing import List, Optional, Union, cast
 
-import InquirerPy.prompt as PI_prompt
+import InquirerPy.prompt as IP_prompt
```
kayoub5, line 7: What? are you sure?

odxtools/comparaminstance.py
```diff
@@ -131,7 +131,7 @@ def get_subvalue(self, subparam_name: str) -> Optional[str]:
 
         result = value_list[idx]
         if result is None:
-            result = comparam_spec.subparams[idx].physical_default_value
+            result = getattr(comparam_spec.subparams[idx], "physical_default_value", None)
```
kayoub5, line 134: Why?

andlaus:
thanks for the hint, I'll have a look. Be aware, though that I only intend to run these linters every blue moon, and only manually..

odxtools/compumethods/createanycompumethod.py
```diff
@@ -184,6 +180,7 @@ def create_any_compu_method_from_et(et_element: ElementTree.Element,
         return TabIntpCompuMethod(
             internal_points=internal_points, physical_points=physical_points, **kwargs)
 
-    # TODO: Implement other categories (never instantiate CompuMethod)
-    logger.warning(f"Warning: Computation category {compu_category} is not implemented!")
+    # TODO: Implement all categories (never instantiate the CompuMethod base class!)
+    odxraise(f"Warning: Computation category {compu_category} is not implemented!")
```
kayoub5, line 184: Blow the whole thing with odxraise since something is not implemented?

andlaus:
BTW: we already use `ruff` as part of the CI system ;)

odxtools/cli/browse.py
andlaus: the history of this is that in its current state `pyright`, complains about this:
```
$ pyright
/home/and/src/odxtools/odxtools/cli/browse.py
  /home/and/src/odxtools/odxtools/cli/browse.py:7:8 - error: Import "InquirerPy.prompt" could not be resolved (reportMissingImports)
  /home/and/src/odxtools/odxtools/cli/browse.py:100:14 - error: Module is not callable (reportCallIssue)
[...]
```

while if this issue is fixed by replacing the line with `from InquirerPy import prompt as IP_prompt  # type: ignore[attr-defined]`, mypy complains about the `# type: ignore` comment being unnecessary (at least on the GH actions runner, locally it is fine on my machine). the `PI` to `IP` rename is a residual of that, because I thought that if I have to change that line anyway, I could use a name that makes sense...

odxtools/comparaminstance.py
andlaus: pyright complains:
```
$ pyright .
/home/and/src/odxtools/odxtools/comparaminstance.py
  /home/and/src/odxtools/odxtools/comparaminstance.py:134:51 - error: Cannot access member "physical_default_value" for type "BaseComparam"
    Member "physical_default_value" is unknown (reportAttributeAccessIssue)
```

odxtools/compumethods/createanycompumethod.py
andlaus: yes. the idea is that if you would like to try anyway, you should give the non-strict mode a shot.

odxtools/comparaminstance.py
kayoub5: He is right, you need `isinstance`

kayoub5 (approved):

PR #278, opened 2024-03-05
Add `PositionableParam`

andlaus:
This is the base class of all parameters for which a position within the PDU can be defined. Interestingly, the XSD does not seem to feature *any* parameters that are not positionable, but let's reflect the class hierarchy implied by the ODX specification...

### DIFF
--- a/odxtools/parameters/parameter.py
+++ b/odxtools/parameters/parameter.py
@@ -3,6 +3,8 @@ import abc
 from dataclasses import dataclass
 from typing import TYPE_CHECKING, Any, Dict, List, Literal, Optional
 
+from typing_extensions import final
+
 from ..decodestate import DecodeState
 from ..element import NamedElement
 from ..encodestate import EncodeState
@@ -31,12 +33,13 @@ ParameterType = Literal[
 
 @dataclass
 class Parameter(NamedElement, abc.ABC):
-    """This class corresponds to the POSITIONABLE-PARAMETER type of
-    the ODX specification.
+    """This class corresponds to POSITIONABLE-PARAM in the ODX
+    specification.
 
     Be aware that, even though the ODX specification seems to make the
     distinction of "positionable" and "normal" parameters, it does not
     define any non-positionable parameter types.
+
     """
     byte_position: Optional[int]
     bit_position: Optional[int]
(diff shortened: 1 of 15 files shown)


odxtools/cli/_print_utils.py
```diff
@@ -153,7 +153,7 @@ def extract_parameter_tabulation_data(parameters: List[Parameter]) -> Dict[str,
 
     for param in parameters:
         name.append(param.short_name)
-        byte.append(param.byte_position)
+        byte.append(getattr(param, "byte_position", None))
```
kayoub5, line 156: using getattr breaks type checking, and penalize performance

kayoub5:
I simply do not see the benefit of this PR, if non-positioned parameters simply don't exist, why introduce the complexity of a non-necessary abstraction layer?

odxtools/cli/_print_utils.py
andlaus: I buy the type checking argument, but performance wise it should be basically the same because in python classes are implemented as glorified dictionaries.

How can this be fixed alternatively? according to the ODX spec we cannot rely on the fact that all parameters have a `byte_position` attribute (even though in practice we can).

andlaus:
> I simply do not see the benefit of this PR, if non-positioned parameters simply don't exist, why introduce the complexity of a non-necessary abstraction layer?

mainly to make it easier to look at the XSD and find the odxtools code. But I agree, we should probably solve this "by documentation" (i.e., add a comment that `Parameter` implements the `POSITIONABLE-PARAM` XSD group), and/or simply rename `Parameter` to `PositionableParameter`?

kayoub5:
> > I simply do not see the benefit of this PR, if non-positioned parameters simply don't exist, why introduce the complexity of a non-necessary abstraction layer?
> 
> mainly to make it easier to look at the XSD and find the corresponding odxtools code. But I agree, we should probably solve this "by documentation" (i.e., add a comment that `Parameter` implements the `POSITIONABLE-PARAM` XSD group), and/or simply rename `Parameter` to `PositionableParameter`?

A documentation change should be more than enough

andlaus:
okay. I moved all the "meat" of this PR into `Parameter`. (I'd like to keep the centralized position handling for the decoding routines.)

odxtools/parameters/parameter.py
```diff
@@ -90,35 +100,27 @@ def get_coded_value_as_bytes(self, encode_state: EncodeState) -> bytes:
...
 
-        Returns
-        -------
-        ParameterValuePair | List[ParameterValuePair]
-            the decoded parameter value (the type is defined by the DOP)
-        int
-            the next byte position after the extracted parameter
-        """
-        pass
+    def _decode_positioned_from_pdu(self, decode_state: DecodeState) -> ParameterValue:
```
kayoub5, line 118: Add `@abstractmethod`

kayoub5: Other than this, LGTM

andlaus: foreshadowing: abstract base classes prevent using the `from_et()` pattern for class hierarchies because for this pattern, base classes must be instantiated in the `.from_et()` constructor of child classes...

PR #279, opened 2024-03-07
Convert the parameter classes to the `.from_et()` pattern

andlaus:
IMO this cleans up the parameter construction code significantly. On the flipside it requires to make the `Parameter` class instantiable...


odxtools/parameters/createanyparameter.py
```diff
@@ -27,205 +23,33 @@
...
-        dop_snref = None
-        if (dop_snref_elem := et_element.find("DOP-SNREF")) is not None:
-            dop_snref = odxrequire(dop_snref_elem.get("SHORT-NAME"))
-
-        if dop_ref is None and dop_snref is None:
-            raise ValueError(
-                f"A parameter of type {parameter_type} must reference a DOP! {dop_ref}, {dop_snref}"
-            )
-
     if parameter_type == "VALUE":
```
kayoub5, line 29: instead of multiple ifs, maybe use a lookup dict?

kayoub5: other than this, LGTM

andlaus: that would make things harder to read and I doubt that the performance would be measurably better...

andlaus:
ok, let's merge this. Thanks for the review @kayoub5!

PR #282, opened 2024-03-28
Create somersault_modified.pdx dynamically

andlaus:
Instead of creating this file from scratch, it takes the "normal" somersault ECU and modifies it programmatically. (This script thus serves as an example of how this can be done.)

The reason for this approach is to automatically carry over changes to the base `sommersaultecu`, thus simplifying keeping these files in sync and having a well-defined set of modifications.


andlaus:
@kayoub5: any opinion on this? I think that this one makes a nice template for copy-and-pasting but the code is relatively uncritical...

examples/mksomersaultmodifiedpdx.py
```diff
@@ -0,0 +1,230 @@
...
+import odxtools.uds as uds
+from examples import somersaultecu
+from odxtools.diaglayer import DiagLayer
+from odxtools.diagservice import DiagService
+from odxtools.nameditemlist import NamedItemList
+from odxtools.odxlink import OdxLinkId, OdxLinkRef
+from odxtools.parameters.codedconstparameter import CodedConstParameter
+from odxtools.parameters.valueparameter import ValueParameter
+from odxtools.request import Request
+from odxtools.response import Response, ResponseType
```
kayoub5, line 21: Looking at those imports, it reminds that we should expose those classes in the top level module, otherwise it gets hard to figure out where each class comes from

examples/mksomersaultmodifiedpdx.py
```diff
@@ -0,0 +1,230 @@
...
+
+T = TypeVar("T")
+
+
+def find_named_object(item_list: List[T], name: str) -> T:
```
kayoub5, line 26: if you really need such a function here, it means the current api is limited

examples/mksomersaultmodifiedpdx.py
```diff
@@ -0,0 +1,230 @@
...
+    formatter_class=argparse.RawTextHelpFormatter,
+)
+
+argparser.add_argument(
+    "output_pdx_file",
+    metavar="OUTPUT_PDX_FILE",
+    help="Path to the where the resulting .pdx file is written",
+)
+
+args = argparser.parse_args()
```
kayoub5, line 49: add `if __name_ == '__main__' ` check

examples/mksomersaultmodifiedpdx.py
```diff
@@ -0,0 +1,230 @@
...
+new_params = [x for x in pr.parameters if getattr(x, "short_name", None) != "sault_time"]
+pr.parameters = NamedItemList(new_params)
+
+# add a "flic-flac" service
+flic_flac_request = Request(
+    odx_id=OdxLinkId("somersault.RQ.flic_flac", doc_frags),
+    short_name="RQ_flic_flac",
+    long_name=None,
+    description=None,
+    admin_data=None,
```
kayoub5, line 121: we need a better object creation api, that don't require providing optional properties

examples/somersaultecu.py
```diff
@@ -767,7 +767,7 @@ class SomersaultSID(IntEnum):
                     long_name=None,
                     semantic=None,
                     description=None,
-                    request_byte_position=3,
+                    request_byte_position=0,
```
kayoub5, line 770: why?

kayoub5 (changes requested):

examples/mksomersaultmodifiedpdx.py
andlaus: I'm not opposed, but I don't see how this can be done without subscribing to a maintenance nightmare: As far as I'm aware, the only way to do this is to explicitly import  each and every class in the toplevel `__init__.py`?

examples/mksomersaultmodifiedpdx.py
andlaus: IMO this is pretty much a corner case that is only relevant when modifying database objects programmatically: If `item_list` is an instance of `NamedItemList`, this is basically equivalent to `item_list[name]`. The problem here is that the `.diag_comms` member of `DiagLayerRaw` is not a `NamedItemList` because it may also contain references (which do not exhibit a `.short_name` attribute)...

examples/mksomersaultmodifiedpdx.py
andlaus: I agree. the problem is that using dataclasses with defaults does not want play with inheritance...

examples/somersaultecu.py
andlaus: because using byte position 3 is outside of the constant prefix of the request and this breaks the [`._prefix_tree`](https://github.com/mercedes-benz/odxtools/blob/main/odxtools/diaglayer.py#L1030) code if the `MatchingRequest` parameter is made to be strict in the sense that it must encode something. (I have a patch for the latter and I did not want to open the can of worms of thinking about how to represent non-existing matching request data during encoding. also, I think this is a nice demonstration why one might want to modify this database dynamically instead of just copying `somersaultecu.py` and fiddling with the copy...)

examples/mksomersaultmodifiedpdx.py
andlaus: fixed.

examples/mksomersaultmodifiedpdx.py
kayoub5: You are already doing that in `examples/somersaultecu.py`, you will just be moving imports from one file to another.

examples/mksomersaultmodifiedpdx.py
kayoub5: How so?

examples/mksomersaultmodifiedpdx.py
andlaus: in `__init__.py` there would be the implicit promise that the imports are complete whilst files containing code only import what they actually use (in fact, this is even enforced by `ruff`...)

kayoub5: Maybe create issues ticket? so it does not get lost in this PR review.

examples/mksomersaultmodifiedpdx.py
kayoub5: Migrate to 3.10 and use `kw_only`

examples/mksomersaultmodifiedpdx.py
andlaus: good point. done: https://github.com/mercedes-benz/odxtools/issues/287

examples/mksomersaultmodifiedpdx.py
andlaus: I opened an issue to discuss this and to keep track of stuff which we should change after the transition: https://github.com/mercedes-benz/odxtools/issues/286

kayoub5 (approved):

Issue #200, opened 2023-09-06
incorrect handling of ref with target in multiple odx files


HerrMedizins:
The inheritance of the request in ODX2 works fine. But the DOP beeing refered to in the request in ODX2 exists in both ODX files in different versions. From my understanding, and the way other odx-interpreting tools handle this situation is, that the DOP of ODX1 is used, and the DOP of ODX2 is ignored/overwritten. 
Odxtools though takes the DOP from ODX2 and ignores/overwrites the DOP from ODX1, which leads to the data of the DOP in ODX1 beeing lost.
![image](https://github.com/mercedes-benz/odxtools/assets/144232818/d637470f-a2af-4e68-b25c-5b1c72a52160)
![image](https://github.com/mercedes-benz/odxtools/assets/144232818/2f05fcb6-edcf-480c-b881-8d1adb0c130b)

andlaus:
> Odxtools though takes the DOP from ODX2 and ignores/overwrites the DOP from ODX1, which leads to the data of the DOP in ODX1 beeing lost.

I'm not sure whether I understand your issue correctly: for ECUs of ODX2 which specify a DOP with the same name, overriding the one from ODX1 it is correct as far as I know. The original DOP still ought to be exposed in ECUs which don't specify a new one and if the DOP from ODX1 can still be referenced via ODXLINK by any ECU (i.e., even the ECUs which override it), but if the ECU does not reside in ODX1, the `DOCREF="ODX1"` attribute needs to be specified by the `DOP-REF`.

But maybe I misunderstood and you mean to "extend" the data of the DOP from ODX1 with the data specified by ODX2; As far as I know, the ODX standard does not allow this, as this would easily lead to breaking the data model as some tags are mutually exclusive and there is no way of "deleting" an arbitrary inherited one.

HerrMedizins:
I'll try another approach to explain my problem.
All ODXs im mentioning belong to the same ECU. ODX1 inherits from another ODX which in turn inherits from ODX2. So what happens is, that everything from ODX2 which is the parent ODX in this case ,is inherited, and then overwritten with the parts  existing in ODX1, as expected. If i now with python parse in the pdx with all the ODXs i get one ECU Object as expected. If i now look into the data_dictionary_spec -> data_object_props i can find the DOP DOP_REF (from images) filled correctly with the data from ODX1. So far all works as expected.
[...]

PR #283, opened 2024-03-31
dop_snref Temporary plan

QWander:
In section 7.3.2.5 "Importing and referencing of cobjects",the ability to cross-reference shortnames using "dop_snref" is
odefined.
dop_snref can reference the content of other files, but the all_data_object_properties here do not define the entire content, which may lead to errors. 
The change here temporarily solves this problem (skipping the use of dop_snref that references other files). This is necessary, otherwise it will trigger odxraise.
Long term change direction: I tend to add the content of the import file when generating all_data_object_properties, rather than processing it in the resolve_snrefs section.

### DIFF
--- a/odxtools/database.py
+++ b/odxtools/database.py
@@ -128,7 +128,7 @@ class Database:
         # let the diaglayers sort out the inherited objects and the
         # short name references
         for dlc in self.diag_layer_containers:
-            dlc._finalize_init(self._odxlinks)
+            dlc._finalize_init(self._odxlinks, self._ecu_shared_datas)
 
     @property
     def odxlinks(self) -> OdxLinkDatabase:
--- a/odxtools/database.py
+++ b/odxtools/database.py
@@ -92,6 +92,9 @@ class Database:
 
     def refresh(self) -> None:
         # Create wrapper objects
+        self._ecu_shared_datas = NamedItemList(chain(*[dlc.ecu_shared_datas
+            for dlc in self.diag_layer_containers if dlc.ecu_shared_datas]))
+
         self._diag_layers = NamedItemList(
             chain(*[dlc.diag_layers for dlc in self.diag_layer_containers]))
 
@@ -131,6 +134,13 @@ class Database:
     def odxlinks(self) -> OdxLinkDatabase:
         """A map from odx_id to object"""
         return self._odxlinks
+
+    @property
+    def ecu_shared_datas(self) -> OdxLinkDatabase:
+        """
+        All ecu_shared_datas defined by this database
+        """
+        return self._ecu_shared_datas
 
     @property
     def protocols(self) -> NamedItemList[DiagLayer]:
--- a/odxtools/diaglayer.py
+++ b/odxtools/diaglayer.py
@@ -78,7 +78,8 @@ class DiagLayer:
 
         self.diag_layer_raw._resolve_odxlinks(odxlinks)
 
-    def _finalize_init(self, odxlinks: OdxLinkDatabase) -> None:
+    def _finalize_init(self, odxlinks: OdxLinkDatabase,
+            ecu_shared_datas: NamedItemList['DiagLayer']) -> None:
         """This method deals with everything inheritance related and
         -- after the final set of objects covered by the diagnostic
         layer is determined -- resolves any short name references in
@@ -209,6 +210,8 @@ class DiagLayer:
             unit_spec=unit_spec,
             sdgs=ddds_sdgs,
         )
+
+        self._add_import_ref(ecu_shared_datas)
 
         #####
         # compute the communication parameters applicable to the
@@ -590,6 +593,27 @@ class DiagLayer:
             return []
 
         return self._compute_available_objects(get_local_objects_fn, not_inherited_fn)
+
+    def _add_import_ref(self, ecu_shared_datas) ->  None:
+        if imp_refs := self.import_refs:
+            for imp in imp_refs:
+                doc_name = next(iter(imp.ref_docs)).doc_name
+                doc_type = next(iter(imp.ref_docs)).doc_type
+                # Delete the “DLC_” name prefix
+                if doc_type == 'CONTAINER':
+                    doc_name = doc_name[4:]
+                if doc_name in ecu_shared_datas:
+                    share_ddds = (ecu_shared_datas[doc_name]
+                        .diag_layer_raw.diag_data_dictionary_spec)
+                    if hasattr(share_ddds, 'all_data_object_properties'):
+                        share_ddds_all_properties = share_ddds.all_data_object_properties
+                    # diag_data_dictionary_spec
+                    if ddds_all_properties := (self.diag_layer_raw.diag_data_dictionary_spec
+                        .all_data_object_properties):
+                        ddds_all_properties.append_list(share_ddds_all_properties)
+                else:
+                    odxraise(f"The imports-refs method used in {self.short_name} is error,"
+                             f"file {doc_name} is undefined")
 
     #####
     # </value inheritance mechanism helpers>
--- a/odxtools/diaglayercontainer.py
+++ b/odxtools/diaglayercontainer.py
@@ -124,17 +124,18 @@ class DiagLayerContainer(IdentifiableElement):
         for ecu_variant in self.ecu_variants:
             ecu_variant._resolve_odxlinks(odxlinks)
 
-    def _finalize_init(self, odxlinks: OdxLinkDatabase) -> None:
+    def _finalize_init(self, odxlinks: OdxLinkDatabase,
+            ecu_shared_datas: NamedItemList[DiagLayer]) -> None:
         for ecu_shared_data in self.ecu_shared_datas:
-            ecu_shared_data._finalize_init(odxlinks)
+            ecu_shared_data._finalize_init(odxlinks, ecu_shared_datas)
         for protocol in self.protocols:
-            protocol._finalize_init(odxlinks)
+            protocol._finalize_init(odxlinks, ecu_shared_datas)
         for functional_group in self.functional_groups:
-            functional_group._finalize_init(odxlinks)
+            functional_group._finalize_init(odxlinks, ecu_shared_datas)
         for base_variant in self.base_variants:
-            base_variant._finalize_init(odxlinks)
+            base_variant._finalize_init(odxlinks, ecu_shared_datas)
         for ecu_variant in self.ecu_variants:
-            ecu_variant._finalize_init(odxlinks)
+            ecu_variant._finalize_init(odxlinks, ecu_shared_datas)
 
     @property
     def diag_layers(self) -> NamedItemList[DiagLayer]:
--- a/odxtools/nameditemlist.py
+++ b/odxtools/nameditemlist.py
@@ -55,6 +55,14 @@ class ItemAttributeList(List[T]):
 
         super().append(item)
 
+    def append_list(self, item_list) -> None:
+        if item_list:
+            for item in item_list:
+                item_name = self._get_item_key(item)
+                if item_name not in self._item_dict:
+                    self._item_dict[item_name] = item
+                    super().append()
+
     def _add_attribute_item(self, item: T) -> None:
         item_name = self._get_item_key(item)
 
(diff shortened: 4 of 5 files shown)


CLA-Mercedes-Benz:
[CLA assistant check](https://cla.opensource.mercedes-benz.com/mercedes-benz/odxtools?pullRequest=283) <br/>Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.opensource.mercedes-benz.com/mercedes-benz/odxtools?pullRequest=283) before we can accept your contribution.<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.opensource.mercedes-benz.com/check/mercedes-benz/odxtools?pullRequest=283) it.</sub>

andlaus:
To be honest, I don't really get what the issue is here: Shouldn't the `DiagDataDictionarySpec.all_data_object_properties` be extended to include the missing objects?

on a different note: the semantics of short name references are pretty awful because the object these references point to depend on the calling object, i.e., the only way to implement them "correctly" would be do  do short name lookups each and every time such an referenced is accessed. since this is slow, and is hopefully not very common, odxtools currently does pre-lookups and relies some hacks to bend the whole database towards the ECU of interest if this issue is present, cf. #200.

andlaus:
Even after reading section 7.3.5.2, I still don't get it: is the issue that your files use IMPORT-REF in addition to PARENT-REF to define the hierarchy of ECUs? (IMPORT-REF is currently not implemented, but I can have a look.)

QWander:
I provide two solutions to this question:
**Solution 1:**
After the initialization of DiagDataDictionarySpec._all_data_object_properties, add a step to append the DiagDataDictionarySpec.all_data_object_properties from the target file for ECUs defined using <IMPORT-REF>. Analogously, ADMIN-DATA, COMPANY-DATAS, etc., may need to be updated accordingly.
**Solution 2:**
Enhance all _resolve_snrefs() functions by adding Database.diag_layer_containers as an input parameter to handle ECUs defined using <IMPORT-REF> appropriately.

andlaus:
> Solution 1:
> After the initialization of DiagDataDictionarySpec._all_data_object_properties, add a step to append the DiagDataDictionarySpec.all_data_object_properties from the target file for ECUs defined using . Analogously, ADMIN-DATA, COMPANY-DATAS, etc., may need to be updated accordingly.

I'm pretty sure that all ODX files that are contained within PDX archive files are already loaded. maybe there are some bugs, but IMO these should be fixed then. The only thing which I can come up with that could be wrong in your case is your files using the IMPORT-REF mechanism (which is a massive pain, because if I understood this correctly it is not a "normal" inheritance mechanism but only imports those objects from the base layer that get referenced by the derived layer...). Would be nice if you could check this...

QWander:
Apart from changes to the coding standard and test cases,my local test have already passed.Please help me check if it is feasible. @andlaus

andlaus:
my main issue with this PR is that I cannot see how it is connected to what's described in the ODX specification, in particular that you do a lot of stuff concerned with SNREFs, while the spec explicitly diallows this (cf. section 7.2.3.5, bullet point 2: `[...] It is not possible to import a data object by a SHORT-NAME reference only.`). I've implemented my interstanding of the spec in #285. would be nice if you gave it a whirl...

QWander:
Okay, I will try it out soon, thank you.

QWander:
The solution you provided was not successful. @andlaus 
`odxtools\parameters\tablestructparameter.py:68: OdxWarning: Table keys cannot yet be defined using SNREFs in TableStructParameters.
  warnings.warn(
  File "odxedit.py", line 158, in <module>
    b = PdxTest(pdx)
  File "odxedit.py", line 33, in __init__ 
    self.db = odxtools.load_pdx_file(pax_path)
  File "odxtools\load_pdx_file.py", line 8, in load_pdx_file
    return Database(pdx_zip=ZipFile(pdx_file))
  File "odxtools\database.py", line 91, in __init__
    self.refresh()
  File "odxtools\database.py", line 128, in refresh
    dlc._finalize_init(self._odxlinks)
  File "odxtools\diaglayercontainer.py", line 129, in _finalize_init
    ecu_shared_data._finalize_init(odxlinks)
  File "odxtools\diaglayer.py", line 273, in _finalize_init
    self.diag_layer_raw._resolve_snrefs(self)
  File "odxtools\diaglayerraw.py", line 291, in _resolve_snrefs
    self.diag_data_dictionary_spec._resolve_snrefs(diag_layer)
  File "odxtools\diagdatadictionaryspec.py", line 232, in _resolve_snrefs
    structure._resolve_snrefs(diag_layer)
  File "odxtools\basicstructure.py", line 319, in _resolve_snrefs
    param._resolve_snrefs(diag_layer)
  File "odxtools\parameters\valueparameter.py", line 54, in _resolve_snrefs
  File "odxtools\parameters\parameterwithdop.py", line 69, in _resolve_snrefs
    self._dop = odxrequire(ddds.all_data_object_properties.get(self.dop_snref))
  File "odxtools\exceptions.py", line 76, in odxrequire
    odxraise(message)
  File "odxtools\exceptions.py", line 39, in odxraise
    raise error_type()
odxtools.exceptions.OdxError`

QWander:
I recommend my solution again:
An attribute named ecu_shared_datas has been added to the Database, which is used to define the common parts of the ODX implementation, such as when being referenced by <IMPORT-REF>;
In the DiagLayer._finalize_init function, Database.ecu_shared_datas has been added as an input parameter. When using <IMPORT-REF>, the corresponding ecu_shared_datas are imported into the reference file. In this way, I have implemented the addition of the imported file's part in DiagDataDictionarySpec._all_data_object_properties;
Thus, during the _resolve_snrefs process, all_data_object_properties represents both the properties of the ECU file and the properties of the referenced part. This effectively implements <IMPORT-REF>.

andlaus:
> `odxtools\parameters\tablestructparameter.py:68: OdxWarning: Table keys cannot yet be defined using SNREFs in TableStructParameters.`

your problem actually is completely unrelated to the ODXLINK mechanism: As the exception says, your file tries to specify a table key for a table struct parameter via a short name reference. I will have a look, in the meantime you can probably use the [non-strict mode](https://github.com/mercedes-benz/odxtools#using-the-non-strict-mode) as long as you do not try to en- or decode a request or response which contains that parameter...

QWander:
> > `odxtools\parameters\tablestructparameter.py:68: OdxWarning: Table keys cannot yet be defined using SNREFs in TableStructParameters.`
> 
> your problem actually is completely unrelated to the IMPORT-REF mechanism: As the exception says, your file tries to specify a table key for a table struct parameter via a short name reference. I will have a look, in the meantime you can probably use the [non-strict mode](https://github.com/mercedes-benz/odxtools#using-the-non-strict-mode) as long as you do not try to en- or decode a request or response which contains that parameter...


File A references file B through the IMPORT-REF method, and then using the short name from file B in file A will cause problems.

odxtools/parameters/parameterwithdop.py
```diff
@@ -66,7 +66,10 @@ def _resolve_snrefs(self, diag_layer: "DiagLayer") -> None:
 
         if self.dop_snref:
             ddds = diag_layer.diag_data_dictionary_spec
-            self._dop = odxrequire(ddds.all_data_object_properties.get(self.dop_snref))
+            if snref_shortname := ddds.all_data_object_properties.get(self.dop_snref):
```
andlaus, line 69: That `if` statment is equivalent to simply
```python
self._dop = ddds.all_data_object_properties.get(self.dop_snref)
```

that said, I think the `odxrequire()` should stay because the reference to the DOP must be resolved. If it isn't (most probably because the referenced DOP is not yet implemented, e.g., `DYNAMIC-ENDMARKER-FIELD`, one should use the non-strict mode for now. This is because quite a bit of code assumes that `ParameterWithDop` objects provide a DOP...)

odxtools/nameditemlist.py
```diff
@@ -152,6 +165,9 @@ def __getattr__(self, key: str) -> T:
 
         return self._item_dict[key]
 
+    def __contains__(self, key) -> bool:
+        return key in self._item_dict
```
andlaus, line 169: this method should unnecessary: `ItemAttributeList` is a `list`, so the `x` in `x in my_named_item_list` should be an item. maybe this function can be renamed to `.contains_name()`?

odxtools/diaglayer.py
```diff
@@ -591,6 +594,27 @@ def not_inherited_fn(parent_ref: ParentRef) -> List[str]:
 
         return self._compute_available_objects(get_local_objects_fn, not_inherited_fn)
 
+    def _add_import_ref(self, ecu_shared_datas) ->  None:
```
andlaus, line 597: I guess that after #285, the IMPORT-REF mechanism is implemented. What can/should be added to `Database` and `DiagLayerContainer` though, are wrapper attributes for the different kinds of `DiagLayers` (analogos to `.ecus` or `.protocols`)

odxtools/database.py
```diff
@@ -125,13 +128,20 @@ def refresh(self) -> None:
...
-            dlc._finalize_init(self._odxlinks)
+            dlc._finalize_init(self._odxlinks, self._ecu_shared_datas)
 
     @property
     def odxlinks(self) -> OdxLinkDatabase:
         """A map from odx_id to object"""
         return self._odxlinks
 
+    @property
+    def ecu_shared_datas(self) -> OdxLinkDatabase:
```
andlaus, line 139: that should return `NamedItemList[DiagLayer]` like the `.ecus` and `.protocols` properties. (on line 95 you define `self._ecu_shared_data` this way...)

odxtools/database.py
```diff
@@ -125,13 +128,20 @@ def refresh(self) -> None:
         # let the diaglayers sort out the inherited objects and the
         # short name references
         for dlc in self.diag_layer_containers:
-            dlc._finalize_init(self._odxlinks)
+            dlc._finalize_init(self._odxlinks, self._ecu_shared_datas)
```
andlaus, line 131: if you need to add such an argument, make it a `Database` object, i.e., this line should become `dlc._finalize_init(self._odxlinks, self)`. I somehow doubt that this is necessary, though...

odxtools/diaglayercontainer.py
```diff
@@ -124,17 +124,18 @@ def _resolve_odxlinks(self, odxlinks: OdxLinkDatabase) -> None:
         for ecu_variant in self.ecu_variants:
             ecu_variant._resolve_odxlinks(odxlinks)
 
-    def _finalize_init(self, odxlinks: OdxLinkDatabase) -> None:
+    def _finalize_init(self, odxlinks: OdxLinkDatabase,
+            ecu_shared_datas: NamedItemList[DiagLayer]) -> None:
         for ecu_shared_data in self.ecu_shared_datas:
-            ecu_shared_data._finalize_init(odxlinks)
+            ecu_shared_data._finalize_init(odxlinks, ecu_shared_datas)
```
andlaus, line 130: as mentioned above, it is better pass the whole `Database` object here (if this is *really* necessary.)

andlaus (changes requested):

andlaus:
> File A references file B through the IMPORT-REF method, and then using the short name from file B in file A will cause problems.

at least as far as described by the specification, this should work: only DIAG-COMM-PROXY, DTC-PROXY, DIAG-VARIABLE-PROXY, and ROW-WRAPPER elements can be imported via this mechanism. DIAG-VARIABLEs are not yet implemented and the ODXLINK references of the remaining proxy objects should be resolved before short names are dealt with. (modulo bugs.) The exception message which you posted in #285 also points into a different direction, cf. #288 ...

odxtools/nameditemlist.py
QWander: `if x in my_named_item_list:   
    print("x in my_named_item_list")    
else:    
    print("x not in my_named_item_list")    `
This method returns a Boolean value (True or False), indicating whether the key exists in the _item_dict dictionary, not the item.

QWander: all code：`return True if key in self._item_dict else False`

andlaus: sure. the point is that this class primarily is a list, not a dict, and lists check for the item values, not the keys like dictionaries...

andlaus:
@QWander: can you please check if your issue still exists with the latest version of odxtools?

andlaus:
okay, I'll close this now: I'm pretty sure this issue goes away if all files belonging to your database are loaded (e.g., using something like
```python
db = odxtools.load_files("protocols.odx-d", "base_variants.odx-d", "ecu_variants.odx-d", "comparams.odx-cs")
```
) and even if not, the [non-strict mode](https://github.com/mercedes-benz/odxtools?tab=readme-ov-file#using-the-non-strict-mode) will probably fix it.

PR #284, opened 2024-04-04
Improve the free parameter info messages

andlaus:
With this, the `parameter_info()` function handles more parameter types, deals with more DOP types and handles complex DOPs (structures) properly. Also, I tried to make the results more consistent, but IMO this is not too important, because the result of this function is intended to be informal and for human consumption.

Besides this I found a fixed a small mistake in the handling of case limits in the multiplexer code.

### DIFF
--- a/odxtools/parameterinfo.py
+++ b/odxtools/parameterinfo.py
@@ -1,105 +1,187 @@
 # SPDX-License-Identifier: MIT
-import re
-from typing import Iterable, Union
+import textwrap
+from io import StringIO
+from typing import Iterable
 
 from .compumethods.identicalcompumethod import IdenticalCompuMethod
 from .compumethods.limit import IntervalType
 from .compumethods.linearcompumethod import LinearCompuMethod
 from .compumethods.texttablecompumethod import TexttableCompuMethod
 from .dataobjectproperty import DataObjectProperty
+from .dtcdop import DtcDop
+from .dynamiclengthfield import DynamicLengthField
 from .endofpdufield import EndOfPduField
+from .exceptions import odxrequire
+from .multiplexer import Multiplexer
 from .odxtypes import DataType
 from .parameters.codedconstparameter import CodedConstParameter
 from .parameters.matchingrequestparameter import MatchingRequestParameter
 from .parameters.parameter import Parameter
 from .parameters.parameterwithdop import ParameterWithDOP
 from .parameters.reservedparameter import ReservedParameter
+from .parameters.tablekeyparameter import TableKeyParameter
+from .parameters.tablestructparameter import TableStructParameter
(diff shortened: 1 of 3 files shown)


odxtools/parameterinfo.py
```diff
@@ -1,105 +1,187 @@
...
+                of.write(f": bytes")
+            elif bdt.name.startswith("A_FLOAT"):
+                of.write(f": float")
+            elif bdt.name.startswith("A_UINT"):
+                of.write(f": uint")
+            elif bdt.name.startswith("A_INT"):
+                of.write(f": int")
+            else:
+                of.write(f": <unknown type>")
+
```
kayoub5, line 164: I think it would be better to move this logic to the `__str__` of the physical datatype class

odxtools/parameterinfo.py
```diff
@@ -1,105 +1,187 @@
...
-                ll_delim = '[' if ll.interval_type == IntervalType.CLOSED else '('
+                ll_delim = '(' if ll.interval_type == IntervalType.OPEN else '['
                 ll_str = f"{ll_delim}{ll._value!r}"
 
-            if ul is None:
+            if ul is None or ul.interval_type == IntervalType.INFINITE:
                 ul_str = "inf)"
             else:
-                ul_delim = ']' if ul.interval_type == IntervalType.CLOSED else ')'
+                ul_delim = ')' if ul.interval_type == IntervalType.OPEN else ']'
```
kayoub5, line 176: this logic is better off being in the __str__ of the class

tests/test_somersault.py
```diff
@@ -305,20 +305,20 @@ def test_free_param_info(self) -> None:
         stdout = StringIO()
         with patch("sys.stdout", stdout):
             request.print_free_parameters_info()
-            expected_output = "forward_soberness_check: uint8\nnum_flips: uint8\n"
+            expected_output = "forward_soberness_check: uint\nnum_flips: uint\n"
```
kayoub5, line 308: no size info?

andlaus: adding a size lead to quite unexpected results like `float16`. I think the sizes should be removed here anyway, as this function is concerned with the physical types of the parameters and in the physical space, the bit sizes of the internal representation of an object should not matter (only the valid range does).

odxtools/parameterinfo.py
kayoub5: Not in the limit class, but in its container

andlaus: you mean in `CompuMethod`? I'd rather leave its `__str__()` method auto-generated by dataclass. This way it is guaranteed to be complete...

kayoub5 (approved):

PR #285, opened 2024-04-05
Implement the IMPORT-REF mechanism

andlaus:
If I understood the spec correctly, this simply extends the pool of objects that are locally referenceable via ODXLINK (i.e., without having to explicitly specify the DOCREF attribute in the reference), and can thus be seen as a kind of "poor man's inheritance" mechanism.

### DIFF
--- a/odxtools/diaglayer.py
+++ b/odxtools/diaglayer.py
@@ -1,6 +1,7 @@
 # SPDX-License-Identifier: MIT
 import re
 import warnings
+from copy import copy
 from dataclasses import dataclass
 from functools import cached_property
 from itertools import chain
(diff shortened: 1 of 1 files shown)


andlaus:
@QWander: Since I do not have any files which use IMPORT-REF, it would be nice if you could check if this fixes the issues which you currently have with the `main` version (cf. #283)

odxtools/diaglayer.py
```diff
@@ -76,6 +77,46 @@ def _build_odxlinks(self) -> Dict[OdxLinkId, Any]:
...
+
+                imported_dl_links = imported_dl._build_odxlinks()
+                for link_id, obj in imported_dl_links.items():
+                    # the imported objects shall behave as if they
+                    # where defined by the importing layer. IOW, they
+                    # must be visible in the same document fragments.
+                    link_id = OdxLinkId(link_id.local_id, self.odx_id.doc_fragments)
+                    imported_links[link_id] = obj
+
+            odxlinks = copy(odxlinks)
```
kayoub5, line 114: Why copy?

andlaus: because this function is not supposed to modify its argument: after the function the object passed to it must be in its original state since the imported references only apply within this specific diag layer. (in C++, the `odxlinks` argument would be declared `const`.)

kayoub5: Put that as comment, otherwise someone will remove the line by accident

andlaus: good idea, done...

kayoub5 (approved):

andlaus:
okay. Even though it seems like the issues @QWander experiences are not caused by the IMPORT-REFS mechnism, there is a decent chance that this patch works as advertised. Let's merge and fix the fallout (if any) later...

PR #291, opened 2024-04-17
Refactor encoding, part 1 (the overture)

andlaus:
This is the first PR in a series concerned with refactoring the encoding code. The goal of the PR series is to "de-spaghettify" this part of the code making it much easier to recognize it as the inverse of the decoding code.

This PR first fixes two unrelated issues (improvements of `NrcConstParameter`  and renaming keyword argument dictionaries to `params` to `kwargs`) and follows by refactoring the `EncodeState` class analogous to `DecodeState` (the next PRs in the series will make use of the new attributes).

### DIFF
--- a/odxtools/basicstructure.py
+++ b/odxtools/basicstructure.py
@@ -245,18 +245,18 @@ class BasicStructure(ComplexDop):
 
         return result
 
-    def encode(self, coded_request: Optional[bytes] = None, **params: ParameterValue) -> bytes:
+    def encode(self, coded_request: Optional[bytes] = None, **kwargs: ParameterValue) -> bytes:
         """
         Composes an UDS message as bytes for this service.
         Parameters:
         ----------
         coded_request: bytes
             coded request (only needed when encoding a response)
-        params: dict
+        kwargs: dict
             Parameters of the RPC as mapping from SHORT-NAME of the parameter to the value
         """
         return self.convert_physical_to_internal(
-            params, triggering_coded_request=coded_request, is_end_of_pdu=True)
+            kwargs, triggering_coded_request=coded_request, is_end_of_pdu=True)
 
     def decode(self, message: bytes) -> ParameterValueDict:
         decode_state = DecodeState(coded_message=message)
--- a/odxtools/basicstructure.py
+++ b/odxtools/basicstructure.py
@@ -76,12 +76,7 @@ class BasicStructure(ComplexDop):
     def coded_const_prefix(self, request_prefix: bytes = b'') -> bytes:
         prefix = b''
         encode_state = EncodeState(
-            bytearray(prefix),
-            parameter_values={},
-            triggering_request=request_prefix,
-            cursor_byte_position=0,
-            cursor_bit_position=0,
-            origin_byte_position=0)
+            bytearray(prefix), parameter_values={}, triggering_request=request_prefix)
         for param in self.parameters:
             if isinstance(param, (CodedConstParameter, NrcConstParameter, MatchingRequestParameter,
                                   PhysicalConstantParameter)):
@@ -143,10 +138,7 @@ class BasicStructure(ComplexDop):
             bytearray(),
             dict(param_value),
             triggering_request=triggering_coded_request,
-            is_end_of_pdu=False,
-            cursor_byte_position=0,
-            cursor_bit_position=0,
-            origin_byte_position=0)
+            is_end_of_pdu=False)
 
         for param in self.parameters:
             if param == self.parameters[-1]:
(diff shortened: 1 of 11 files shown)


odxtools/basicstructure.py
```diff
@@ -76,7 +76,12 @@ def get_static_bit_length(self) -> Optional[int]:
     def coded_const_prefix(self, request_prefix: bytes = b'') -> bytes:
         prefix = b''
         encode_state = EncodeState(
-            bytearray(prefix), parameter_values={}, triggering_request=request_prefix)
+            bytearray(prefix),
+            parameter_values={},
+            triggering_request=request_prefix,
+            cursor_byte_position=0,
+            cursor_bit_position=0,
+            origin_byte_position=0)
```
kayoub5, line 84: In what case where `cursor_byte_position` and others would not be zero during EncodeState creation?

andlaus: this happens if a partially completed PDU is supposed to be modified. But I agree, this does not happen in the current code base, so these attributes ought to default to 0. fixed: c564c4a80df1acec67a845e2d3596ae7057ae5a0

odxtools/encodestate.py
```diff
@@ -28,31 +37,36 @@ class EncodeState:
...
-                warnings.warn(
-                    f"Object '{param_name}' overlaps with another parameter (bytes are already set)",
-                    OdxWarning,
-                    stacklevel=1,
+        for i in range(len(new_data)):
+            # insert new byte. this is pretty hacky: it will fail if
+            # the value to be inserted is bitwise "disjoint" from the
+            # value which is already in the PDU...
+            if self.coded_message[pos + i] & new_data[i] != 0:
+                odxraise(
```
kayoub5, line 65: Don't change from warning to raise, or you will be break the PDX files I have.

odxtools/parameters/nrcconstparameter.py
```diff
@@ -88,7 +88,7 @@ def is_required(self) -> bool:
     @property
     @override
     def is_settable(self) -> bool:
-        return False
+        return True
```
kayoub5, line 91: Sure about this?

andlaus: hm, okay. I've read up the spec about NRC-CONST and I now think that it is supposed to overlapp with a value parameter with the aim of distinguishing negative responses based on the value of the value parameter (i.e., some strange kind of multiplexer?!). when it comes to encoding, it should thus probably not do anything except moving the cursor position. this would require some additional refactoring, mainly in the decoding logic, but I guess I have to think about it a bit more...

andlaus:
@kayoub5: do you agree that this can be merged? I think that the present PR should not regress anything and while the situation with the `NrcConstParameter` is certainly not ideal, fixing it requires more extensive changes that are out of scope for this PR series...

odxtools/parameters/nrcconstparameter.py
kayoub5: This is fine from my point of view, just wanted to make sure, since it seemed off topic compared to the rest of the PR

tests/test_encoding.py
```diff
@@ -287,7 +287,10 @@ def test_encode_overlapping(self) -> None:
             parameters=NamedItemList([param1, param2, param3]),
             byte_size=None,
         )
-        self.assertEqual(req.encode(), bytearray([0x12, 0x34, 0x56]))
+
+        with self.assertRaises(EncodeError):
```
kayoub5, line 291: This should not throw, only warn

andlaus: it only raises an exception when in strict mode...

odxtools/encodestate.py
andlaus: can you use the non-strict mode? if that is activated, it only logs a warning under the "odxtools" topic...

andlaus: (I think overlapping usually parameters are a bug in the PDX file (maybe except for NRC-CONST), and should thus be clearly flagged because if this is encountered the results are undefined, i.e., they might be surprising.)

odxtools/parameters/nrcconstparameter.py
andlaus: Good! I'm constantly surprised about the arcaneness of some of the mechanisms specified by the ODX format. In this case, why are NRC-CONST parameters limited to negative responses (and why are they around in the first place? It seems to me that they can always be replaced by a TABLE-STRUCT parameter)?

odxtools/encodestate.py
andlaus: no, but a while ago, somebody [added a piece of code](https://github.com/mercedes-benz/odxtools/commit/b844c7f396cffd079e9b4fd6698eaeabe5275f78) that spits out a warning if they do. I think that it is sensible to turn this up a notch when we're paranoid about undefined behavior (that's how I interpret the strict mode). But I see your point: [reverted](https://github.com/mercedes-benz/odxtools/pull/291/commits/35fd6ce8bb57be0f9af8bbcaef5d8c8381c924dd)...

kayoub5 (approved):

PR #292, opened 2024-04-18
Refactor encoding, part 2 (diag coded types)

andlaus:
This converts the diag coded types to the new and shiny encoding infrastructure. Note that doing this as a stand-alone change requires some kludges like creating temporary `EncodeState` objects. These will be removed by later PRs in the series...

### DIFF
--- a/odxtools/dtcdop.py
+++ b/odxtools/dtcdop.py
@@ -147,9 +147,21 @@ class DtcDop(DopBase):
                               f" DiagnosticTroubleCode but got {physical_value!r}.")
 
         internal_trouble_code = self.compu_method.convert_physical_to_internal(trouble_code)
+        tmp_state = EncodeState(
+            bytearray(),
+            encode_state.parameter_values,
+            triggering_request=encode_state.triggering_request,
+            is_end_of_pdu=False,
+            cursor_byte_position=0,
+            cursor_bit_position=0,
+            origin_byte_position=0)
+        encode_state.cursor_bit_position = encode_state.cursor_bit_position
+        self.diag_coded_type.encode_into_pdu(internal_trouble_code, encode_state=tmp_state)
 
-        return self.diag_coded_type.convert_internal_to_bytes(
-            internal_trouble_code, encode_state=encode_state, bit_position=bit_position)
+        encode_state.length_keys.update(tmp_state.length_keys)
+        encode_state.table_keys.update(tmp_state.table_keys)
+
+        return tmp_state.coded_message
 
     def _build_odxlinks(self) -> Dict[OdxLinkId, Any]:
         odxlinks = super()._build_odxlinks()
--- a/tests/test_diag_coded_types.py
+++ b/tests/test_diag_coded_types.py
@@ -416,6 +416,7 @@ class TestParamLengthInfoType(unittest.TestCase):
             coded_message=bytearray([0xcc]), parameter_values={}, cursor_byte_position=2)
         dct.encode_into_pdu(0x12345, state)
         self.assertEqual(state.coded_message.hex(), "cc00012345")
+        self.assertEqual(state.length_keys.get("length_key"), 24)
 
     def test_end_to_end(self) -> None:
         # diag coded types
--- a/tests/test_diag_coded_types.py
+++ b/tests/test_diag_coded_types.py
@@ -14,7 +14,7 @@ from odxtools.diaglayer import DiagLayer
 from odxtools.diaglayerraw import DiagLayerRaw
 from odxtools.diaglayertype import DiagLayerType
 from odxtools.encodestate import EncodeState
-from odxtools.exceptions import DecodeError, EncodeError, odxrequire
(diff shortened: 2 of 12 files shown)


odxtools/dtcdop.py
```diff
@@ -147,9 +147,21 @@ def convert_physical_to_bytes(self,
...
+        tmp_state = EncodeState(
+            bytearray(),
+            encode_state.parameter_values,
+            triggering_request=encode_state.triggering_request,
+            is_end_of_pdu=False,
+            cursor_byte_position=0,
+            cursor_bit_position=0,
+            origin_byte_position=0)
+        encode_state.cursor_bit_position = encode_state.cursor_bit_position
+        self.diag_coded_type.encode_into_pdu(internal_trouble_code, encode_state=tmp_state)
```
kayoub5, line 159: make a copy/clone function in the encode state, instead of all this manual attributes copying

tests/test_encoding.py
```diff
@@ -287,7 +287,7 @@ def test_encode_overlapping(self) -> None:
             parameters=NamedItemList([param1, param2, param3]),
             byte_size=None,
         )
-        self.assertEqual(req.encode(), bytearray([0x12, 0x34, 0x56]))
+        self.assertEqual(req.encode(), bytearray([0x33, 0x75, 0x56]))
```
kayoub5, line 290: why?

tests/test_diag_coded_types.py
```diff
@@ -694,9 +703,10 @@ def test_encode_min_max_length_type_zero(self) -> None:
             termination="ZERO",
             is_highlow_byte_order_raw=None,
         )
-        state = EncodeState(bytearray([0x12]), parameter_values={}, is_end_of_pdu=False)
-        byte_val = dct.convert_internal_to_bytes("Hi", state, bit_position=0)
-        self.assertEqual(byte_val, bytes([0x48, 0x69, 0x0]))
+        state = EncodeState(
+            coded_message=bytearray([0x00]), parameter_values={}, is_end_of_pdu=False)
```
kayoub5, line 707: why was the 0x12 replaced with zero?

tests/test_diag_coded_types.py
```diff
@@ -407,9 +412,10 @@ def test_encode_param_info_length_type_uint(self) -> None:
         odxlinks = OdxLinkDatabase()
         odxlinks.update({length_key_id: length_key})
         dct._resolve_odxlinks(odxlinks)
-        state = EncodeState(bytearray([0x10]), {length_key.short_name: 40})
-        byte_val = dct.convert_internal_to_bytes(0x12345, state, bit_position=0)
-        self.assertEqual(byte_val.hex(), "0000012345")
+        state = EncodeState(
+            coded_message=bytearray([0xcc]), parameter_values={}, cursor_byte_position=2)
+        dct.encode_into_pdu(0x12345, state)
+        self.assertEqual(state.coded_message.hex(), "cc00012345")
```
kayoub5, line 418: why did the assertion changed here?

tests/test_diag_coded_types.py
```diff
@@ -70,62 +70,61 @@ def test_decode_leading_length_info_type_zero_length(self) -> None:
...
+        # little endian
         dct = LeadingLengthInfoType(
             base_data_type=DataType.A_UNICODE2STRING,
             bit_length=8,
             base_type_encoding=None,
             is_highlow_byte_order_raw=False,
         )
-        state = DecodeState(bytes([0x12, 0x4, 0x61, 0x00, 0x39, 0x00]), cursor_byte_position=1)
+        state = DecodeState(
+            bytes([0x12, 0x4, 0x61, 0x00, 0x39, 0x00, 0xff, 0x00]), cursor_byte_position=1)
```
kayoub5, line 123: different bytes?

tests/test_encoding.py
andlaus: overlapping parameters result in undefined behaviour and I was recently persuated that it is better not to raise an exception in this case, not even in strict mode... ;)

tests/test_diag_coded_types.py
andlaus: that was done to avoid those pesky "overlapping parameter" warnings: `convert_internal_to_bytes()` simply did not care what already was in the PDU, `encode_into_pdu()` does...

tests/test_diag_coded_types.py
andlaus: it's been a while since I did this, so take this with a grain of salt: I suppose it is because this sets the cursor byte position to 2, and the initial value of the PDU to 0xcc. I don't remember why the old code produced two leading zero bytes...

tests/test_diag_coded_types.py
andlaus: this tests for little-endian UTF-16 encoding, so the bytes representing the two letters are byte swapped compared to what's on line 106. (the first and the trailing two bytes are just garbage to check if the code properly works with arbitrary positions.)

kayoub5 (approved):

PR #293, opened 2024-04-22
Refactor encoding, part 3 (parameters)

andlaus:
The quest continues. This time the parameters are converted to make use of the new and shiny encoding machinery...

### DIFF
--- a/odxtools/basicstructure.py
+++ b/odxtools/basicstructure.py
@@ -74,13 +74,13 @@ class BasicStructure(ComplexDop):
         return ((length + 7) // 8) * 8
 
     def coded_const_prefix(self, request_prefix: bytes = b'') -> bytes:
-        prefix = b''
         encode_state = EncodeState(
-            bytearray(prefix), parameter_values={}, triggering_request=request_prefix)
+            coded_message=bytearray(), parameter_values={}, triggering_request=request_prefix)
+
         for param in self.parameters:
-            if isinstance(param, (CodedConstParameter, NrcConstParameter, MatchingRequestParameter,
-                                  PhysicalConstantParameter)):
-                encode_state.coded_message = bytearray(param.encode_into_pdu(encode_state))
+            if (isinstance(param, MatchingRequestParameter) and param.request_byte_position < len(request_prefix)) or \
+                isinstance(param, (CodedConstParameter, NrcConstParameter, PhysicalConstantParameter)):
+                param.encode_into_pdu(physical_value=None, encode_state=encode_state)
             else:
                 break
         return encode_state.coded_message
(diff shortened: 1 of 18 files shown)


odxtools/basicstructure.py
```diff
@@ -122,52 +122,69 @@ def convert_physical_to_internal(self,
...
-                # these keys), to avoid getting an "overlapping
+                # these keys). To avoid getting an "overlapping
                 # parameter" warning, we must encode a value of zero
                 # into the PDU here and add the real value of the
-                # parameter in a post processing step.
-                tmp = encode_state.parameter_values.pop(param.short_name)
-                encode_state.coded_message = bytearray(param.encode_into_pdu(encode_state))
-                encode_state.parameter_values[param.short_name] = tmp
+                # parameter in a post-processing step.
+                param.encode_placeholder_into_pdu(
```
andlaus, line 166: the mechanism for length and table key parameters also changed: in the first loop, key parameters only encode all-zero placeholders into the PDU, and in the post processing step below, the real data is added via `.encode_value_into_pdu()`...

odxtools/basicstructure.py
```diff
@@ -122,52 +122,69 @@ def convert_physical_to_internal(self,
...
+                    odxraise(f"Value for unknown parameter '{param_value_name}' specified "
+                             f"for structure {self.short_name}")
 
         for param in self.parameters:
-            if param == self.parameters[-1]:
-                # The last parameter is at the end of the PDU if the
-                # structure itself is at the end of the PDU. TODO:
-                # This assumes that the last parameter specified in
-                # the ODX is located last in the PDU...
+            if id(param) == id(self.parameters[-1]):
```
kayoub5, line 151: Why?

odxtools/basicstructure.py
```diff
@@ -122,52 +122,69 @@ def convert_physical_to_internal(self,
...
-            encode_state.coded_message = encode_state.coded_message.ljust(self.byte_size, b"\0")
+            param.encode_into_pdu(
+                physical_value=param_value.get(param.short_name), encode_state=encode_state)
+
+        if self.byte_size is not None:
+            if len(encode_state.coded_message) < self.byte_size:
+                # Padding bytes needed
+                encode_state.coded_message = encode_state.coded_message.ljust(self.byte_size, b"\0")
+            elif len(encode_state.coded_message) > self.byte_size:
+                odxraise(
```
kayoub5, line 183: Isn't this already in `_validate_coded_message_size` ?

odxtools/parameters/tablekeyparameter.py
```diff
@@ -133,46 +133,98 @@ def is_settable(self) -> bool:
...
+    @final
+    def _encode_positioned_into_pdu(self, physical_value: Optional[ParameterValue],
+                                    encode_state: EncodeState) -> None:
+        # if you get this exception, you ought to use
+        # `.encode_placeholder_into_pdu()` followed by (after the
+        # value of the table key has been determined)
+        # `.encode_value_into_pdu()`.
+        raise RuntimeError("_encode_positioned_into_pdu() cannot be called for table keys.")
+
+    def encode_placeholder_into_pdu(self, physical_value: Optional[ParameterValue],
```
kayoub5, line 145: Not possible to merge _encode_positioned_into_pdu and encode_placeholder_into_pdu into one?

odxtools/basicstructure.py
andlaus: mainly  because comparing complex objects is potentially very expensive while comparing object IDs is cheap. (and at this particular place we get away with ids.)

odxtools/basicstructure.py
andlaus: good point. fixed.

odxtools/parameters/tablekeyparameter.py
andlaus: okay, I've just implemented a prototype doing it this way. It saves 10-20 lines of code, but I still don't like it because the semantics of `.encode_into_pdu()` would be completely different for non-key parameters compared to "normal" ones while exhibiting the same API (which would make bugs due to incorrect API usage very easy to do and hard to find). On the other hand, the proposed approach makes it quite obvious if `.encode_into_pdu()` is used incorrectly...

kayoub5 (approved):

PR #295, opened 2024-04-23
Refactor encoding, part 5 (the finale)

andlaus:
this PR contains a few cleanups/improvements for the encoding machinery which have been made possible by the rest of the series.

### DIFF
--- a/odxtools/basicstructure.py
+++ b/odxtools/basicstructure.py
@@ -208,7 +208,7 @@ class BasicStructure(ComplexDop):
                 # position directly after the structure and let
                 # EncodeState add the padding as needed.
                 encode_state.cursor_byte_position = encode_state.origin_byte_position + self.byte_size
-                encode_state.emplace_atomic_value(b'', "<PADDING>")
+                encode_state.emplace_bytes(b'', "<PADDING>")
 
         # encode the length- and table keys. This cannot be done above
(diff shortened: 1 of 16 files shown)


odxtools/basicstructure.py
```diff
@@ -247,26 +246,6 @@ def decode_from_pdu(self, decode_state: DecodeState) -> ParameterValue:
 
         return result
 
-    def encode(self, coded_request: Optional[bytes] = None, **kwargs: ParameterValue) -> bytes:
```
kayoub5, line 250: I need the encode function on the structure itself:
Use case: I split services into static parts an dynamic parts (that are mostly structures), and use the dynamic part to generate `udsoncan` configuration as per https://udsoncan.readthedocs.io/en/latest/udsoncan/examples.html#reading-a-did-with-readdatabyidentifier

andlaus: can't you use `.encode_into_pdu()`? (creating an `EncodeState` object and calling `self.encode_into_pdu()` all `.encode()` does.) The reason I would like to move the toplevel function to the Request/Response classes is that they require different parameters, i.e., `Response.encode()` needs `triggering_request` which does not make sense for requests...

andlaus:
@kayoub5: do you see any remaining issues with this?

kayoub5 (approved):

PR #299, opened 2024-05-06
Improve short name reference handling

andlaus:
This PR contains two SNREF related changes:

- We make it possible for any parameter to reference other parameters via short name reference. This can be regarded as a cleanup and should come with with a very limited practical impact. 
- Short name resolutions are no longer done by (mis)using `NamedItemList.get()`. This makes it possible to reference objects which do not exhibit short names that are valid python identifiers. (E.g. because they start with a digit.)

### DIFF
--- a/odxtools/utils.py
+++ b/odxtools/utils.py
@@ -1,11 +1,8 @@
 # SPDX-License-Identifier: MIT
 import dataclasses
 import re
-from typing import Any, Dict, Iterable, Optional, Type, overload
+from typing import Any, Dict, Optional
 from xml.etree import ElementTree
-
-from .exceptions import odxraise
-from .nameditemlist import OdxNamed, TNamed
 
 
 def create_description_from_et(et_element: Optional[ElementTree.Element],) -> Optional[str]:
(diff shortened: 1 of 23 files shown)


odxtools/utils.py
```diff
@@ -60,3 +65,33 @@ def is_short_name_path(test_val: str) -> bool:
     See also: ISO 22901 section 7.3.13.3
     """
     return _short_name_path_pattern.fullmatch(test_val) is not None
+
+
+@overload
+def resolve_snref(target_short_name: str,
```
kayoub5, line 71: I think this method should be in `odxlink` file

odxtools/utils.py
```diff
@@ -60,3 +65,33 @@ def is_short_name_path(test_val: str) -> bool:
...
+    """Properly resolve a short name reference"""
+    ...
+
+
+@overload
+def resolve_snref(target_short_name: str, items: Iterable[OdxNamed], expected_type: Type[T]) -> T:
+    ...
+
+
+def resolve_snref(target_short_name: str,
```
kayoub5, line 83: No `@overload` here?

odxtools/utils.py
```diff
@@ -1,9 +1,14 @@
...
 import dataclasses
 import re
-from typing import Any, Dict, Optional
+from typing import Any, Dict, Iterable, Optional, Type, TypeVar, overload
 from xml.etree import ElementTree
 
+from .exceptions import odxraise
+from .nameditemlist import OdxNamed
+
+T = TypeVar("T")
```
kayoub5, line 10:
```suggestion
T = TypeVar("T",  OdxNamed)
```

odxtools/utils.py
andlaus: this is the implementation. As far as I know, these cannot be marked `@overload` because their signature must suitable for any of the overloads...

odxtools/utils.py
andlaus: changed, thanks!

odxtools/utils.py
andlaus: the thought was that `odxlink.py` contains the infrastructure required for the `ODXLINK` mechanism as specified by the standard (i.e., `FOO-REF ID="bar"` tags) while short name references are a entirely different mechanism (one with awful semantics, cf. #200). If you believe it is better to move the function there, I'll do so...

kayoub5: odxlink seems more suitable/related than a generic utils file

andlaus: okay, done: https://github.com/mercedes-benz/odxtools/pull/299/commits/ac56647d3aaa7bb014269ad7ef1c285055e35495

kayoub5 (approved):

PR #304, opened 2024-05-13
diag coded types: change instantiation to the `.from_et()` pattern

andlaus:
This switches creation of the diag coded types to the standard `.from_et()` pattern.

### DIFF
--- a/odxtools/createanydiagcodedtype.py
+++ b/odxtools/createanydiagcodedtype.py
@@ -1,85 +1,28 @@
 # SPDX-License-Identifier: MIT
-from typing import List, cast
+from typing import List
 from xml.etree import ElementTree
 
 from .diagcodedtype import DiagCodedType
-from .exceptions import odxraise, odxrequire
+from .exceptions import odxraise
 from .globals import xsi
 from .leadinglengthinfotype import LeadingLengthInfoType
 from .minmaxlengthtype import MinMaxLengthType
-from .odxlink import OdxDocFragment, OdxLinkRef
-from .odxtypes import DataType, odxstr_to_bool
+from .odxlink import OdxDocFragment
 from .paramlengthinfotype import ParamLengthInfoType
 from .standardlengthtype import StandardLengthType
 
 
 def create_any_diag_coded_type_from_et(et_element: ElementTree.Element,
                                        doc_frags: List[OdxDocFragment]) -> DiagCodedType:
-    base_type_encoding = et_element.get("BASE-TYPE-ENCODING")
-
-    base_data_type_str = odxrequire(et_element.get("BASE-DATA-TYPE"))
-    try:
-        base_data_type = DataType(base_data_type_str)
-    except ValueError:
-        base_data_type = cast(DataType, None)
-        odxraise(f"Unknown base data type {base_data_type_str}")
-
-    is_highlow_byte_order_raw = odxstr_to_bool(et_element.get("IS-HIGHLOW-BYTE-ORDER"))
-
     dct_type = et_element.get(f"{xsi}type")
-    bit_length = None
     if dct_type == "LEADING-LENGTH-INFO-TYPE":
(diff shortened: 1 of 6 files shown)


odxtools/diagcodedtype.py
```diff
@@ -43,9 +61,10 @@ def get_static_bit_length(self) -> Optional[int]:
         return None
 
     @property
-    @abc.abstractmethod
     def dct_type(self) -> DctType:
-        pass
+        odxraise(f"Class {type(self).__name__} does not override required method "
+                 f"dct_type()", NotImplementedError)
+        return "STANDARD-LENGTH-TYPE"
```
kayoub5, line 67: why is this kind of weird abc implementation?

andlaus: the problem with normal ABCs is that they cannot be instantiated, yet we need to do so for the `from_et()` pattern to work with class hierarchies. the `return "STANDARD-LENGTH-TYPE"` statement is added to have a fighting chance of the calling code not crashing in non-strict mode.

kayoub5: what error are you getting?

andlaus: I have not seen any, but this code path is taken in non-strict mode if an unimplemented DCT is encountered, and the normal procedure for non-strict mode is to try to continue gracefully.

kayoub5: those kind of problems are solved by making a NotSupprtedDtc class, instead of creating an instance of an abstract class

andlaus: how do you do inheritance then? Instantiating `NotSupportedDtc` in `.from_et()` whilst deriving from a different class feels pretty unnatural for me for this...

(note that in C++ the constructors could simply be overloaded, but unfortunately this is impossible with python.)

odxtools/createanydiagcodedtype.py
```diff
@@ -1,85 +1,28 @@
...
-            bit_mask=bit_mask,
-            is_condensed_raw=is_condensed_raw,
-            base_type_encoding=base_type_encoding,
-            is_highlow_byte_order_raw=is_highlow_byte_order_raw,
-        )
-    raise NotImplementedError(f"I do not know the diag-coded-type {dct_type}")
+        return StandardLengthType.from_et(et_element, doc_frags)
+
+    odxraise(f"Unknown DIAG-CODED-TYPE {dct_type}", NotImplementedError)
+    return DiagCodedType.from_et(et_element, doc_frags)
```
kayoub5, line 28: `NotSupportedDtc.from_et` will only be used here instead of the abstract class

andlaus: yes, but I think it is weird to derive from one class and call the constructor of another one which is seemingly unrelated and has a very confusing name. Also, this would add a basically empty class to the code base. tl;dr: Although far from perfect, I vastly prefer the "manual ABC" option...

odxtools/diagcodedtype.py
kayoub5: what dtc type we still don't support?

andlaus: as far a I can see we currently have all mentioned by the ODX 2.2 XSD, i.e., this code path only affects incorrect files or files using a future version of the standard (assuming that this version introduces new or renames existing DCTs).

kayoub5: So you can just return `None` ?

andlaus: yeah, good point. that can be done. fixed: https://github.com/mercedes-benz/odxtools/pull/304/commits/2ddb472981b0278c3d864622e44650e9101359ba

andlaus:
@kayoub5: Do you see any additional issues with this?

kayoub5 (approved):

PR #305, opened 2024-05-16
More list comprehensions

andlaus:
This replaces `create_sdgs_from_et()` and `create_company_datas_from_et()` by list comprehensions. The reasons are that these functions were pretty trivial, doing it via list comprehensions is more explicit and it is even slightly shorter overall.


odxtools/comparamsubset.py
```diff
@@ -45,7 +43,10 @@ def from_et(et_element: ElementTree.Element,
         kwargs = dataclass_fields_asdict(IdentifiableElement.from_et(et_element, doc_frags))
 
         admin_data = AdminData.from_et(et_element.find("ADMIN-DATA"), doc_frags)
-        company_datas = create_company_datas_from_et(et_element.find("COMPANY-DATAS"), doc_frags)
+        company_datas = NamedItemList([
+            CompanyData.from_et(cde, doc_frags)
+            for cde in et_element.iterfind("COMPANY-DATAS/COMPANY-DATA")
```
kayoub5, line 48: Maybe there should be a common base class for SDG / ADMIN DATA / COMPANY-DATA ?

andlaus: no: these are all store quite different data for different purposes, cf the XSD...

kayoub5: what I meant is that classes that have company data, should use a common base class, not that sdg and company data should have a common class

andlaus: we should stay with what the XSD defines, lest it become very confusing...

kayoub5 (approved):

PR #306, opened 2024-05-22
Refactor the compu methods

andlaus:
This PR is quite a whopper (sorry), but all of the changes are closely related to the compu methods and I could not think of a good way of splitting this into smaller chunks.

With this PR, compu methods work as they are defined by the XSD: All parameters of all compu methods are specified using attributes of the basic `CompuMethod` class (i.e., mainly `COMPU-INTERNAL-TO-PHYS` and `COMPU-PHYS-TO-INTERNAL`) and the specific behavior of the individual categories are implemented in subclasses. "Specific behavior" encompasses extracting the data relevant for the concrete compu method from their generic form via `__post_init__()` methods in the subclasses and, based on this, the actual computations to translate internal values to/from physical ones. This approach has the advantages that it is quite a bit cleaner than the previous code (albeit still quite messy IMO), that all compu methods can be handled using the same output writing code and -- most importantly -- that the code now reflects the specification reasonably closely now.


odxtools/compumethods/compudefaultvalue.py
```diff
@@ -0,0 +1,27 @@
...
+from dataclasses import dataclass
+from typing import List, Optional
+from xml.etree import ElementTree
+
+from ..odxlink import OdxDocFragment
+from .compuinversevalue import CompuInverseValue
+
+
+@dataclass
+class CompuDefaultValue:
```
kayoub5, line 11: CompuDefaultValue could inherit from CompuConst

odxtools/compumethods/compuinversevalue.py
```diff
@@ -0,0 +1,20 @@
+# SPDX-License-Identifier: MIT
+from dataclasses import dataclass
+from typing import List, Optional
+from xml.etree import ElementTree
+
+from ..odxlink import OdxDocFragment
+
+
+@dataclass
+class CompuInverseValue:
```
kayoub5, line 10: This looks just like a CompuConst

odxtools/compumethods/linearsegment.py
```diff
@@ -0,0 +1,176 @@
...
+
+        if self.internal_type in [
+                DataType.A_INT32,
+                DataType.A_UINT32,
+        ]:
+            result = round(result)
+
+        return result
+
+    def __compute_physical_limits__(self) -> None:
```
kayoub5, line 97: functions names that end and start with double underscore are reserved for python itself

odxtools/compumethods/scalelinearcompumethod.py
```diff
@@ -1,39 +1,95 @@
...
+            return
+
+        compu_scales = self.compu_internal_to_phys.compu_scales
+
+        for scale in compu_scales:
+            self._segments.append(
+                LinearSegment.from_compu_scale(
+                    scale, internal_type=self.internal_type, physical_type=self.physical_type))
+
+    def convert_physical_to_internal(self, physical_value: AtomicOdxType) -> Union[float, int]:
```
kayoub5, line 65: float as internal value?

odxtools/compumethods/scalelinearcompumethod.py
```diff
@@ -1,39 +1,95 @@
...
+
+    def convert_internal_to_physical(self, internal_value: AtomicOdxType) -> Union[float, int]:
+        applicable_segments = [
+            seg for seg in self._segments if seg.internal_applies(internal_value)
+        ]
+        if not applicable_segments:
+            odxraise(r"No applicable segment for value {internal_value} found")
+            return None
+        elif len(applicable_segments):
+            odxraise(r"Multiple applicable segments for value {internal_value} found")
```
kayoub5, line 86: Is there something in the specs that says this is not allowed?

odxtools/compumethods/scalelinearcompumethod.py
```diff
@@ -1,39 +1,95 @@
...
+
+        seg = applicable_segments[0]
+        return seg.convert_physical_to_internal(physical_value)
+
+    def convert_internal_to_physical(self, internal_value: AtomicOdxType) -> Union[float, int]:
+        applicable_segments = [
+            seg for seg in self._segments if seg.internal_applies(internal_value)
+        ]
+        if not applicable_segments:
+            odxraise(r"No applicable segment for value {internal_value} found")
```
kayoub5, line 83: What does the specs say about this?

odxtools/compumethods/compudefaultvalue.py
andlaus: yes, but it does not do so in the XSD. given that CompuConst is pretty trivial, I prefer to stick with the XSD...

andlaus: okay, I changes my mind: it is (slightly) beneficial to derive from `CompuConst` because of the additional logic to access the value converted to the real data type. fixed: https://github.com/mercedes-benz/odxtools/pull/306/commits/cbd5a8f2232036e27181ace65dc531f67d16b0dd

odxtools/compumethods/compuinversevalue.py
andlaus: good point. I fixed this by adding making `CompuInverseValue` alias of `CompuConst`  in the latest revision.

odxtools/compumethods/linearsegment.py
andlaus: Aren't they just private methods, i.e. methods that cannot be called outside of the class?

odxtools/compumethods/scalelinearcompumethod.py
andlaus: yeah, the signature of this method compatible with the signature of the base class because it only exhibits a narrower return type. `mypy` is okay with that because the calling code expecting the signature of the base class works with this as well...

odxtools/compumethods/scalelinearcompumethod.py
andlaus: it says "The boundaries of the intervals shall not overlap." (section 7.3.6.6.4). That said, it allows "gaps" in the ranges...

andlaus: for this method, this is not relevant, though, because we must be able to decode a given internal value...

odxtools/compumethods/scalelinearcompumethod.py
andlaus: we must be able to decode the internal value. probably I should throw a `DecodeError` instead, though. fixed: https://github.com/mercedes-benz/odxtools/pull/306/commits/10bc14b68151debc7084042f17ee86f72a775416

odxtools/compumethods/linearsegment.py
andlaus: You're right, I looked this up: PEP 8 states that private methods use to leading underscores while leading and trailing underscores are reserved for magic functions. fixed: https://github.com/mercedes-benz/odxtools/pull/306/commits/58ea6317eeb654b607c5942462ba37bfa96ba503 (I wonder why `mypy` and `ruff` do not complain about this...)

odxtools/compumethods/scalelinearcompumethod.py
kayoub5: I am asking if ODX allows internal values to be a float, I though only integers were allowed

kayoub5 (approved):

Issue #224, opened 2023-10-20
load a ".odx-d" file(diag-layer-container) failed due to ID-REF to an object from another ".odx-d"(functional group) file


chenyin0126:
I tried to load a base-variant .odx-d file to generate a database object, with `def load_file(file_name: str) -> Database:`, this file contains diag-layer-container information, 

,however, the load file failed due to 

`\"ODXLINK reference OdxLinkRef('FGL_UDS') could not be resolved for any of the document fragments [OdxDocFragment(doc_name='DLC_FGL_UDS', doc_type='CONTAINER')]\`

the reason for this is we only make one odx-d file as input to load the file, but this base variant odx-d file contains a ref to the other functional group odx-d file. So the load file process failed. 

Can we make more than one file as input to load odx-d? if we can load both base variant and functional group, we can avoid this issue; or can we make load function to ignore if we cannot link to the object from another file?

andlaus:
as a quick fix to get the file loading, you can try to use the [non-strict mode](https://github.com/mercedes-benz/odxtools#using-the-non-strict-mode). The resulting database object might or might not be very useful, though.

a better solution is allowing to specify a list of file names to [`load_file()`]()https://github.com/mercedes-benz/odxtools/blob/main/odxtools/load_file.py#L7) and its friends. Such patches would be highly welcome...

andlaus:
barring bugs, this has been implemented via #307. I'll close this for now...

PR #307, opened 2024-05-24
Improve the file loading mechansims

andlaus:
This PR refactors the file loading mechanism with the goal to be able to create databases from multiple constitutive files. It works by introducing a new user facing `odxtools.load_files(filename1, filename2, ...)` function.

Besides this, `odxtools.load_directory(dirname)` is added. This function loads all files contained in a directory into a single database. The main application is probably to directly load an unzipped PDX file. This is a quite convenient feature when developing tools or a dataset.


andlaus:
this fixes #224 and hopefully #283.

@QWander: can you check if the references are properly resolved if you load your dataset using
```
odxtools.load_files('myFile1.odx-d', 'myFile2.odx-d', ...)
```

odxtools/loadfile.py
```diff
@@ -0,0 +1,63 @@
...
+
+def load_odx_d_file(odx_d_file_name: str) -> Database:
+    db = Database()
+    db.add_odx_file(odx_d_file_name)
+    db.refresh()
+
+    return db
+
+
+def load_file(file_name: str) -> Database:
```
kayoub5, line 23: `load_file` feels like a useless alias of load_files

odxtools/loadfile.py
```diff
@@ -0,0 +1,63 @@
...
+        elif p.suffix.lower().startswith(".odx"):
+            db.add_odx_file(file_name)
+        elif p.name.lower() != "index.xml":
+            db.add_auxiliary_file(file_name)
+
+    db.refresh()
+    return db
+
+
+def load_directory(dir_name: str) -> Database:
```
kayoub5, line 47: `load_files` can simply be extended to check if input file is a folder with `os.path.isdir`

When it comes to Linux philosophy, a directory is also a file

odxtools/loadfile.py
andlaus: yeah, but we should try not to break too many people's code. (and using `load_file()` for multiple files feels a bit stange to me...)

odxtools/loadfile.py
andlaus: I thought about this too, but I think it is better not to load directories recursively. (I guess that this would be a too easily accessible footgun...)

andlaus: I agree that `load_directory()` can be basically seen as a convenience wrapper around `load_files()`. using the same logic, `load_files()` can basically be seen as a convenience wrapper for `Database`, though?!

odxtools/loadfile.py
```diff
@@ -0,0 +1,63 @@
...
+def load_file(file_name: str) -> Database:
+    if file_name.lower().endswith(".pdx"):
+        return load_pdx_file(file_name)
+    elif file_name.lower().endswith(".odx-d"):
+        return load_odx_d_file(file_name)
+    else:
+        raise RuntimeError(f"Could not guess the file format of file '{file_name}'!")
+
+
+def load_files(*file_names: str) -> Database:
```
kayoub5, line 32: Nice to have: allow users to provide a Path object or str

andlaus: fixed: https://github.com/mercedes-benz/odxtools/pull/307/commits/c008db2d5ceaf2e091643ffbe0eb22ff720aa967

kayoub5 (approved):

PR #310, opened 2024-06-04
Improve auxiliary file handling

andlaus:
This PR improves the handling of auxiliary files:

- Besides being read, these files are now also written.
- `ProgCode` objects now provide direct access to the contents of the referenced files. This required to back-link the `Database` object from `DiagLayer` which in turn required some custom `__deepcopy__()`  methods to avoid unexpected results like a deepcopy of a single diag layer containing the whole database.

Besides this, this PR first fixes the `mksomersaultmodifiedpdx.py` example: the somersaultecu has been slightly changed (by adding an auxiliary file for a single-ECU job), and thus the modified somersault ECU needed to be dragged along. Unfortunately the script has been broken for the last few weeks...


kayoub5:
@andlaus those are 3 unrelated topics in one PR, better to split them

odxtools/progcode.py
```diff
@@ -52,4 +56,9 @@ def _resolve_odxlinks(self, odxlinks: OdxLinkDatabase) -> None:
         pass
 
     def _resolve_snrefs(self, diag_layer: "DiagLayer") -> None:
-        pass
+        db = diag_layer.database
+
+        self._code = odxrequire(
+            db.auxiliary_files.get(self.code_file),
```
kayoub5, line 62: This will kill performance for flashing PDX files, since they can contain large binary data, better to provide a lazy API that only extracts the binary bytes when needed.

odxtools/progcode.py
```diff
@@ -20,6 +20,10 @@ class ProgCode:
     entrypoint: Optional[str]
     library_refs: List[OdxLinkRef]
 
+    @property
+    def code(self) -> bytes:
+        return self._code
```
kayoub5, line 25: Better to return an object with API similar to `importlib.resources.abc.Traversable`, most PDX files I have seen use `jar` files, for those the bytes are useless, and an actual file is needed.

odxtools/progcode.py
andlaus: it was my assumption that python uses a reference here? (a copy would be `bytes(db.auxiliary_files.get(self.code_file))`?) Maybe I'm wrong, though. Can you test or do you know a good reference to the documentation?

andlaus: okay, I ran two tests. The first was
```python
import tracemalloc

tracemalloc.start()

x = chr(0)*1000*1000
db = []
for i in range(999):
    #x = chr(i%256)*1000*1000
    db.append(x)

print(tracemalloc.get_traced_memory())

tracemalloc.stop()
```

and the second
```python
import tracemalloc

tracemalloc.start()

x = chr(0)*1000*1000
db = []
for i in range(999):
    #x = chr(i%256)*1000*1000
    db.append(x)

print(tracemalloc.get_traced_memory())

tracemalloc.stop()
```

for the first test, the memory consumption was basically identical to not having the `db` object (8832 bytes overhead for the list with 999 items), the second used almost exactly 999 times more memory than the first script (999.07 MB vs 1.01 MB). What I considered interesting is when the first script was modified to copy the items: this resulted in exactly the same memory consumption, so python seems to use a "copy on write" mechanism for strings...

odxtools/progcode.py
andlaus: (I hope I didn't misunderstand your point. If so: sorry :/)

andlaus:
> @andlaus those are 3 unrelated topics in one PR, better to split them

I agree in principle, but the auxiliary file stuff requires the first two patches, while the "provide database via the diaglayer" stuff would be unused without the `ProgCode` changes. Do you still want it to be split?

kayoub5:
> > @andlaus those are 3 unrelated topics in one PR, better to split them
> 
> I agree in principle, but the auxiliary file stuff requires the first two patches, while the "provide database via the diaglayer" stuff would be unused without the `ProgCode` changes. Do you still want it to be split?

Yes please, the changes are too much to review all at once

odxtools/progcode.py
kayoub5: Comment was on the wrong line, I am worried about memory consumed by the `aux_file_data = open(aux_file_name, "rb").read()` operation

andlaus:
okay, done: cf #311 and #312...

odxtools/progcode.py
andlaus: I think we can assume that odxtools are used on desktop class machines, i.e. there should be a few gigabytes of RAM available? What's IMO more worrying is the size of the uncompressed ODX XML files: I've git a PDX which contains an `*.odx-d` file exhibiting a size of > 650 MB, and parsing this to an `ElementTree` requires quite a bit of additional RAM, i.e. loading this file on a 4GB RaspberryPi is "fun"...

kayoub5: PDX file that ship ECU firmware can be even bigger, multiple gigabytes of binary that basically contain a whole OS and its applications, those kind of auxiliary_files should only be loaded when explicitly requested, otherwise you run out of RAM fast.

odxtools/progcode.py
```diff
@@ -52,4 +56,9 @@ def _resolve_odxlinks(self, odxlinks: OdxLinkDatabase) -> None:
         pass
 
     def _resolve_snrefs(self, diag_layer: "DiagLayer") -> None:
-        pass
+        db = diag_layer.database
```
kayoub5, line 59: It would be better to add db as argument to `_resolve_snrefs`, rather than a property of the DiagLayer

andlaus: Most SNREFs can only be resolved if the applicable diagnostic layer is known (e.g. TABLE-SNREFs, DIAG-COMM-SNREFs, etc.), but it is impossible to determine the relevant diaglayer given only the database: e.g., when resolving SNREFs within a service, it is completely unclear which of the diag layers that the respective service is part of ought to be used.

We probably should to rethink the SNREF resolution mechanism anyway: SNREFs lead to a really bad situation with inheritance where the object that is referenced depends on how a given object is accessed (cf #200). Example:

- DiagLayer `B` ("Base") defines a request `R` which references a DOP `Foo` via a short name reference
- Two DiagLayers `D1` and `D2` are derived from `B`. Both `D1` and `D2` redefine the DOP `Foo`
- In this case, `D1.R.Foo` ought to be different from `D2.R.Foo` even though `R` is the same object in both cases.

So far, I have not come up with an idea of how to implement this in a good way. The only solution which I've come up with is to make all objects in a database use a given diaglayer for their SNREFs, but that invalidates the database for any other diaglayer...

kayoub5: I am talking about adding it as an extra argument
```py
def _resolve_snrefs(self, diag_layer: "DiagLayer", database: "Database") -> None:
```
A more generic approach would be
```py
def _resolve_snrefs(self, lookup) -> None:
```
`lookup` will provide two functions `resolve_snref(name)` and `get_file(name)`

the implementation of those two functions could be adapted later to account for corner cases like #200

andlaus: I basically agree. I would propose to implement this by merging `_resolve_snrefs()` and `_resolve_odxlinks()` into `_resolve_references()` and passing it a context object which provides access to all relevant objects (OdxLinkDatabase, Database, Diaglayer, etc). This is quite a bit of work, though, and IMO thus a bit out of scope of this PR (i.e., improving the auxiliary file handling). An issue I can see with this approach is that the state these context objects must exhibit in order to call these methods is quite magical...

andlaus: another issue might be properly dealing with inheritance.

odxtools/progcode.py
andlaus: I see. fixed: https://github.com/mercedes-benz/odxtools/pull/310/commits/fe5423a7f2033a3fd9f46ab1c0d98e29cb744ae1

andlaus:
alright: with #311 and #312 merged, I rebased this on top of the latest master. I fixed two further issues as drive-bys (https://github.com/mercedes-benz/odxtools/pull/310/commits/76364d88355df30348ede7fdf598d0271118f585 and https://github.com/mercedes-benz/odxtools/pull/310/commits/63958a8eb26d85648369c830ab22208a6b710884 ), but I don't think that these warrant a separate PR...

kayoub5 (approved):

PR #312, opened 2024-06-05
DiagLayer: remember the database

andlaus:
this is a spin-off of #310: We provide a backlink from the diagnostic layer to the database which contains the layer in question (this is necessary to access the contents of auxiliary files from objects contained by the diag layer). Note that this seemingly trivial change is surprisingly difficult to implement, as it requires some custom `__deepcopy__()` methods, or else the whole database would be copied if some relatively trivial objects like `ParentRef` are copied...

### DIFF
--- a/odxtools/diaglayer.py
+++ b/odxtools/diaglayer.py
@@ -1,11 +1,12 @@
 # SPDX-License-Identifier: MIT
 import re
 import warnings
-from copy import copy
+from copy import copy, deepcopy
 from dataclasses import dataclass
 from functools import cached_property
 from itertools import chain
-from typing import Any, Callable, Dict, Iterable, List, Optional, Tuple, TypeVar, Union, cast
+from typing import (TYPE_CHECKING, Any, Callable, Dict, Iterable, List, Optional, Tuple, TypeVar,
+                    Union, cast)
 from xml.etree import ElementTree
 
 from deprecation import deprecated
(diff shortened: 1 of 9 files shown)


odxtools/diaglayer.py
```diff
@@ -125,7 +129,22 @@ def _resolve_odxlinks(self, odxlinks: OdxLinkDatabase) -> None:
 
         self.diag_layer_raw._resolve_odxlinks(odxlinks)
 
-    def _finalize_init(self, odxlinks: OdxLinkDatabase) -> None:
+    def __deepcopy__(self, memo: Dict[int, Any]) -> Any:
```
kayoub5, line 132: Where is the deep copy needed? I can't find it in the diff

andlaus: that's for the `mksomersaultmodifiedpdx.py` example.

andlaus: (because `deepcopy()` makes a copy of all attributes of an object, it is advisable not to copy the new `_database` attribute. In some sense, what we really want is a variant of `deepcopy()` that only copies the fields of dataclasses.)

kayoub5:
> @kayoub5: do you see any further issues or do you consider this ready to be merged?

main objection is https://github.com/mercedes-benz/odxtools/pull/310#discussion_r1628060365

kayoub5:
make the database property private, so you can remove it later once https://github.com/mercedes-benz/odxtools/pull/310#discussion_r1632861139 is done

kayoub5 (approved):

PR #313, opened 2024-06-11
introduce short name reference resolution contexts

andlaus:
this patch emerged from a discussion in #312. It is quite large but also pretty simple: instead of passing the relevant containers directly to the `_resolve_snref()` methods, we now put every object that is possibly required for short name resolution into a new `SnRefContext` object. Those parts of the context which are not applicable are simply `None` and the respective `_resolve_snref()` methods are responsible for ensuring that the parts of the context which they require are present...

### DIFF
--- a/odxtools/diaglayer.py
+++ b/odxtools/diaglayer.py
@@ -305,7 +305,7 @@ class DiagLayer:
         context = SnRefContext(database=database)
         context.diag_layer = self
         self.diag_layer_raw._resolve_snrefs(context)
-        #context.diag_layer = None
+        context.diag_layer = None
 
     #####
     # <convenience functionality>


odxtools/diaglayer.py
```diff
@@ -301,7 +302,10 @@ def _finalize_init(self, database: "Database", odxlinks: OdxLinkDatabase) -> Non
         # by the spec (So far, I haven't found any definitive
         # statement...)
         #####
-        self.diag_layer_raw._resolve_snrefs(self)
+        context = SnRefContext(database=database)
+        context.diag_layer = self
+        self.diag_layer_raw._resolve_snrefs(context)
+        #context.diag_layer = None
```
kayoub5, line 308: Commented out code

kayoub5: other than this LGTM

PR #318, opened 2024-07-04
Multiplexer coding fixes

andlaus:
This PR fixes the en- and decoding routines of multiplexers. The mistake was that we assumed that `multiplexer.byte_position` specifies the start position of the whole multiplexer object while in reality it only specifies the position of its content part.

Besides this, the PR contains a few smaller quality-of-life improvements:

- `DiagLayer.decode()` now re-raises encountered `DecodeError`s if the binary data is also not decodable as a global negative response. The old behavior accidentally was to never raise  `DecodeError`.
- The values of multiplexer parameters can now be specified using a 2-tuple or a list of length two in addition to a dictionary with a single entry
- The key of multiplexer parameters can now be specified by their raw integer value or the `MultiplexerCase` object in addition to the short name of the selected case
- The default case of multiplexers can now be selected by specifying an integer value which is not featured by any of the "ordinary" cases.


andlaus:
okay, I added a unit test for this functionality (I had to fix up a few mistakes in `test_diag_data_dictionary_spec` first) and rebased it on top of #319...

odxtools/odxtypes.py
```diff
@@ -28,7 +28,7 @@ def bytefield_to_bytearray(bytefield: str) -> bytearray:
 # sub-parameters, so this is a recursive type, and fields encompass
 # multiple items, so this can be a list of objects.
 TableStructParameterValue = Tuple[str, "ParameterValue"]
-ParameterValue = Union[AtomicOdxType, "ParameterValueDict", TableStructParameterValue,
+ParameterValue = Union[None, AtomicOdxType, "ParameterValueDict", TableStructParameterValue,
```
kayoub5, line 31: why?

andlaus: the reference to a structure  is optional for multiplexer cases. I thought that it is kind of intuitive to use `None` as the physical value...

andlaus: alternatively, we could use an empty dict,  but IMO `None` is a bit better...

andlaus: I've [changed my mind](https://github.com/mercedes-benz/odxtools/pull/318/commits/3f0add18bb82d8198c195379725860a75e0c85a9): always returning a dict will remove a "footgun" because the code which processes the decoded message does not need to expect `None`. thanks for bringing this up.

andlaus: it is included in https://github.com/mercedes-benz/odxtools/commit/3f0add18bb82d8198c195379725860a75e0c85a9

andlaus:
@kayoub5: do you see any hurdles for merging this?

kayoub5 (approved):

andlaus:
done

PR #321, opened 2024-07-15
implement de- and encoding of environment data descriptions

andlaus:
This PR implements en- and decoding of environment data descriptions, the fourth multiplexer mechanism specified by the ODX standard besides tables, NRC-CONSTs and multiplexers. This requires to keep track of the de-/encoded parameters because the environment data that ought to be used by a given environment data description depends on the value used for one of the preceding parameters.


odxtools/dtcdop.py
```diff
@@ -130,22 +130,35 @@ def decode_from_pdu(self, decode_state: DecodeState) -> ParameterValue:
             sdgs=[],
         )
 
-    @override
-    def encode_into_pdu(self, physical_value: Optional[ParameterValue],
-                        encode_state: EncodeState) -> None:
-        if isinstance(physical_value, DiagnosticTroubleCode):
-            trouble_code = physical_value.trouble_code
-        elif isinstance(physical_value, int):
+    def numerical_trouble_code(self, dtc_value: ParameterValue) -> int:
```
kayoub5, line 133: function names usually start with a verb

odxtools/dtcdop.py
```diff
@@ -130,22 +130,35 @@ def decode_from_pdu(self, decode_state: DecodeState) -> ParameterValue:
...
-            trouble_code = physical_value
-        elif isinstance(physical_value, str):
+            return dtc_value
+        elif isinstance(dtc_value, str):
             # assume that physical value is the short_name
-            dtcs = [dtc for dtc in self.dtcs if dtc.short_name == physical_value]
-            odxassert(len(dtcs) == 1)
-            trouble_code = dtcs[0].trouble_code
+            dtcs = [dtc for dtc in self.dtcs if dtc.short_name == dtc_value]
+            if len(dtcs) != 1:
```
kayoub5, line 142: Could it happen that length > 1 ?

odxtools/decodestate.py
```diff
@@ -46,6 +47,11 @@ class DecodeState:
     #: values of the table key parameters decoded so far
     table_keys: Dict[str, "TableRow"] = field(default_factory=dict)
 
+    #: List of parameters that have been decoded so far. The journal
+    #: is used by some types of parameters which depend on the values of
+    #: other parameters; i.e., environment data description parameters
+    journal: List[Tuple["Parameter", Optional[ParameterValue]]] = field(default_factory=list)
```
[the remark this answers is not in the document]
andlaus: we need the sequence in which the parameters were processed, i.e. that would require at least `OrderedDict`. That said, I'm not sure how such a lookup would be realized and also parameters frequently appear more than once, e.g., for fields (which are repeated structure objects)...

odxtools/dtcdop.py
andlaus: I don't think so: in this case, it would be impossible to decide which of the environment datas that apply for the DTC ought to be used. (I suppose that this constitutes an incorrect PDX file?)

odxtools/dtcdop.py
andlaus: do you prefer `get_numerical_trouble_code()`, `convert_to_numerical_trouble_code()` or something else?

kayoub5: Either would be fine

odxtools/dtcdop.py
kayoub5: Does the specs say anything about that?

odxtools/decodestate.py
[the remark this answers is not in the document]
kayoub5: You are using the journal with `reversed`, meaning only the last value is actually needed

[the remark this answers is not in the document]
andlaus: yes, but then a different name than `.journal` is also required IMO. (I think that conceptually, the concept of a "change log" for en-/decoding is quite nice.) Performance-wise I don't think that it will matter much because responses containing environment data descriptions are probable quite rare (e.g., I don't have any files which use them and so far we have not gotten any bug reports about missing en-/decoding support.)

odxtools/dtcdop.py
andlaus: okay, I renamed it to [`convert_to_numerical_trouble_code()`](https://github.com/mercedes-benz/odxtools/pull/321/commits/4e79723736bb6d233bbe850cb439860282ead605)

odxtools/dtcdop.py
andlaus: I'm not aware that it explicitly mentions normal multiple environment datas matching the same DTC, but it says that there can only be a single ALL-DATA environment data. Since multiple ALL-DATA environment datas is exactly the same problem as multiple "normal" ones for a given DTC, I infer that this is forbidden. (in which order shall these environment datas be processed anyway?)

andlaus:
@kayoub5: do you see any further issues here?

odxtools/dtcdop.py
andlaus: ok, sorry: I was in the wrong mental context for my previous two comments (interestingly they are still applicable to some extend): more than one DTC with the same short name cannot happen in valid ODX datasets because short names are required to be unique in their respective context. (if they were not, SNREFs would not work...)

kayoub5 (approved):

PR #322, opened 2024-07-23
More small fixes

andlaus:
This PR contains fixes for a few small issues which I've recently stumbled over, in particular minor improvements regarding the ODXLINK mechanism.

### DIFF
--- a/odxtools/odxlink.py
+++ b/odxtools/odxlink.py
@@ -26,7 +26,11 @@ class OdxDocFragment:
         return self.doc_name == other.doc_name
 
     def __hash__(self) -> int:
-        return hash(self.doc_name) + hash(self.doc_type)
+        return hash(self.doc_name)
+
+    def __equals__(self, other: "OdxDocFragment") -> bool:
+        return self.doc_name == other.doc_name and (self.doc_type is None and other.doc_type is None
+                                                    or self.doc_type == other.doc_type)
 
 
 @dataclass(frozen=True)
--- a/odxtools/odxlink.py
+++ b/odxtools/odxlink.py
@@ -125,6 +125,7 @@ class OdxLinkRef:
 
         id_ref = et.attrib.get("ID-REF")
         if id_ref is None:
+            odxraise(f"Tag {et.tag} is not a ODXLINK reference")
             return None
 
         doc_ref = et.attrib.get("DOCREF")
@@ -158,9 +159,9 @@ T = TypeVar("T")
 
 class OdxLinkDatabase:
     """
-    A database holding all objects which ehibit OdxLinkIds
+    A database holding all objects which exhibit OdxLinkIds
 
-    This can resolve references to such.
+    This can resolve ODXLINK references.
     """
 
     def __init__(self) -> None:
@@ -186,7 +187,7 @@ class OdxLinkDatabase:
             if doc_frag_db is None:
                 # No object featured by the database uses the document
                 # fragment mentioned by the reference. This should not
-                # happen for correct databases...
+                # happen for correct ODX databases...
                 warnings.warn(
                     f"Warning: Unknown document fragment {ref_frag} "
                     f"when resolving reference {ref}",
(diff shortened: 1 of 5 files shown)


odxtools/odxlink.py
```diff
@@ -26,7 +26,11 @@ def __eq__(self, other: Any) -> bool:
         return self.doc_name == other.doc_name
 
     def __hash__(self) -> int:
-        return hash(self.doc_name) + hash(self.doc_type)
+        return hash(self.doc_name)
+
+    def __equals__(self, other: "OdxDocFragment") -> bool:
+        return self.doc_name == other.doc_name and (self.doc_type is None and other.doc_type is None
+                                                    or self.doc_type == other.doc_type)
```
kayoub5, line 33: This can cause issues, since you can have A == B, and B == C, but A != C

andlaus: I know; this is why I described this as a hack in the PR description. The intention is to be able to is `OdxDocFragment` objects as keys in dictionaries and getting a result if the document type is unspecified for the object used as index (a la `self._db[doc_frag]` in `OdxLinkDatabase`. if multiple document fragments using the same name but different document types exist, I'm willing to let the result to be undefined). The "clean" solution would be to to use a custom comparison function for these dictionaries, but I have not found a way to do this in a satisfactory manner...

andlaus: okay; it turns out that a custom comparison function is not necessary here: the spec mandates that the DOCREF and the DOCTYPE attributes are both either present or not present in references and if they are not present, we have to use the "local" document fragments where the reference is located. In conjunction this means that we get away with requiring that the `doc_type` attribute is always present and always including it in comparisons.

(also, overloading comparisons works by redefining `__eq__()`, not `__equals__()`. I have no idea what hit me when I did this.)

anyway, [fixed](https://github.com/mercedes-benz/odxtools/pull/322/commits/5e1079731e365a752aa07a165e13787b1924696b). do you see any further issues with this PR @kayoub5?

odxtools/odxlink.py
```diff
@@ -4,29 +4,14 @@
 from typing import Any, Dict, Iterable, List, Optional, Type, TypeVar, overload
 from xml.etree import ElementTree
 
-from .exceptions import OdxWarning, odxassert, odxraise
+from .exceptions import OdxWarning, odxassert, odxraise, odxrequire
 from .nameditemlist import OdxNamed, TNamed
 
 
 @dataclass(frozen=True)
 class OdxDocFragment:
```
kayoub5, line 12: NamedTuple is better if you want to use this as dict key

andlaus: Hm, I'm not so sure: `NamedTuple` seems to be slightly faster and come with a slightly lower memory footprint but it severely limits flexibility because AFAICS there is no possibility to add custom methods like `from_et()`, i.e., it is not feasible to use it for e.g., `OdxLinkId` and `OdxLinkRef` which would IMO benefit quite a bit more from making them `NamedTuple`s than `OdxDocFragment`. That said, maybe I just do not understand `NamedTuple` fully yet?

andlaus: after a bit of googling, it turns out that it is possible to [create `NamedTuples` with user defined methods](https://stackoverflow.com/questions/3223236/creating-a-namedtuple-with-a-custom-hash-function). That said, inheritance is [quite akward](https://stackoverflow.com/questions/42385916/inheriting-from-a-namedtuple-base-class) for named tuples and -- more importantly -- requires quite a bit of manual code because the constructors of child classes must be aware of *all* constructor arguments required by the base class, i.e., I have not found a way of how an analog to `dataclass_fields_asdict()` can be implemented. (My thinking is that if we were to go the `NamedTuple` route thoroughly, we would need to subclass `OdxLinkRef` for e.g. `DynEndDopRef` and `RelatedDiagCommRef`. And if we are not doing this consistently, it is IMO better not to take this route at all...)

kayoub5 (approved):

PR #323, opened 2024-07-26
Implement diagnostic variables

andlaus:
In true "design by committee" spirit, the ODX specification supports to layer an entirely different communication paradigm over its default RPC-based model: Diagnostic variables. In this model, communication with the ECU is done by reading from and writing to a set of "variables". ODX glues this communication approach on its "normal" service-oriented model by specifying a set of variables plus setter and getter services for each.

Note that, since odxtools is not concerned with communicating with ECUs directly, odxtools can only expose the required information for this, not implement this approach itself.

Besides the diagnostic variables infrastructure, this patch also adds the code to internalize the information necessary for dynamically defined messages (`DYN-DEFINED-SPEC`), i.e., messages which exhibit formats that are defined at runtime instead of being known a-priori.

Note that, since I don't have any data file which use either of these features, your mileage may vary.

### DIFF
--- a/odxtools/commrelation.py
+++ b/odxtools/commrelation.py
@@ -1,0 +1,133 @@
+# SPDX-License-Identifier: MIT
+from dataclasses import dataclass
+from enum import Enum
+from typing import Any, Dict, List, Optional
+from xml.etree import ElementTree
+
+from .description import Description
+from .diagcomm import DiagComm
+from .diagservice import DiagService
+from .exceptions import odxraise, odxrequire
+from .odxlink import OdxDocFragment, OdxLinkDatabase, OdxLinkId, OdxLinkRef, resolve_snref
+from .parameters.parameter import Parameter
+from .snrefcontext import SnRefContext
+
+
+class CommRelationValueType(Enum):
+    CURRENT = "CURRENT"
+    STORED = "STORED"
+    STATIC = "STATIC"
+    SUBSTITUTED = "SUBSTITUTED"
+
+
+@dataclass
+class CommRelation:
+    description: Optional[Description]
+    relation_type: str
+    diag_comm_ref: Optional[OdxLinkRef]
+    diag_comm_snref: Optional[str]
+    in_param_if_snref: Optional[str]
+    in_param_if_snpathref: Optional[str]
+    out_param_if_snref: Optional[str]
+    out_param_if_snpathref: Optional[str]
+    value_type_raw: Optional[CommRelationValueType]
+
+    @property
+    def diag_comm(self) -> DiagComm:
+        return self._diag_comm
+
+    @property
+    def in_param_if(self) -> Optional[Parameter]:
+        return self._in_param_if
+
+    @property
+    def out_param_if(self) -> Optional[Parameter]:
+        return self._out_param_if
+
+    @property
+    def value_type(self) -> CommRelationValueType:
+        if self.value_type_raw is None:
+            return CommRelationValueType.CURRENT
+
+        return self.value_type_raw
+
+    @staticmethod
+    def from_et(et_element: ElementTree.Element, doc_frags: List[OdxDocFragment]) -> "CommRelation":
+        description = Description.from_et(et_element.find("DESC"), doc_frags)
+        relation_type = odxrequire(et_element.findtext("RELATION-TYPE"))
+
+        diag_comm_ref = OdxLinkRef.from_et(et_element.find("DIAG-COMM-REF"), doc_frags)
+        diag_comm_snref = None
+        if (diag_comm_snref_elem := et_element.find("DIAG-COMM-SNREF")) is not None:
+            diag_comm_snref = odxrequire(diag_comm_snref_elem.get("SHORT-NAME"))
+
+        in_param_if_snref = None
+        if (in_param_if_snref_elem := et_element.find("IN-PARAM-IF-SNREF")) is not None:
+            in_param_if_snref = odxrequire(in_param_if_snref_elem.get("SHORT-NAME"))
+
+        in_param_if_snpathref = None
+        if (in_param_if_snpathref_elem := et_element.find("IN-PARAM-IF-SNPATHREF")) is not None:
+            in_param_if_snpathref = odxrequire(in_param_if_snpathref_elem.get("SHORT-NAME-PATH"))
+
+        out_param_if_snref = None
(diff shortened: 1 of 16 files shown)


odxtools/commrelation.py
```diff
@@ -0,0 +1,133 @@
...
+        if not isinstance(service, DiagService):
+            odxraise(f"DIAG-VARIABLE references non-service {type(service).__name__} "
+                     f"diagnostic communication")
+
+        self._in_param_if = None
+        if self.in_param_if_snref is not None:
+            self._in_param_if = resolve_snref(self.in_param_if_snref,
+                                              odxrequire(service.request).parameters, Parameter)
+        elif self.in_param_if_snpathref is not None:
+            odxraise("Resolving SNPATHREFS", NotImplementedError)
```
kayoub5, line 121: I don't like the idea of something not implemented blocking the loading of a file

non strict mode should be used for invalid files, not as a workaround for uncomplete implementation

odxtools/diaglayerraw.py
```diff
@@ -178,6 +185,28 @@ def from_et(et_element: ElementTree.Element, doc_frags: List[OdxDocFragment]) ->
...
+        diag_variables_raw: List[Union[DiagVariable, OdxLinkRef]] = []
+        if (dv_elems := et_element.find("DIAG-VARIABLES")) is not None:
+            for dv_proxy_elem in dv_elems:
+                dv_proxy: Union[OdxLinkRef, DiagVariable]
+                if dv_proxy_elem.tag == "DIAG-VARIABLE-REF":
+                    dv_proxy = OdxLinkRef.from_et(dv_proxy_elem, doc_frags)
+                elif dv_proxy_elem.tag == "DIAG-VARIABLE":
+                    dv_proxy = DiagVariable.from_et(dv_proxy_elem, doc_frags)
+                else:
+                    odxraise()
```
kayoub5, line 197: No message?

odxtools/diagvariable.py
```diff
@@ -0,0 +1,109 @@
...
+    sdgs: List[SpecialDataGroup]
+    is_read_before_write_raw: Optional[bool]
+
+    @property
+    def variable_group(self) -> VariableGroup:
+        return self._variable_group
+
+    @property
+    def table_row(self) -> Optional[TableRow]:
+        raise NotImplementedError
```
kayoub5, line 40: Better not to put a property if it's just not implemented, add the property when you actually implement it

tests/test_unit_spec.py
```diff
@@ -207,7 +207,9 @@ def test_resolve_odxlinks(self) -> None:
             ecu_variant_patterns=[],
             comparam_spec_ref=None,
             prot_stack_snref=None,
-        )
+            diag_variables_raw=[],
+            variable_groups=NamedItemList(),
+            dyn_defined_spec=None)
```
kayoub5, line 212:
```suggestion
            dyn_defined_spec=None
        )
```

odxtools/commrelation.py
andlaus: if we would internalize the SNPATHREF here, we would get an exception in non-strict mode as soon as `.in_param_if` was accessed.  I'm ambivalent about whether this is better: on one hand, the exception will only be raised if this feature is used, on the other hand it might be raised in unexpected contexts (and only for some files) which might make the resulting code more unreliable. If you still prefer to delay raising the exception, I don't mind changing the code, though...

odxtools/diaglayerraw.py
andlaus: This is forbidden by the XSD, but I added a message. Thanks for the nudge!

odxtools/diagvariable.py
andlaus: good point. fixed. :)

tests/test_unit_spec.py
andlaus: I can remove the comma, but our coding style police (`yapf`) enforces the closing bracket to be on a separate line. (for many such list I like to use a comma for the last argument because it allows to add more items without having to modify the last line. This is currently inconsistent and just a IMO very unimportant detail..)

odxtools/commrelation.py
kayoub5: At this phase of the feature, simply don't process in_param_if_snpathref at all, don't check it, don't assert on it

You may however warn, that it is not supported

andlaus: note that the `._{in,out}_param_if` properties still exists but accessing them will raise an attribute error if if the respective parameter has been specified via SNPATHREF. This is because these parameters are essential for implementing the diagnostic variable functionality (if my understanding of this is correct) and the spec thus requires them to be defined by either SNREF or SNPATHREF...

odxtools/commrelation.py
```diff
@@ -0,0 +1,126 @@
...
+        else:
+            odxraise(f"No IN-PARAM-IF referenced (note that SNPATHREFs are not yet supported)")
+
+        self._out_param_if = None
+        if self.out_param_if_snref is not None:
+            self._out_param_if = resolve_snref(self.out_param_if_snref,
+                                               odxrequire(service.positive_responses[0]).parameters,
+                                               Parameter)
+        else:
+            odxraise(f"No OUT-PARAM-IF referenced (note that SNPATHREFs are not yet supported)")
```
kayoub5, line 126: This does not make the situation any better, odxraise should only and only be used when the ODX file is wrong

andlaus: yeah, that was my previous point: we do not implement SNPATHREFs (yet?) and thus we only have the following options:

1) do not resolve the param objects at all. this would make it pretty hard to actually use the functionality and would be contradictory to basically any other place where references are involved. 
2) make it work for the approximately 95% of the cases where SNREFs are used and
  (a) complain about SNPATHREFs during load time when in strict mode
  (b) raise an (almost certainly unexpected) exception once the object referenced via SNPATHREF is to be accessed by the user code.

originally, this PR was in camp (2a), currently it is using (2b). I think (2b) is not as bad as it looks, though, because I'm pretty sure that 99% of all odxtools users will never touch the diagnostic variables functionality. That said, maybe I missed something...

andlaus: ok, I guess I misunderstood what you meant: you said even with the latest version of the PR we were essentially still in camp (2a)? I removed these `odxraise` statements, so we are hopefully in (2b) now...

PS: do you have files which define diag variables? if yes, it would be cool if you could test if this code works...

kayoub5 (approved):

Issue #324, opened 2024-07-30
Not support TABLE-SNREF?


lz4v4:
when i change the table reference type in the example "somersault.pdx" from "`<TABLE-REF ID-REF="somersault.table.last_flip_details" />`" to "`<TABLE-SNREF SHORT-NAME="last_flip_details"/>`"

this error appears:
```
  File "/Library/Python/3.8/site-packages/odxtools/parameters/tablekeyparameter.py", line 103, in _resolve_snrefs
    self._table = resolve_snref(self.table_snref, tables, Table)
NameError: name 'Table' is not defined
```

andlaus:
thanks for your report. can you test if #325 fixes your issue?

lz4v4:
> thanks for your report. can you test if #325 fixes your issue?

I've tested it, and the fix in #325 works perfectly. Thanks a lot!"

PR #325, opened 2024-07-31
tablekeyparameter: fix short name reference resolution

andlaus:
unfortunately, the `Table` and `TableRow` classes are only available during type checking (i.e., not during normal execution) because this would lead to circular imports.

this should fix https://github.com/mercedes-benz/odxtools/issues/324.

### DIFF
--- a/odxtools/parameters/tablekeyparameter.py
+++ b/odxtools/parameters/tablekeyparameter.py
@@ -99,13 +99,20 @@ class TableKeyParameter(Parameter):
 
         if self.table_snref is not None:
             tables = odxrequire(context.diag_layer).diag_data_dictionary_spec.tables
-            self._table = resolve_snref(self.table_snref, tables, Table)
+            if TYPE_CHECKING:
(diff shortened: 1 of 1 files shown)


kayoub5:
Not possible to just remove `if TYPE_CHECKING:` in line 17 ?

andlaus:
unfortunately not. this leads to cylic imports...

andlaus:
@lz4v4: have you tested this?

kayoub5:
Do the import inside `_resolve_snrefs`, that way it's lazy and does not cause circular reference

andlaus:
> Do the import inside _resolve_snrefs, that way it's lazy and does not cause circular reference

I'd prefer not to since I'm unsure about the performance implications. (and IMO it is very inelegant as well...)

anyway, since @lz4v4 [confirmed](https://github.com/mercedes-benz/odxtools/issues/324#issuecomment-2264641619) that the patch works, let's merge it for now and maybe switch to another solution later.

PR #326, opened 2024-07-31
refactor `DiagLayer` into a class hierarchy

andlaus:
Instead of a single class for all types of diagnostic layers, ODX specifies a hierarchy of types: ECU shared datas, protocols, functional groups, base variants and ECU variants where all but ECU shared data layers derive from a "hierarchy element" to indicate that these may be a member of a inheritance hierarchy defined using `PARENT-REF`s. This patch implements this approach.


odxtools/multiplexercase.py
```diff
@@ -71,7 +71,7 @@ def _mux_case_resolve_odxlinks(self, odxlinks: OdxLinkDatabase, *,
 
     def _resolve_snrefs(self, context: SnRefContext) -> None:
         if self.structure_snref:
-            ddds = odxrequire(context.diag_layer).diag_data_dictionary_spec
+            ddds = odxrequire(odxrequire(context.diag_layer).diag_data_dictionary_spec)
```
kayoub5, line 74: please move drive-by fixes to their own PR, this PR is massive and un-reviwable as is.

kayoub5, line 74: why would a diag layer be none here?

odxtools/parentref.py
```diff
@@ -10,7 +10,7 @@
 from .utils import dataclass_fields_asdict
 
 if TYPE_CHECKING:
-    from .diaglayer import DiagLayer
```
kayoub5, line 13: keep a placeholder diaglayer.py file that just imports diaglayers.diaglayer content, in order to reduce the ammout of unrelated refactoring in this PR

examples/data/SAE_J2411_SWCAN_CPSS.odx-cs
```diff
@@ -0,0 +1,315 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<!-- edited with XMLSPY v2004 rel. 2 U (http://www.xmlspy.com) by Christian Hecker (Softing AG) -->
```
kayoub5, line 2: what's this file?

odxtools/diaglayers/ecuvariant.py
```diff
@@ -0,0 +1,122 @@
...
+
+    #######
+    # <properties subject to value inheritance>
+    #######
+    @property
+    def diag_variables(self) -> NamedItemList[DiagVariable]:
+        return self._diag_variables
+
+    @property
+    def variable_groups(self) -> NamedItemList[VariableGroup]:
```
kayoub5, line 53: those are exclusive to ecu variants?

odxtools/diaglayers/ecuvariant.py
```diff
@@ -0,0 +1,122 @@
...
+        super()._compute_value_inheritance(odxlinks)
+
+        self._diag_variables = NamedItemList(self._compute_available_diag_variables(odxlinks))
+        self._variable_groups = NamedItemList(self._compute_available_variable_groups(odxlinks))
+
+    def _compute_available_diag_variables(self,
+                                          odxlinks: OdxLinkDatabase) -> Iterable[DiagVariable]:
+
+        def get_local_objects_fn(dl: DiagLayer) -> Iterable[DiagVariable]:
+            if not hasattr(dl.diag_layer_raw, "diag_variables"):
```
kayoub5, line 100: hasattr in object oriented programming is usually a sign of a bad class diagram design

odxtools/diaglayers/protocol.py
```diff
@@ -0,0 +1,62 @@
...
+        """
+
+        result = super().__deepcopy__(memo)
+
+        # note that the self.protocol_raw object is *not* copied at
+        # this place because the attribute points to the same object
+        # as self.diag_layer_raw.
+        result.protocol_raw = deepcopy(self.protocol_raw, memo)
+
+        return result
```
kayoub5, line 62: I am seeing almost identical copy-pasted classes, and still don't see the benefit in term of making things simpler

tests/test_decoding.py
```diff
@@ -499,8 +495,8 @@ def test_decode_nrc_const(self) -> None:
             sdgs=[],
         )
 
-        diag_layer_raw = DiagLayerRaw(
-            variant_type=DiagLayerType.BASE_VARIANT,
+        ecu_variant_raw = EcuVariantRaw(
+            variant_type=DiagLayerType.ECU_VARIANT,
```
kayoub5, line 499: why?

tests/test_decoding.py
```diff
@@ -721,7 +712,7 @@ def test_decode_request_coded_const_undefined_byte_position(self) -> None:
                 "coded_const_parameter_4": 0x56,
             },
         )
-        decoded_message = diag_layer.decode(coded_message)[0]
+        decoded_message = ecu_variant.decode(coded_message)[0]
```
kayoub5, line 715: please do those renaming changes in a follow up PR

examples/data/SAE_J2411_SWCAN_CPSS.odx-cs
andlaus: this is a definition of communication parameters which is included in the ZIP file of the ASAM standard and which has been missing so far. The reason why this is now required is that the `CPS_ISO_15765_3_on_ISO_15765_2` comparam spec is now [internalized](https://github.com/mercedes-benz/odxtools/pull/326/files#diff-659a733f80deab14348f5c5c4b5ff936732c06257d5c6d8decf86518eb50735fR2397-R2403) by  `somersaultecu.py` and it features a reference to the `SAE_J2411_SWCAN` comparam subset which is contained in the missing file. I should factor these changes out into the separate "drive bys PR" but it is hard to catch such things when having stared at the code for more than a week. Thanks for the catch!

odxtools/multiplexercase.py
andlaus: `diag_data_dictionary_spec` is optional according to the XSD, but so far it has been mandatory in `DiagLayer` objects...

odxtools/matchingparameter.py
```diff
@@ -12,7 +12,7 @@
 class MatchingParameter:
     """According to ISO 22901, a MatchingParameter contains a string value identifying
     the active ECU variant. Moreover, it references a DIAG-COMM via snref and one of its
-    positiv response's OUT-PARAM-IF via snref or snpathref.
+    positive response's OUT-PARAM-IF via snref or snpathref.
```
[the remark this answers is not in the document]
andlaus: I'll make a second such PR (the diag variables stuff was actually the first), but a lot of the changes in this PR are interdependent (an trivial changes such as this often slip through the cracks...)

odxtools/parentref.py
andlaus: I think that this would only make things more confusing because most of the import-related code churn stems from the fact that additional classes need to be imported and IMO the canonical locations should be used for these anyway...

odxtools/diaglayers/ecuvariant.py
andlaus: nope, but the XSD defines them separately for ECU-SHARED-DATA, FUNCTIONAL-GROUP, BASE-VARIANT and ECU-VARIANT layers (note: not for PROTOCOL). To avoid confusion, I decided to simply match the spec even if I think that their approach is rather inelegant...

odxtools/diaglayers/ecuvariant.py
andlaus: yeah, but all diag layer types except `Protocol` exhibit this property. Alternatively, `isinstance(dl, (EcuSharedData, FunctionalGroup, BaseVariant, EcuVariant))` could be used here and for the other diag layer types which exhibit diag variables, but IMO this is even worse than `hasattr`...

odxtools/diaglayers/ecuvariantraw.py
```diff
@@ -0,0 +1,123 @@
...
+from ..odxlink import OdxDocFragment, OdxLinkDatabase, OdxLinkId, OdxLinkRef
+from ..parentref import ParentRef
+from ..snrefcontext import SnRefContext
+from ..utils import dataclass_fields_asdict
+from ..variablegroup import VariableGroup
+from .hierarchyelementraw import HierarchyElementRaw
+
+
+@dataclass
+class EcuVariantRaw(HierarchyElementRaw):
```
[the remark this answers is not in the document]
andlaus: not quite: the reason why we have the split between "raw" and "full" layers is that the "full" layers provide more objects than they define directly because of the various inheritance mechanisms. IOW, it is impossible to reconstruct the contents of the original XML file using just full layers. I agree this makes it quite complex and error prone, but so far I could not come up with a better solution...

odxtools/diaglayers/protocol.py
andlaus: the purpose of this PR is not to make this stuff simpler, it is to cling to the spec and allowing to eliminate a few systematic errors this way. E.g., if the spec says that a protocol ought to be referenced, we can now expect a `Protocol` class instead of a `DiagLayer`. Also, attributes which are not present in the specification do no longer exist in our classes (e.g., parent refs in ECU-SHARED-DATA)

tests/test_decoding.py
andlaus: because ECU variants are the only kinds of diag layers which are supposed to directly describe of how to  communicate with a concrete ECU. (base variants are just helpers from which concrete ECU variants can be derived from)

odxtools/diaglayers/ecuvariant.py
kayoub5: python allows multi inheritance, so move them to shared class

odxtools/diaglayers/ecuvariant.py
kayoub5: isinstance is better than hasattr, you could use typing `Protocol` as well instead of listing all of them.

tests/test_decoding.py
kayoub5: In practice, not quite, I have seen multiple cases where a Base variant is used to provide communication information for a whole group of ECUs

Usually it goes something like
* ECU variant is used for flashing, or to describe features specific to one ECU model.
* Base variant is used when describing features that are valid for multiple ECU models/variants

I have seen PDX files that have Base variants, but no ECU variants at all, and those are used to communicate with an entire group of cars, but provide only the generic shared diagnostic jobs

odxtools/diaglayers/ecuvariant.py
andlaus: I think that this would be quite confusing for people who try to match the XSD with the python code. (if the specification did it this way, I would be all for it...)

tests/test_decoding.py
andlaus: yes, base variants describe common functionality for a range of ECU variants and can in principle also be used for communication, but if you have a piece of hardware in front of you, I'd strongly advise to use the ECU-Variant that targets the firmware version which is currently flashed on that hardware so that the quirks of that specific version can be covered. (I think it is plausible that some vendors do not clearly distinguish between base variants and ECU variants, but I think that `odxtools`'s unit tests should adhere to the intended way unless the test is explicitly about non-conformant behavior.)

tests/test_decoding.py
andlaus: ok, I hope I caught the most severe drive-bys in #327 (since everything is intermingeled, the present PR is build on top of it...)

tests/test_decoding.py
kayoub5: The idea is that the same API available to ECU variant, should also be available to Base variant

odxtools/diaglayers/protocol.py
kayoub5: ODX model is for serialization, for actual diagnostic encoding/decoding, you may want to take a look at the standard D-PDU API model.

odxtools/cli/_print_utils.py
```diff
@@ -243,8 +243,11 @@ def print_dl_metrics(variants: List[DiagLayer], print_fn: Callable[..., Any] = p
         name.append(variant.short_name)
         type.append(variant.variant_type.value)
         num_services.append(len(all_services))
-        num_dops.append(len(variant.diag_data_dictionary_spec.data_object_props))
-        num_comparam_refs.append(len(variant.comparam_refs))
+        if (ddds := variant.diag_data_dictionary_spec) is not None:
+            num_dops.append(len(ddds.data_object_props))
+        else:
+            num_dops.append(0)
+        num_comparam_refs.append(len(getattr(variant, "comparams_refs", [])))
```
kayoub5, line 250: What's going on here?

odxtools/diagcomm.py
```diff
@@ -216,12 +216,12 @@ def _resolve_snrefs(self, context: SnRefContext) -> None:
...
             self._protocols = NamedItemList([
-                resolve_snref(prot_snref, diag_layer.protocols, DiagLayer)
+                resolve_snref(prot_snref, getattr(diag_layer, "protocols", []), Protocol)
                 for prot_snref in self.protocol_snrefs
             ])
         else:
             diag_layer = odxrequire(context.diag_layer)
             self._protocols = NamedItemList([
-                resolve_snref(prot_snref, diag_layer.protocols)
+                resolve_snref(prot_snref, getattr(diag_layer, "protocols", []))
```
kayoub5, line 225: Avoid `getattr`, it allows for typing bugs that mypy can't catch, use isinstance instead

odxtools/diaglayers/ecuvariant.py
```diff
@@ -0,0 +1,122 @@
...
+from ..odxlink import OdxDocFragment, OdxLinkDatabase, OdxLinkRef
+from ..parentref import ParentRef
+from ..variablegroup import HasVariableGroups, VariableGroup
+from .diaglayer import DiagLayer
+from .ecuvariantraw import EcuVariantRaw
+from .hierarchyelement import HierarchyElement
+
+
+@dataclass
+class EcuVariant(HierarchyElement):
```
kayoub5, line 23: Anything that BaseVariant have, should also be available in EcuVariant
```suggestion
class EcuVariant(BaseVariant):
```

odxtools/diaglayers/ecuvariant.py
```diff
@@ -0,0 +1,122 @@
...
+
+    @property
+    def ecu_variant_raw(self) -> EcuVariantRaw:
```
kayoub5, line 26: Why expose the raw data?

odxtools/diaglayers/ecuvariant.py
```diff
@@ -0,0 +1,122 @@
...
+        return cast(EcuVariantRaw, self.diag_layer_raw)
+
+    @property
+    def diag_variables_raw(self) -> List[Union[DiagVariable, OdxLinkRef]]:
```
kayoub5, line 30: The raw part should be internal, why expose it as a property ?

odxtools/diaglayers/ecuvariantraw.py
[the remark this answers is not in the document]
kayoub5: the "raw" part could be shared as far as I have seen the code, you don't need a new class for each "raw" type

odxtools/diaglayers/basevariant.py
```diff
@@ -0,0 +1,126 @@
...
+                return []
+
+            return dl.diag_layer_raw.diag_variables  # type: ignore[no-any-return]
+
+        def not_inherited_fn(parent_ref: ParentRef) -> List[str]:
+            return parent_ref.not_inherited_variables
+
+        return self._compute_available_objects(get_local_objects_fn, not_inherited_fn)
+
+    def _compute_available_variable_groups(self,
```
kayoub5, line 114: I see copy-paste of this function and similar "compute" functions, move them to a helper class

odxtools/diagvariable.py
```diff
@@ -17,6 +18,14 @@
 from .variablegroup import VariableGroup
 
 
+@runtime_checkable
+class HasDiagVariables(typing.Protocol):
```
kayoub5, line 22: What layers don't have diag variables?

odxtools/multiplexercase.py
kayoub5: We are creating an empty spec when XML does not have one, right?

andlaus: > We are creating an empty spec when XML does not have one, right?

ok, maybe I misunderstood you here: currently, we indeed always create a DDDS object in the diag layer. after this PR, this is not the case for for ECU-SHARED-DATA layers. probably I should change that (good point, thanks!)

andlaus: changed in https://github.com/mercedes-benz/odxtools/pull/326/commits/d15213b31a0a24f4b124acf39baabcc744f7620e

odxtools/diaglayers/ecuvariantraw.py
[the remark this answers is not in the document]
andlaus: I'd prefer to keep it simple and stupid even if this comes with a few copy-and-pasted lines. (who knows, maybe ODX 2.3 will modify one of the layers but not the others..)

odxtools/diaglayers/protocol.py
andlaus: I use this ODX data to talk to ECUs all the time. as far as I know, the PDX files are even canonical source of diagnostics data uppon which all tools except candela studio build uppon...

tests/test_decoding.py
andlaus: Ah, I see: your point was primarily about test coverage? fixed (https://github.com/mercedes-benz/odxtools/pull/326/commits/5038cf69ce6ad7155b5e2f3a50d1d97177c3f42a), thanks!

odxtools/cli/_print_utils.py
andlaus: not all diag layers provide `.comparam_refs` anymore. (Also, the `compare` tool is in serious need of refactoring...)

odxtools/diaglayers/ecuvariant.py
andlaus: I agree, but unfortunately this is not how the spec defines things:

Screenshot_20240805_183755

odxtools/diaglayers/ecuvariant.py
andlaus: because some low-level code might need to know what data was defined by the XML and in order to make typecheckers (`mypy`) happy, we cannot use `.diag_layer` to access `EcuVariant`-specific attributes...

odxtools/diaglayers/ecuvariant.py
andlaus: basically just to make the API of `EcuVariant` a superset of `EcuVariantRaw` (i.e., substituting `EcuVariantRaw` objects by `EcuVariant`  ones should just work without any changes.)

odxtools/diagvariable.py
andlaus: AFAICS just `Protocol`.

odxtools/diaglayers/ecuvariant.py
kayoub5: "might" or "do need" ? if "might" then delete, you are not gonna need it, see https://en.m.wikipedia.org/wiki/You_aren't_gonna_need_it

odxtools/diaglayers/basevariant.py
andlaus: While I also don't like the copy and paste across diag layers, given that this is not too much code, the added complexity of the cure would IMO be worse than the disease...

odxtools/diagcomm.py
andlaus: I just tried this: This is pretty difficult to do as part of a list comprehension and it would not come with a major benefit because `resolve_snref(..., Protocol)`  returns a protocol object from the type checking perspective.

odxtools/diaglayers/ecuvariant.py
andlaus: for code which is currently in the main branch `odxtools`, but I have some private scripts which fumble with the internals of ODX databases, and they would become even hackier than they currently are anyway...

odxtools/parentref.py
andlaus: AFAICS this would not work very well anyway, because the attributes of the specialized diag layers have changed. I think we should just make the next release 8.0. (I suppose the number of users which would be saved from trouble due to this is very small.)

odxtools/diaglayers/ecuvariant.py
andlaus: (note that base variants contain "BASE-VARIANT-PATTERNS" instead of "ECU-VARIANT-PATTERNS".)

odxtools/parentref.py
andlaus: okay: I just released v7.4.0 to have a release which contains the fixes which have gone into `main` over the last few weeks but without the potentially disruptive diaglayer changes (which will only go into v8.0)

andlaus:
@kayoub5: do you consider this ready to be merged? (after merging this, I'll do an v8.0 release.)

kayoub5:
> @kayoub5: if no new issues crop up until then, I'll merge this PR on Thursday or Friday...

I am on the beach, with little access to the pc, I believe I already reviewed most of this PR.

If I find issues later, I will create hotfix PRs.

andlaus:
great, thanks. Let's get this merged. (have a nice vacation :))

PR #327, opened 2024-08-02
Diag layer refactoring spinoffs

andlaus:
This PR contains most (i.e., all I found) of the changes of #326 which are not or only tangentially related to refactoring the `DiagLayer` class into a class hierarchy corresponding to the ODX specification.


odxtools/comparamspec.py
```diff
@@ -51,8 +51,7 @@ def from_et(et_element: ElementTree.Element, doc_frags: List[OdxDocFragment]) ->
 
     def _build_odxlinks(self) -> Dict[OdxLinkId, Any]:
         odxlinks: Dict[OdxLinkId, Any] = {}
-        if self.odx_id is not None:
-            odxlinks[self.odx_id] = self
+        odxlinks[self.odx_id] = self
```
kayoub5, line 54: Even in ODX 2.0 ?

odxtools/comparamspec.py
```diff
@@ -26,7 +26,7 @@ class ComparamSpec(IdentifiableElement):
     def from_et(et_element: ElementTree.Element, doc_frags: List[OdxDocFragment]) -> "ComparamSpec":
 
         short_name = odxrequire(et_element.findtext("SHORT-NAME"))
-        doc_frags = [OdxDocFragment(short_name, str(et_element.tag))]
+        doc_frags = [OdxDocFragment(short_name, "COMPARAM-SPEC")]
```
kayoub5, line 29: I think this will break in ODX 2.0

odxtools/comparamspec.py
andlaus: I just checked: in ODX 2.0, COMPARAM-SPEC is derived from ODX-CATEGORY for which the ID attribute is mandatory.

tl;dr: yes

odxtools/comparamspec.py
andlaus: ok, I removed the patch...

odxtools/servicebinner.py
```diff
@@ -91,7 +91,7 @@ def __repr__(self) -> str:
         """
         result = StringIO()
         result.write("[ ")
-        result.write(", ".join([f"0x{x}" for x in self._service_groups if x is not None]))
+        result.write(", ".join([f"0x{x:x}" for x in self._service_groups if x is not None]))
```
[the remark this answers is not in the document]
andlaus: the standard way for such things is to use format specifiers in f-strings: e.g. printing a floating point value with two decimal places accuracy and filled up with zeros would be `[float_val:.03f}`. there are already a few instances of this in the code base, so I think we should use the same mechanism everywhere...

kayoub5 (approved):

PR #328, opened 2024-08-05
SDGs: if the SDG caption is referenced, do not add its IDs to the odxlink database

andlaus:
This prevents these SDG captions to be added multiple times (i.e., for SDGs which only reference the caption but do not define it). Under normal circumstances, this only results in slightly lower performance, but in special situations like having to modify a database, it can lead to SDG captions added which are defined nowhere. (I recently stumbled over such a case.)


andlaus:
I think this is trivial and "objectively correct" enough that there is no need to bother @kayoub5 with a review. @kayoub5: If you would like to have some changes to this, I'll do them in a separate PR...

PR #329, opened 2024-08-05
EnvironmentDataDescription: do not add links to referenced environment data objects

andlaus:
Just like for https://github.com/mercedes-benz/odxtools/pull/328, under normal circumstances, this only results in slightly lower performance, but in special situations like having to modify a database based on another, it can lead to environment data objects being falsely added. (I recently stumbled over such a case.)


andlaus:
just as with #328, I don't think this is worth to get reviewed. @kayoub5: As with #328, if you would like to see some changes , I'll do it in a separate PR. (just leave a comment here...)

PR #330, opened 2024-08-09
Random fixes

andlaus:
This PR contains a fix for the output writing code (the values of limits might be strings with special characters and must thus be escaped) and for the function which summarizes the expected parameters of codec objects (we don't want to see the representation of `CodedConst` objects there).


andlaus:
since this is "obviously correct" (value escaping fix) and at best a question of preference (the CompuConst output improvement), I'll merge this without bothering for a review. @kayoub5: If for some strange reason I'm wrong and you would like to see changes to this, I'm happy to implement them in a separate PR...

PR #334, opened 2024-08-28
odx 2.0 does not use DYN-END-DOP-REF tag

nada-ben-ali:
ODX version 2.0 uses DATA-OBJECT-PROP-REF tag instead of DYN-END-DOP-REF tag.


andlaus:
thanks. with this your mileage varies a bit less for ODX 2.0 datasets :)

PR #335, opened 2024-08-28
odxtypes.parse_int: detect the value base automatically

nada-ben-ali:
The current implementation of **odxtypes.parse_int** function only supports the conversion of base 10 values into integers, not hex or other bases.


kayoub5 (approved):

kayoub5 (approved):

andlaus:
looks good, thanks!

PR #336, opened 2024-08-29
multiplexer.encode_into_pdu: Fix when case_spec is a string and it is the default case

nada-ben-ali:
An EncodeError occurs, showing that the multiplexer does not know the current case when it is the default case.


kayoub5 (approved):

kayoub5 (changes requested):
CI Failed

kayoub5 (approved):

andlaus (approved):
looks good.

PR #340, opened 2024-09-10
Introduce OID attributes

andlaus:
OID stands for "Object ID" and is meant to be a UUID for objects, i.e. it is supposed to be an invariant globally unique identifier for objects. It can e.g., be used to locate objects in an ODX editor across projects, but there is no formal mechanism to reference an object via its OID within the ODX data model.

### DIFF
--- a/odxtools/parameters/parameter.py
+++ b/odxtools/parameters/parameter.py
@@ -40,6 +40,7 @@ class Parameter(NamedElement):
     define any non-positionable parameter types.
 
     """
+    oid: Optional[str]
     byte_position: Optional[int]
     bit_position: Optional[int]
     semantic: Optional[str]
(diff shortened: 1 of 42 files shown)


odxtools/parameters/parameter.py
```diff
@@ -40,6 +40,7 @@ class Parameter(NamedElement):
     define any non-positionable parameter types.
 
     """
+    oid: Optional[str]
```
andlaus, line 43: note that, for whatever reason, most parameters extending `POSITIONABLE-PARAM` do not exhibit an ID but an OID. AFAICS, the only parameters exhibiting IDs are `LENGTH-KEY` and `TABLE-KEY`. (why oh why?)

PR #344, opened 2024-09-18
Truncate the mask when its length is greater than the length of the internal value

nada-ben-ali:
This is to avoid the “OverflowError: int too big to convert” exception when the length of the mask is greater than the length of the internal value.
odxtools version: 8.0.5


kayoub5 (approved):

andlaus:
thanks!

nada-ben-ali:
@andlaus Thank you a lot 🤩

Issue #342, opened 2024-09-15
Is it only for benz?


CCXia1997:
It seems error when use pdx of volkswagen.

kayoub5:
what's the error you are getting?

andlaus:
odxtools is most definitely only for Mercedes cars. That said, I obviously do not have access to diagnostics files made by/for other manufacturers and as @kayoub5 said, we're thus reliant on bug reports/patches from our users. In this case, it is likely that the issue can be fixed if you provided a traceback. also, there is a chance that your files will load if you enable the [non-strict mode](https://github.com/mercedes-benz/odxtools?tab=readme-ov-file#using-the-non-strict-mode) (in this case your mileage will vary, though).

andlaus:
odxtools is most definitely not only for Mercedes cars. That said, I obviously do not have access to diagnostics files made by/for other manufacturers and as @kayoub5 said, we're thus reliant on bug reports/patches from our users. In this case, it is likely that the issue can be fixed if you provided a traceback. also, there is a chance that your files will load if you enable the [non-strict mode](https://github.com/mercedes-benz/odxtools?tab=readme-ov-file#using-the-non-strict-mode) (in this case your mileage will vary, though).

Ch4rg3r:
Is it possible that the specification is different? It seems that odxtools is based on version 2.2 and 2.0.1 for example is jot fully supported?

CCXia1997:
> what's the error you are getting?
Version is 2.0.1 and error when use odxtools.load_odx_d_file() to resolve PROTOCOL file:
**KeyError: "ODXLINK reference OdxLinkRef('ES_XXX') could not be resolved for any of the document fragments [OdxDocFragment(doc_name='ES_XXX', doc_type='LAYER')]"**
ES_XXX is ECU-SHARED-DATA file and PROTOCOL file fragment is:
**<IMPORT-REFS>
     <IMPORT-REF DOCREF="ES_XXX" DOCTYPE="LAYER" ID-REF="ES_XXX" />
</IMPORT-REFS>**

andlaus:
> Is it possible that the specification is different? It seems that odxtools is based on version 2.2 and 2.0.1 for example is not fully supported?

yes, that's possible: we target odx 2.2, and odx 2.0 is only supported on a "best effort" basis, i.e., we happily accept patches provided that they do not break support for odx 2.2, but I personally do not have any datasets which use this version, so your mileage may vary.

[...]

PR #346, opened 2024-09-20
Fix the document fragments for objects not featured by the base diag layer

andlaus:
Since diagnostic layers constitute document fragments in addition to the containers which contain them, the odxlink ids must reflect that. So far, this did not happen for objects which are added by one of the concrete diagnostic layers classes, i.e. objects which were not handled by the diagnostic layer base class.


andlaus:
this might fix #342, but probably it won't. (that said, while doing this PR, I verified that ECU-SHARED-DATA diaglayers can be referenced via their layer document fragment in IMPORT-REFs.)

PR #347, opened 2024-09-23
Implement en- and decoding of SYSTEM parameters

andlaus:
If the value of these parameters are not explicitly specified, their value is determined by looking at the current state of the system the program is running on. (If their type is one of the values specified by the ODX standard.)

Andreas Lauser &lt;andreas.lauser@mercedes-benz.com&gt;, on behalf of [MBition GmbH](https://mbition.io/).
[Provider Information](https://github.com/mercedes-benz/foss/blob/master/PROVIDER_INFORMATION.md)

PR #348, opened 2024-09-24
Improve the `parameter_info()` function

andlaus:
It now knows about system parameters, `SCALE-LINEAR`, `RAT-FUNC` and `SCALE-RAT-FUNC` and `COMPU-CODE` parameters. Also, the code is a bit easier to read and contains a few smaller drive-by fixes.

this is a spin-off of #347 and thus build on top of that PR.

### DIFF
--- a/odxtools/odxtypes.py
+++ b/odxtools/odxtypes.py
@@ -241,3 +241,18 @@ class DataType(Enum):
             return True
         else:
             return False
+
+    def __str__(self) -> str:
+        if self == DataType.A_INT32:
+            return "int"
+        elif self == DataType.A_UINT32:
+            return "uint"
+        elif self in (DataType.A_FLOAT32, DataType.A_FLOAT64):
+            return "float"
+        elif self == DataType.A_BYTEFIELD:
+            return "bytefield"
+        elif self in (DataType.A_UNICODE2STRING, DataType.A_ASCIISTRING, DataType.A_UTF8STRING):
+            return "string"
+        else:
+            odxraise(f"Type info for type {self.value}", NotImplementedError)
+            return "<unknown type>"
--- a/odxtools/odxtypes.py
+++ b/odxtools/odxtypes.py
@@ -254,5 +254,4 @@ class DataType(Enum):
         elif self in (DataType.A_UNICODE2STRING, DataType.A_ASCIISTRING, DataType.A_UTF8STRING):
             return "string"
         else:
-            odxraise(f"Type info for type {self.value}", NotImplementedError)
-            return "<unknown type>"
+            return f"<unknown type '{self.value}'>"
--- a/odxtools/parameterinfo.py
+++ b/odxtools/parameterinfo.py
@@ -3,15 +3,21 @@ import textwrap
 from io import StringIO
 from typing import Iterable
 
+from .compumethods.compucodecompumethod import CompuCodeCompuMethod
 from .compumethods.identicalcompumethod import IdenticalCompuMethod
 from .compumethods.limit import IntervalType
 from .compumethods.linearcompumethod import LinearCompuMethod
+from .compumethods.linearsegment import LinearSegment
+from .compumethods.ratfunccompumethod import RatFuncCompuMethod
+from .compumethods.ratfuncsegment import RatFuncSegment
+from .compumethods.scalelinearcompumethod import ScaleLinearCompuMethod
+from .compumethods.scaleratfunccompumethod import ScaleRatFuncCompuMethod
 from .compumethods.texttablecompumethod import TexttableCompuMethod
 from .dataobjectproperty import DataObjectProperty
 from .dtcdop import DtcDop
 from .dynamiclengthfield import DynamicLengthField
 from .endofpdufield import EndOfPduField
-from .exceptions import odxrequire
+from .exceptions import odxraise, odxrequire
 from .multiplexer import Multiplexer
 from .odxtypes import DataType
 from .parameters.codedconstparameter import CodedConstParameter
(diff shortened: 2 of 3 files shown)


odxtools/parameters/systemparameter.py
```diff
@@ -1,19 +1,25 @@
...
 from ..encodestate import EncodeState
-from ..exceptions import odxrequire
+from ..exceptions import odxraise, odxrequire
 from ..odxlink import OdxDocFragment
 from ..odxtypes import ParameterValue
 from ..utils import dataclass_fields_asdict
 from .parameter import ParameterType
 from .parameterwithdop import ParameterWithDOP
 
+PREDEFINED_SYSPARAM_VALUES = [
```
kayoub5, line 18: Those changes should not be in this PR

andlaus: they are not really: since the changes to `parameter_info()` require the changes of #347, the changes this PR is build on top of #347, i.e., this stuff will go away once #347 has been merged...

odxtools/parameterinfo.py
```diff
@@ -20,12 +26,65 @@
...
+    elif odx_type == DataType.A_BYTEFIELD:
+        return "bytefield"
+    elif odx_type in (DataType.A_UNICODE2STRING, DataType.A_ASCIISTRING, DataType.A_UTF8STRING):
+        return "string"
+    else:
+        odxraise(f"Type info for type {odx_type.value}", NotImplementedError)
+        return "<unknown type>"
+
+
+def _get_linear_segment_info(segment: LinearSegment) -> str:
```
kayoub5, line 52: I have a feeling this should be in the `__str__` or `__repr__` function of LinearSegment

odxtools/parameterinfo.py
```diff
@@ -20,12 +26,65 @@
...
 from .parameters.parameterwithdop import ParameterWithDOP
 from .parameters.reservedparameter import ReservedParameter
+from .parameters.systemparameter import SystemParameter
 from .parameters.tablekeyparameter import TableKeyParameter
 from .parameters.tablestructparameter import TableStructParameter
 from .paramlengthinfotype import ParamLengthInfoType
 from .staticfield import StaticField
 
 
+def _get_type_info(odx_type: DataType) -> str:
```
kayoub5, line 36: I have a feeling this should be in the __str__ or __repr__ function of DataType

andlaus: certainly not `__repr__()`, because if I got this correctly `__repr__()` returns a string containing all the information required to re-create the respective object. for `__str__()`, this is a matter of taste, but I think that for most contexts where a `DataType` object shall be printed, this is a bit too little information?

odxtools/parameterinfo.py
andlaus: I don't think that `LinearSegment` and `RationalSegment` objects should be printed under normal circumstances, and if they are, usually something like the slope and the offset are desired? (IMO this is not the case in the context of `parameter_info()` were we are only interested in what ranges  can be specified  for a given parameter. Opinions welcome...)

odxtools/parameterinfo.py
kayoub5: It's enough for 99% of use, the only thing that's a bit hidden here is the string encoding, and most ODX files use UTF-8 or ASCII, so encoding does not really matter much for them.

andlaus: ok, for `DataType`, I budge ;): https://github.com/mercedes-benz/odxtools/pull/348/commits/6503e1c17fa74667e407144dfead4d6069a05d9e (for the segment objects, I still think it makes little sense but it would constitute a (minor) footgun...)

odxtools/odxtypes.py
```diff
@@ -241,3 +241,18 @@ def isinstance(self, value: Any) -> bool:
...
+        elif self == DataType.A_UINT32:
+            return "uint"
+        elif self in (DataType.A_FLOAT32, DataType.A_FLOAT64):
+            return "float"
+        elif self == DataType.A_BYTEFIELD:
+            return "bytefield"
+        elif self in (DataType.A_UNICODE2STRING, DataType.A_ASCIISTRING, DataType.A_UTF8STRING):
+            return "string"
+        else:
+            odxraise(f"Type info for type {self.value}", NotImplementedError)
```
kayoub5, line 257: Avoid raising in `__str__`, maybe just return `f"<unknown type: {self.value}>"`

odxtools/parameterinfo.py
kayoub5: The idea is that the function that describes the object is human readable way, should be in the object itself, even if it's just for the sake of showing better values in the debugger.

odxtools/odxtypes.py
andlaus: good point. fixed: https://github.com/mercedes-benz/odxtools/pull/348/commits/4a9f43a21510ada21c4b8a3ad55bc268ec871fa9

odxtools/odxtypes.py
```diff
@@ -254,5 +254,4 @@ def __str__(self) -> str:
         elif self in (DataType.A_UNICODE2STRING, DataType.A_ASCIISTRING, DataType.A_UTF8STRING):
             return "string"
         else:
-            odxraise(f"Type info for type {self.value}", NotImplementedError)
-            return "<unknown type>"
+            return "<unknown type '{self.value}'>"
```
kayoub5, line 257:
```suggestion
            return f"<unknown type '{self.value}'>"
```

odxtools/parameterinfo.py
andlaus: yes. my main point is that in most contexts other than `parameter_info()`, this probably requires a bit more information than just the domain where the segment is defined (i.e., some info what the segment does)...

odxtools/odxtypes.py
andlaus: arrg. I'm never too stupid for such mistakes. fixed, thanks!

kayoub5 (approved):

PR #350, opened 2024-10-16
implement libraries

andlaus:
AFAICS, these are pretty much equivalent to `PROG-CODE`.

Besides implementing libraries, this patch also fixes a few minor but important issues with the `PROG-CODE` handling.

(Note that ODX likes to pluralize everything by appending a `s` to the singular, so we get gems like `LIBRARYS` or `FUNCTIONAL-CLASSS` in the standard.)

### DIFF
--- a/odxtools/diaglayers/diaglayerraw.py
+++ b/odxtools/diaglayers/diaglayerraw.py
@@ -48,7 +48,7 @@ class DiagLayerRaw(IdentifiableElement):
     state_charts: NamedItemList[StateChart]
     additional_audiences: NamedItemList[AdditionalAudience]
     # sub_components: List[DiagLayer] # TODO
-    librarys: NamedItemList[Library]  # (sic!)
+    libraries: NamedItemList[Library]
     sdgs: List[SpecialDataGroup]
 
     @property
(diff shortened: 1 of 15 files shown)


odxtools/diaglayers/diaglayerraw.py
```diff
@@ -149,6 +150,10 @@ def from_et(et_element: ElementTree.Element, doc_frags: List[OdxDocFragment]) ->
             for el in et_element.iterfind("ADDITIONAL-AUDIENCES/ADDITIONAL-AUDIENCE")
         ]
 
+        librarys = [
```
kayoub5, line 153: I don't think following the standard mistakes in naming is a good idea

andlaus: agreed. fixed: 5d7cf47d91ff2fce8867cca77aef7f94eb905084

tests/test_singleecujob.py
```diff
@@ -463,6 +464,19 @@ def test_resolve_odxlinks(self) -> None:
...
+            librarys=NamedItemList([
+                Library(
+                    short_name="great_lib",
+                    long_name=None,
+                    description=None,
+                    odx_id=OdxLinkId("my.favourite.lib", doc_frags),
+                    oid=None,
+                    code_file="great_lib.py",
+                    encryption=None,
+                    syntax="PYTHON",
```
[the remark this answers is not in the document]
andlaus: I could not find any in the XSD nor in the PDX of the spec: it is defined as `type="xsd:string"`.  (IMO, encrypting auxiliary files while leaving the diagnostics data itself unencrypted does not make much sense anyway, but whatever...)

kayoub5 (approved):

Issue #354, opened 2024-10-22
infinite loop in DynDefinedSpec._build_odxlinks?


svnmic:
```
 def _build_odxlinks(self) -> Dict[OdxLinkId, Any]:
        ...

        result.update(self._build_odxlinks())
        ...
```

andlaus:
#356 should fix that. Let's close the issue for now...

Issue #355, opened 2024-10-22
oid parameter missing in DtcDop.decode_from_pdu


svnmic:
```
        return DiagnosticTroubleCode(
            trouble_code=trouble_code,
            odx_id=cast(OdxLinkId, None),
            short_name=f'DTC_{trouble_code:06x}',
            long_name=None,
            description=None,
            text=None,
            display_trouble_code=None,
            level=None,
            is_temporary_raw=None,
            sdgs=[],
        )
```

andlaus:
#356 should fix that. Let's close the issue for now...

PR #356, opened 2024-10-28
attempt to fix issue #354 and #355

andlaus:
due to some copy-and-pasto, `DynDefinedSpec._build_odxlinks()` unconditionally called itself and the `oid` parameter was missing from the fallback code for DTC decoding. (Be aware that the latter points towards an incomplete dataset.)


kayoub5 (approved):

andlaus:
okay, I suppose this fixes two clear bugs, so let's merge it even if it might not be sufficient to get around all of @svnmic's issues...

PR #358, opened 2024-10-30
Priority should be given to local refs

nada-ben-ali:
This merge request introduces an "overwrite" parameter to the update method of the OdxLinkDatabase class. This parameter, set to False for imported references, allows local references to take priority over imported ones.

In diaglayer. _resolve_odxlinks, the overwrite=False setting ensures that imported references do not override locally defined references. 

odxtools version: 8.0.5

### DIFF
--- a/odxtools/diaglayers/diaglayer.py
+++ b/odxtools/diaglayers/diaglayer.py
@@ -124,7 +124,7 @@ class DiagLayer:
             # imported references only apply within this specific
             # diagnostic layer
             extended_odxlinks = copy(odxlinks)
-            extended_odxlinks.update(imported_links)
+            extended_odxlinks.update(imported_links, False)
 
             self.diag_layer_raw._resolve_odxlinks(extended_odxlinks)
             return
--- a/odxtools/diaglayers/diaglayer.py
+++ b/odxtools/diaglayers/diaglayer.py
@@ -124,7 +124,7 @@ class DiagLayer:
             # imported references only apply within this specific
             # diagnostic layer
             extended_odxlinks = copy(odxlinks)
-            extended_odxlinks.update(imported_links, False)
+            extended_odxlinks.update(imported_links, overwrite=False)
 
             self.diag_layer_raw._resolve_odxlinks(extended_odxlinks)
             return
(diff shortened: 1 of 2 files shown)


odxtools/diaglayers/diaglayer.py
```diff
@@ -124,7 +124,7 @@ def _resolve_odxlinks(self, odxlinks: OdxLinkDatabase) -> None:
             # imported references only apply within this specific
             # diagnostic layer
             extended_odxlinks = copy(odxlinks)
-            extended_odxlinks.update(imported_links)
+            extended_odxlinks.update(imported_links, False)
```
kayoub5, line 127: @andlaus this is the case that we did manage to catch, not sure if there are other cases that we missed

andlaus (commented):
I like the PR. It would be nice to have that new parameter keyword-only, though because it is very unusual to set it...

odxtools/diaglayers/diaglayer.py
andlaus: as far as I'm aware, IMPORT-REFs should be the only place were this is needed. on the other hand, this could probably also be implemented by reversing the logic, i.e. copying the imported links and overriding them with the local ones? i.e., something like
```python
            extended_odxlinks = copy(imported_links)
            extended_odxlinks.update(odxlinks)
```

(maybe I'm missing something, though.)

andlaus: come think about it, we can probably rename the `imported_links` variable to `extended_links` and replace these two lines by `extended_odxlinks.update(odxlinks)`?

kayoub5: We tried that, `copy(imported_links)` crashed, don't recall the error now

kayoub5, line 127:
```suggestion
            extended_odxlinks.update(imported_links, overwrite=False)
```

andlaus: we can't because `odxlinks` is an `OdxLinkDatabase` object while `imported_links` is a dictionary. (I think the approach outlined by the PR is the best way forward.)

nada-ben-ali: I added it at the same time, thank you @kayoub5 .

kayoub5 (approved):

nada-ben-ali:
@andlaus can I proceed to merge the changes if you approve them? The fix is highly needed.

andlaus:
No need, I'll merge this. thanks for the catch :)

PR #359, opened 2024-11-03
change list command output from tabulate to rich tables

Haris-Ali007:
Issue resolved: I have updated the list bash output by replacing the tabulate table with rich tables. 

File updated: cli/_print_utils.py

The output looks like this:
image

### DIFF
--- a/odxtools/cli/_print_utils.py
+++ b/odxtools/cli/_print_utils.py
@@ -141,12 +141,7 @@ def extract_parameter_tabulation_data(parameters: List[Parameter]) -> Table:
 
     # Create Rich table
     table = Table(
-        title="",
-        show_header=True,
-        header_style="bold cyan",
-        border_style="blue",
-        show_lines=True
-    )
+        title="", show_header=True, header_style="bold cyan", border_style="blue", show_lines=True)
 
     # Add columns with appropriate styling
     table.add_column("Name", style="green")
@@ -234,13 +229,11 @@ def extract_parameter_tabulation_data(parameters: List[Parameter]) -> Table:
             dop.append(None)
 
     # Add all rows at once by zipping dictionary values
-    rows = zip(*[[value if value != None else "" for value in col] 
-                 for col in [name, byte,
-                    bit_length, semantic, param_type,
-                    value, value_type, data_type, dop]])
+    rows = zip(*[[value if value != None else "" for value in col] for col in
+                 [name, byte, bit_length, semantic, param_type, value, value_type, data_type, dop]])
     for row in rows:
         table.add_row(*map(str, row))
-    
+
     return table
 
 
@@ -254,34 +247,24 @@ def print_dl_metrics(variants: List[DiagLayer], print_fn: Callable[..., Any] = p
     """
     # Create Rich table
     table = Table(
-        title="",
-        show_header=True,
-        header_style="bold cyan",
-        border_style="blue",
-        show_lines=True
-    )
-    
+        title="", show_header=True, header_style="bold cyan", border_style="blue", show_lines=True)
+
     # Add columns with appropriate styling
     table.add_column("Name", style="green")
     table.add_column("Variant Type", style="magenta")
     table.add_column("Number of Services", justify="right", style="yellow")
     table.add_column("Number of DOPs", justify="right", style="yellow")
     table.add_column("Number of communication parameters", justify="right", style="yellow")
-    
+
     # Process each variant
     for variant in variants:
         assert isinstance(variant, DiagLayer)
         all_services: List[Union[DiagService, SingleEcuJob]] = sorted(
-            variant.services, key=lambda x: x.short_name
-        )
+            variant.services, key=lambda x: x.short_name)
         ddds = variant.diag_data_dictionary_spec
-        
+
         # Add row to table
-        table.add_row(
-            variant.short_name,
-            variant.variant_type.value,
-            str(len(all_services)),
-            str(len(ddds.data_object_props)),
-            str(len(getattr(variant, "comparams_refs", [])))
-        )
+        table.add_row(variant.short_name, variant.variant_type.value, str(len(all_services)),
+                      str(len(ddds.data_object_props)),
+                      str(len(getattr(variant, "comparams_refs", []))))
     print_fn(table)
--- a/odxtools/cli/_print_utils.py
+++ b/odxtools/cli/_print_utils.py
@@ -155,8 +155,8 @@ def extract_parameter_tabulation_data(parameters: List[Parameter]) -> Table:
     table.add_column("Linked DOP", justify="left", style="white")
 
     name: List[str] = []
-    byte: List[Any] = []
-    bit_length: List[Any] = []
+    byte: List[Optional[int]] = []
+    bit_length: List[Optional[int]] = []
     semantic: List[Optional[str]] = []
     param_type: List[Optional[str]] = []
     value: List[Optional[str]] = []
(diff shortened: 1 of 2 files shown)


CLA-Mercedes-Benz:
[CLA assistant check](https://cla.opensource.mercedes-benz.com/mercedes-benz/odxtools?pullRequest=359) <br/>Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.opensource.mercedes-benz.com/mercedes-benz/odxtools?pullRequest=359) before we can accept your contribution.<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.opensource.mercedes-benz.com/check/mercedes-benz/odxtools?pullRequest=359) it.</sub>

odxtools/cli/_print_utils.py
```diff
@@ -4,7 +4,8 @@
 from typing import Any, Callable, Dict, List, Optional, Union
 
 import markdownify
-from tabulate import tabulate  # TODO: switch to rich tables
```
kayoub5, line 7: tabulate dependency not removed from pyproject.toml

odxtools/cli/_print_utils.py
```diff
@@ -109,8 +107,7 @@ def print_service_parameters(service: DiagService,
         print_fn(f" Negative Response '{resp.short_name}':")
         print_fn(f"   Parameters:\n")
         table = extract_parameter_tabulation_data(list(resp.parameters))
-        table_str = textwrap.indent(tabulate(table, headers='keys', tablefmt='presto'), "    ")
```
kayoub5, line 112: are you sure we don't need the indent anymore?

kayoub5 (changes requested):

odxtools/cli/_print_utils.py
Haris-Ali007: let me update that.

odxtools/cli/_print_utils.py
Haris-Ali007: No you are right. It needs indentation.

Haris-Ali007: I have added indentation to the table.

odxtools/cli/_print_utils.py
Haris-Ali007: I cannot remove tabulate from dependency some other files are using it. This PR was specifically for list command output.

andlaus (commented):
except for the fact that the CI system complains about this, this looks good to me. (replacing the `tabulate` module for `browse` and `compare` would be nice, too...)

andlaus:
@Haris-Ali007: what's the status here? before I can merge this, at least the CI complaints must be fixed...

Haris-Ali007:
> @Haris-Ali007: what's the status here? before I can merge this, at least the CI complaints must be fixed...

Yes apologies for the delay. I am fixing the issues.

odxtools/cli/browse.py
```diff
@@ -5,7 +5,7 @@
 from typing import List, Optional, Union, cast
 
 import InquirerPy.prompt as IP_prompt
-from tabulate import tabulate  # TODO: switch to rich tables
+# from tabulate import tabulate  # TODO: switch to rich tables
```
kayoub5, line 8: commented out code

Haris-Ali007: I removed the tabulate table as the function call returns the rich table. So it can simply print it.

Haris-Ali007: okay

odxtools/cli/_print_utils.py
```diff
@@ -154,15 +154,15 @@ def extract_parameter_tabulation_data(parameters: List[Parameter]) -> Table:
...
-    name = []
-    byte = []
-    bit_length: List[Optional[int]] = []
-    semantic = []
-    param_type = []
-    value: List[Optional[str]] = []
-    value_type: List[Optional[str]] = []
-    data_type: List[Optional[str]] = []
-    dop: List[Optional[str]] = []
+    name: List[Optional[object]] = []
```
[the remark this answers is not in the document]
Haris-Ali007: I defined it as object to solve the cli issue. So you say I keep it simply like name = [ ]?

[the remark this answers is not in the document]
kayoub5: I am just proposing a simpler syntax, yours work if you want to keep it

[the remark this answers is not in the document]
Haris-Ali007: okay sure thank you.

kayoub5 (approved):

kayoub5 (changes requested):
CI failed

odxtools/cli/_print_utils.py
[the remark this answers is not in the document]
andlaus: AFAICS this code is still unchanged? But actually I prefer to use `List[Any]` here, or -- even better -- replacing the `Any` by the types which can actually occur here. Since the latter might be quite a bit of work, `Any` is okay for now...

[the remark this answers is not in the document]
Haris-Ali007: Yeah. you are right. let me fix this.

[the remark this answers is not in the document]
Haris-Ali007: @andlaus I have made changes, I kept the Any for two variables because here I was getting static type error if defined as int, str or None.

odxtools/cli/_print_utils.py
```diff
@@ -149,20 +149,20 @@ def extract_parameter_tabulation_data(parameters: List[Parameter]) -> Table:
...
-    bit_length: List[Optional[object]] = []
-    semantic: List[Optional[object]] = []
-    param_type: List[Optional[object]] = []
-    value: List[Optional[object]] = []
-    value_type: List[Optional[object]] = []
-    data_type: List[Optional[object]] = []
-    dop: List[Optional[object]] = []
+    name: List[str] = []
+    byte: List[Any] = []
+    bit_length: List[Any] = []
```
andlaus, line 159: that should probably be
```suggestion
    byte: List[Optional[int]] = []
    bit_length: List[Optional[int]] = []
```
?

odxtools/cli/_print_utils.py
```diff
@@ -228,9 +228,10 @@ def extract_parameter_tabulation_data(parameters: List[Parameter]) -> Table:
             value_type.append(None)
             dop.append(None)
 
+    for lst in [byte, semantic, bit_length, value, value_type, data_type, dop]:
+        lst[:] = ["" if x is None else x for x in lst]
```
andlaus, line 232: great that you made this code here more readably. line 232 should probably better be
```suggestion
        # replace `None` entries by empty strings
        lst = ["" if x is None else x for x in lst]
```

Haris-Ali007: I have tried using simple `lst` it doesn't override the list in-place.

andlaus: why does it need to be overwritten in-place? (memory wise it should not make any difference because the current solution creates a temporary list as well and also the required memory hopefully is miniscule...)

Haris-Ali007: Yes it does create a temporary list in both case, but in case of `list[:]` it copies the temporary list to the original address. So in a way I am able to edit the list in it's place.

andlaus: okay, I had hands-on look and I now understand why the in-place copy of the list is necessary. That said, declaring `List[Any]` for any of the variables above just causes `mypy` to give up here, so -- while not optimal either -- I prefer to use `List[Optional[int]]` above and add a `# type: ignore[attr-defined, index]` comment here...

Haris-Ali007: okay thanks for the that. I am new to this static type checks. I'll push the changes

PR #360, opened 2024-11-06
Implement LINKED-DTC-DOPs

andlaus:
This turned out to be a bit more involved than anticipated: In contrast to what the term "linked" implies, linked DTC DOPs are not just informative, but an inheritance mechanism for DTCs...

### DIFF
--- a/odxtools/dtcdop.py
+++ b/odxtools/dtcdop.py
@@ -79,6 +79,9 @@ class DtcDop(DopBase):
     linked_dtc_dops_raw: List[LinkedDtcDop]
     is_visible_raw: Optional[bool]
 
+    def __post_init__(self) -> None:
+        self._init_finished = False
+
     @staticmethod
     def from_et(et_element: ElementTree.Element, doc_frags: List[OdxDocFragment]) -> "DtcDop":
         """Reads a DTC-DOP."""
@@ -256,8 +259,10 @@ class DtcDop(DopBase):
         # times. This is required, because the linked DTC DOP feature
         # requires the "parent" DTC DOPs to be fully initialized but
         # the standard does not define a formal ordering of DTC DOPs.
-        if hasattr(self, "_linked_dtc_dops"):
+        if self._init_finished:
             return
+
+        self._init_finished = True
 
         super()._resolve_snrefs(context)
 
@@ -274,10 +279,7 @@ class DtcDop(DopBase):
         # requires that there are no cycles in the "link-hierarchy"
         dtc_short_names = {dtc.short_name for dtc in self._dtcs}
         for linked_dtc_dop in self.linked_dtc_dops_raw:
-            # hack to make sure that the dtcs of the linked DTC DOP
-            # objects have been determined.
-            if not hasattr(linked_dtc_dop.dtc_dop, "_linked_dtc_dops"):
-                linked_dtc_dop.dtc_dop._resolve_snrefs(context)
+            linked_dtc_dop.dtc_dop._resolve_snrefs(context)
 
             for dtc in linked_dtc_dop.dtc_dop.dtcs:
                 if dtc.short_name in dtc_short_names:
--- a/odxtools/dtcdop.py
+++ b/odxtools/dtcdop.py
@@ -15,11 +15,57 @@ from .dopbase import DopBase
 from .encodestate import EncodeState
 from .exceptions import DecodeError, EncodeError, odxassert, odxraise, odxrequire
 from .nameditemlist import NamedItemList
-from .odxlink import OdxDocFragment, OdxLinkDatabase, OdxLinkId, OdxLinkRef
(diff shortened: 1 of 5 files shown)


odxtools/dtcdop.py
```diff
@@ -201,14 +248,52 @@ def _resolve_odxlinks(self, odxlinks: OdxLinkDatabase) -> None:
...
-        self._linked_dtc_dops = NamedItemList(linked_dtc_dops)
+        for linked_dtc_dop in self.linked_dtc_dops_raw:
+            linked_dtc_dop._resolve_odxlinks(odxlinks)
 
     def _resolve_snrefs(self, context: SnRefContext) -> None:
+        # hack to avoid initializing the DtcDop object multiple
+        # times. This is required, because the linked DTC DOP feature
+        # requires the "parent" DTC DOPs to be fully initialized but
+        # the standard does not define a formal ordering of DTC DOPs.
+        if hasattr(self, "_linked_dtc_dops"):
```
kayoub5, line 259: hasattr have performance impact and mypy can't check it, better to use a bool attribute set it to False in init, and True here

odxtools/dtcdop.py
```diff
@@ -201,14 +248,52 @@ def _resolve_odxlinks(self, odxlinks: OdxLinkDatabase) -> None:
...
 
         for dtc_proxy in self.dtcs_raw:
             if isinstance(dtc_proxy, DiagnosticTroubleCode):
                 dtc_proxy._resolve_snrefs(context)
+
+        for linked_dtc_dop in self.linked_dtc_dops_raw:
+            linked_dtc_dop._resolve_snrefs(context)
+
+        # add the inherited DTCs from linked DTC DOPs. Note that this
+        # requires that there are no cycles in the "link-hierarchy"
```
kayoub5, line 274: required by standard, or current logic?

odxtools/dtcdop.py
```diff
@@ -201,14 +248,52 @@ def _resolve_odxlinks(self, odxlinks: OdxLinkDatabase) -> None:
...
+        dtc_short_names = {dtc.short_name for dtc in self._dtcs}
+        for linked_dtc_dop in self.linked_dtc_dops_raw:
+            # hack to make sure that the dtcs of the linked DTC DOP
+            # objects have been determined.
+            if not hasattr(linked_dtc_dop.dtc_dop, "_linked_dtc_dops"):
```
kayoub5, line 279: This check really needed? `_resolve_snrefs` already perform this check

odxtools/dtcdop.py
```diff
@@ -30,7 +76,7 @@ class DtcDop(DopBase):
     physical_type: PhysicalType
     compu_method: CompuMethod
     dtcs_raw: List[Union[DiagnosticTroubleCode, OdxLinkRef]]
-    linked_dtc_dop_refs: List[OdxLinkRef]
+    linked_dtc_dops_raw: List[LinkedDtcDop]
```
kayoub5, line 79: The linked dtcs don't affect serialization logic?

odxtools/dtcdop.py
andlaus: IMO the performance impact is not relevevant here: this function is only called once per `DtcDop` when loading a database, IOW typically a few dozen times. I guess I'll implement the solution which you've proposed tomorrow in order to see if I like it better than the current hack. stay tuned...

odxtools/dtcdop.py
andlaus: If I didn't miss something, the standard does not mention anything about "linked-dop-ref cycles".

That said, although this is not explicitly mentioned in the text, I got the impression that the standard assumes that all linked DTC DOPs are defined in one of the parent diag layers of the layer which contains the DTC-DOP in question...

andlaus: (non-cyclic link graphs are required to get well-defined results, or at least the standard would need to specify how to handle the corner cases that emerge from them.)

odxtools/dtcdop.py
andlaus: they do, that's why `printDOP.xml.jinja2` is modified by this commit...

kayoub5: I am asking if they affect encoding / decoding messages

andlaus: they do not directly affect en-/decoding: just make additional DTCs appear for value parameters that use a DTC-DOP, so additional DTCs can be decoded.

odxtools/dtcdop.py
andlaus: okay, [changed]( https://github.com/mercedes-benz/odxtools/pull/360/commits/bf81d92c8830238ac0cc19d2d3d388a9e0669848): I played around with the idea of a separate `_is_initialized` member and I think that the result is not worse than the `hasattr()` route (IMO it is not better either, but ...)

kayoub5 (approved):

PR #364, opened 2024-11-21
Add codec objects

andlaus:
Codecs are any objects that can be en- and decoded into PDUs. The approach so far was to derive all "composite" codecs from `BasicStructure` and to use a subset of that API for "simple" codable objects like DOPs or parameters. In particular, the previous approach to composite objects was sub-optimal because `BASIC-STRUCTURE` features subtags that other codable objects like `Request` and `Response` do not exhibit (i.e., `.byte_size`).

Since codecs do not play very nicely with inheritance, the new approach is based on typing protocols (Java would call these "interfaces"): `codec.py` defines a few runtime checkable `typing.Protocol` classes which codecs must implement: `Codec` for the basic functionality, `CompositeCodec` defines the API for codecs that are composed of a list of parameters like structures, requests, responses or environment datas, and `ToplevelCodec` defines the "external" API for requests and responses that is supposed to be called by user scripts. Any codable class implementing these APIs can check if it is compliant using `isinstance()` in the `__post_init__()` method of the respective class.


odxtools/basicstructure.py
```diff
@@ -31,6 +26,9 @@ class BasicStructure(ComplexDop):
     parameters: NamedItemList[Parameter]
     byte_size: Optional[int]
 
+    def __post_init__(self) -> None:
+        assert isinstance(self, CompositeCodec)
```
kayoub5, line 30: Why?

tests/test_encoding.py
```diff
@@ -85,7 +85,6 @@ def test_encode_coded_const_infer_order(self) -> None:
             admin_data=None,
             sdgs=[],
             parameters=NamedItemList([param1, param2]),
-            byte_size=None,
```
kayoub5, line 88: Why was `byte_size=None,` removed?

odxtools/response.py
```diff
@@ -23,32 +33,106 @@ class ResponseType(Enum):
...
+        return composite_codec_get_free_parameters(self)
+
+    def print_free_parameters_info(self) -> None:
+        """Return a human readable description of the structure's
+        free parameters.
+        """
+        from .parameterinfo import parameter_info
 
-        super()._resolve_snrefs(context)
+        print(parameter_info(self.free_parameters), end="")
```
kayoub5, line 135: print?

odxtools/response.py
```diff
@@ -23,32 +33,106 @@ class ResponseType(Enum):
 # now, we derive from it anyway because it simplifies the en- and
 # decoding machinery...
 @dataclass
-class Response(BasicStructure):
+class Response(IdentifiableElement):
```
kayoub5, line 36: Response is not a struct?

odxtools/request.py
```diff
@@ -1,39 +1,122 @@
...
+    def _resolve_odxlinks(self, odxlinks: OdxLinkDatabase) -> None:
+        for param in self.parameters:
+            param._resolve_odxlinks(odxlinks)
 
-        return Request(**kwargs)
+    def _resolve_snrefs(self, context: SnRefContext) -> None:
+        context.request = self
+        context.parameters = self.parameters
+
+        try:
```
kayoub5, line 80: Why try? it's not expected to raise

odxtools/paramlengthinfotype.py
```diff
@@ -20,9 +20,16 @@
...
-
     length_key_ref: OdxLinkRef
 
+    @property
+    def dct_type(self) -> DctType:
+        return "PARAM-LENGTH-INFO-TYPE"
+
+    @property
+    def length_key(self) -> "LengthKeyParameter":
+        return self._length_key
```
kayoub5, line 31: PR is already complex as is, those kind of non-functional code changes make the review even harder, please revert and move them to their own PR

odxtools/parameters/parameter.py
```diff
@@ -46,6 +47,9 @@ class Parameter(NamedElement):
     semantic: Optional[str]
     sdgs: List[SpecialDataGroup]
 
+    def __post_init__(self) -> None:
+        assert isinstance(self, Codec)
```
kayoub5, line 51: Why?

odxtools/basicstructure.py
andlaus: to somehow enforce that the class adheres to the CompositeCodec protocol. A better way would be to use inheritance, but whenever I tried I ran into troubles with MRO. Deriving a class from a CompositeCodec (which derives from Codec) and a base class that is a Codec but not a CompositeCodec does not work. This happens e.g., for BasicStructure...

odxtools/response.py
andlaus: yes, IMO this is okay because the name of the calling method is `print_free_parameters_info()`. (it is meant to get overview of the parameters expected by `encode()`/produced by `decode()` during development. IOW it is not meant to be called during "normal" program execution.)

odxtools/response.py
andlaus: not according to the XSD...

odxtools/request.py
andlaus: this is a defensive measure to make it not screw up more than necessary even if we are in non-strict mode and an exception is raised by a parameter's `.resolve_snref()`

kayoub5: code comment to explain would be nice then

odxtools/basicstructure.py
kayoub5: such checks should be unit tests, not runtime checks.

kayoub5: isn't mypy doing the type checking already anyway?

tests/test_encoding.py
kayoub5: possible to move this change to its own PR? this particular change is too noisy in this PR

andlaus: that's probably possible, but as I see it, most of the point of this PR is not having to derive `Request` and `Response` from `BasicStructure` anymore...

odxtools/basicstructure.py
andlaus: good point. fixed: https://github.com/mercedes-benz/odxtools/pull/364/commits/1ebedcea1ef6b3e28a512aa602c0d2478ff542a0

odxtools/codec.py
```diff
@@ -0,0 +1,227 @@
...
+
+
+def composite_codec_get_free_parameters(codec: CompositeCodec) -> List["Parameter"]:
+    """Return the list of parameters which can be freely specified by
+    the user when encoding the structure
+
+    This means all required parameters plus parameters that can be
+    omitted because they specify a default.
+    """
+    return [p for p in codec.parameters if p.is_settable]
```
kayoub5, line 99: not much of a helper method if it's only one line

As far as I know, it's not even needed by internal API

odxtools/codec.py
```diff
@@ -0,0 +1,227 @@
...
+        result[param.short_name] = value
+
+    # decoding of the structure finished. go back the original origin.
+    decode_state.origin_byte_position = orig_origin
+
+    return result
+
+
+@runtime_checkable
+class ToplevelCodec(typing.Protocol):
```
kayoub5, line 216: as far as code goes, this interface have no practical use case in this PR, it's not used except for type checking that could be removed

odxtools/dataobjectproperty.py
```diff
@@ -108,6 +108,10 @@ def _resolve_snrefs(self, context: SnRefContext) -> None:
     def unit(self) -> Optional[Unit]:
         return self._unit
 
+    @property
+    def byte_size(self) -> Optional[int]:
+        return self.diag_coded_type.byte_size
```
kayoub5, line 113: I don't think adding byte_size actually needed

odxtools/request.py
```diff
@@ -1,39 +1,122 @@
...
+            for param in self.parameters:
+                param._resolve_snrefs(context)
+        finally:
+            context.request = None
+            context.parameters = None
+
+    def get_static_bit_length(self) -> Optional[int]:
+        return composite_codec_get_static_bit_length(self)
+
+    def print_free_parameters_info(self) -> None:
```
kayoub5, line 90: personally I think that no object should have print function, that limits the re-usability of the function

examples/somersaultecu.py
```diff
@@ -1008,10 +1003,45 @@ class SomersaultSID(IntEnum):
                     sdgs=[],
                 ),
             ]),
-            byte_size=None,
         ),
 }
 
+somersault_structures = {
```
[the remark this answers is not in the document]
andlaus: this is because requests and responses are no longer basic structures, so they cannot be referenced by STRUCT-REF anymore. (IOW, the somersault PDX file that is currently shipped with `main` is invalid...)

odxtools/request.py
andlaus: come think about it, you're right: this code can be removed. fixed: https://github.com/mercedes-benz/odxtools/pull/364/commits/e5d21e67e16a72abf25a16b3aa2ed07b868e15e6

odxtools/parameters/parameter.py
andlaus: not much of a point. removed, thanks!

odxtools/codec.py
andlaus: yeah, it is used for the `print_parameters` stuff of composite codecs. I'd like to keep it, because it avoids discrepancies between `Request`, `Response` and `BasicStructure` if this gets copy-and pasted...

odxtools/request.py
andlaus: I guess I disagree slightly here: When exploring an ODX database using the REPL (which I frequently do), I find it quite a bit more convenient to be able to do something like
```
service.print_f<TAB><RETURN>
```
than
```
print(service.print_f<TAB>)<RETURN>
```
(that's basically the use case for this method...)

odxtools/paramlengthinfotype.py
andlaus: I agree, and I should have reverted this before opening the PR. That said, moving this into a separate PR now that it is already reviewed would increase your workload instead of reducing it, because github would show this as clutter in the "diff since your last review" view?

kayoub5: you have the habit of using force-push, that means I have to do full review anyway.

odxtools/request.py
kayoub5: you don't need to use `print` if the function return a string, REPL will auto print the return value of the function

kayoub5:
@andlaus you have a merge conflict

andlaus:
> @andlaus you have a merge conflict

right, fixed (sorry)

odxtools/request.py
andlaus: kind of: IMO the readability of strings with new line characters is quite bad of it is not explictly printed...

odxtools/paramlengthinfotype.py
andlaus: sorry, I was not aware that github does not take kindly to that. if you prefer, I can push the fixups as separate patches and only squash them before merge...

kayoub5: please do that from now on

odxtools/cli/browse.py
```diff
@@ -192,56 +207,58 @@ def encode_message_from_string_values(
...
-        parameter = param_dict.get(parameter_sn)
+        parameter = resolve_snref(parameter_sn, sub_service.parameters, Parameter)
         if parameter is None:
             print(f"I don't know the parameter {parameter_sn}")
             continue
 
         if isinstance(parameter_value, dict):
             # parameter_value refers to a structure (represented as dict of params)
+            dop = getattr(parameter, "dop", None)
+            inner_params = getattr(dop, "parameters", None)
```
kayoub5, line 243: not using the interface you created?

odxtools/codec.py
```diff
@@ -0,0 +1,212 @@
...
+@runtime_checkable
+class CompositeCodec(Codec, typing.Protocol):
+    """Any object which can be en- or decoded to be transferred over
+    the wire which is composed of multiple parameter implements this
+    API.
+
+    """
+
+    @property
+    def parameters(self) -> List["Parameter"]:
```
kayoub5, line 44: I am not 100% sure, but I think this is more correct 
```suggestion
    def parameters(self) -> NamedItemList[Parameter]:
```

odxtools/codec.py
```diff
@@ -0,0 +1,212 @@
...
+            return None
+        elif param.byte_position is not None:
+            cursor = param.byte_position
+
+        cursor += ((param.bit_position or 0) + param_bit_length + 7) // 8
+        byte_length = max(byte_length, cursor)
+
+    # Round up to account for padding bits (all structures are
+    # byte aligned)
+    return byte_length * 8
```
kayoub5, line 80: where is the round up code?

odxtools/codec.py
```diff
@@ -0,0 +1,212 @@
...
+    omitted because they specify a default.
+    """
+    return [p for p in codec.parameters if p.is_settable]
+
+
+def composite_codec_get_coded_const_prefix(codec: CompositeCodec,
+                                           request_prefix: bytes = b'') -> bytes:
+    from .parameters.codedconstparameter import CodedConstParameter
+    from .parameters.matchingrequestparameter import MatchingRequestParameter
+    from .parameters.physicalconstantparameter import PhysicalConstantParameter
```
kayoub5, line 106: why is the import here?

odxtools/codec.py
```diff
@@ -0,0 +1,212 @@
...
+    orig_is_end_of_pdu = encode_state.is_end_of_pdu
+    encode_state.is_end_of_pdu = False
+
+    # ensure that no values for unknown parameters are specified.
+    if not encode_state.allow_unknown_parameters:
+        param_names = {param.short_name for param in codec.parameters}
+        for param_value_name in physical_value:
+            if param_value_name not in param_names:
+                odxraise(f"Value for unknown parameter '{param_value_name}' specified "
+                         f"for structure {codec.short_name}")
```
kayoub5, line 148: word "structure" don't fit here anymore

odxtools/codec.py
```diff
@@ -0,0 +1,212 @@
...
+
+    for param in codec.parameters:
+        if id(param) == id(codec.parameters[-1]):
+            # The last parameter of the structure is at the end of
+            # the PDU if the structure itcodec is at the end of the
```
kayoub5, line 153: grammatically the sentence makes no sense

odxtools/codec.py
```diff
@@ -0,0 +1,212 @@
...
+            # PDU. TODO: This assumes that the last parameter
```
kayoub5, line 154: don't start a todo in the middle of a comment

odxtools/diagcodedtype.py
```diff
@@ -55,6 +55,10 @@ def _resolve_snrefs(self, context: SnRefContext) -> None:  # noqa: B027
         """Recursively resolve any short-name references"""
         pass
 
+    @property
+    def byte_size(self) -> Optional[int]:
+        return None
```
kayoub5, line 60: really needed?

odxtools/message.py
```diff
@@ -29,8 +27,3 @@ class Message:
 
     def __getitem__(self, key: str) -> ParameterValue:
         return self.param_dict[key]
-
-    @property
-    @deprecated("use .coding_object")  # type: ignore[misc]
-    def structure(self) -> Union["Request", "Response"]:
```
kayoub5, line 35: breaking change, merit its own PR

odxtools/request.py
```diff
@@ -1,39 +1,114 @@
...
 
-        return encode_state.coded_message
+        for param in self.parameters:
+            result.update(param._build_odxlinks())
+
+        return result
+
+    def _resolve_odxlinks(self, odxlinks: OdxLinkDatabase) -> None:
+        for param in self.parameters:
+            param._resolve_odxlinks(odxlinks)
```
kayoub5, line 67: what about the other properties?

odxtools/cli/browse.py
```diff
@@ -192,56 +207,58 @@ def encode_message_from_string_values(
...
-            if param.is_required and parameter_values.get(param_sn) is None:
-                missing_parameter_names.append(param_sn)
+            if param.is_required and parameter_values.get(param.short_name) is None:
+                missing_parameter_names.append(param.short_name)
 
     if len(missing_parameter_names) > 0:
-        print("The following parameters are required but missing!")
-        print(" - " + "\n - ".join(missing_parameter_names))
+        print("The following parameters are required but missing:")
+        print(" - " + "\n - ".join(sorted(missing_parameter_names)))
```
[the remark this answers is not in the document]
andlaus: Not 100% sure. I guess to make it easier to read.

odxtools/cli/browse.py
andlaus: not every parameter exhibits a DOP (e.g. `ReservedParameter`) and not every DOP has parameters (only `ComplexDop`s like structures and environment datas)...

odxtools/codec.py
andlaus: I think it should stay at `List`, because `NamedItemList` is a `List` and we should probably not rely on any `NamedItemList` specifics when we assume a `Codec` object...

odxtools/codec.py
andlaus: not needed: a byte is always 8 bits ;)

andlaus: (the comment is a bit misleading, so I removed it.)

odxtools/codec.py
andlaus: because it is needed to compute the static bit size of composite codec objects...

odxtools/diagcodedtype.py
andlaus: nope. removed, thanks.

odxtools/codec.py
kayoub5: ok, remove the extra quotes then

odxtools/request.py
kayoub5: I am asking why not call _resolve_odxlinks for the admin data and sdgs?

odxtools/cli/browse.py
andlaus: is `isintance(dop, CompositeCodec)` more performant than `getattr`? (I assume that it is significantly slower)

odxtools/codec.py
andlaus: Not sure if this is a good idea: `Parameter` is a `CompositeCodec`, so whilst this would work in the present version of the PR (but not in the original one!), importing `Parameter` in `codec.py`  would IMO be something like a "cyclic import by spirit". (this sounds quite esoteric :D. in any case, mypy works as expected because of the conditional import of `Parameter` at the beginning of the file.)

odxtools/codec.py
andlaus: because in the original version of the PR it would have lead to cyclic imports. That said, if we move these imports to the top we can also import `Parameter` (i.e. forget about my comment above.)

odxtools/codec.py
andlaus: ok, changed my mind. (see below). probably I should factor out the helper functions from `codec.py` into `codec_helpers.py`, but that's work for another day...

odxtools/codec.py
andlaus: fixed.

odxtools/request.py
andlaus: ops. right, fixed. Thanks!

odxtools/cli/browse.py
kayoub5: when it comes to performance, the only way to know for sure is to measure

odxtools/message.py
kayoub5: topic still open

andlaus: removal reversed...

odxtools/cli/browse.py
andlaus: if I did not make a mistake, `getattr` is faster by 5 orders of magnitude:

```
> cat bla.py
import odxtools
import timeit

from typing import Optional, List
from odxtools.codec import CompositeCodec
from odxtools.odxtypes import ParameterValue
from odxtools.decodestate import DecodeState
from odxtools.encodestate import EncodeState
from odxtools.parameters.parameter import Parameter
from odxtools.codec import Codec
from odxtools.codec import CompositeCodec

class NotACodec:
    def foo(self):
        pass

class ABasicCodec(NotACodec):
    @property
    def short_name(self) -> str:
        return "foo"

    def encode_into_pdu(self, physical_value: Optional[ParameterValue],
                        encode_state: EncodeState) -> None:
        pass

    def decode_from_pdu(self, decode_state: DecodeState) -> ParameterValue:
        ...

    def get_static_bit_length(self) -> Optional[int]:
        ...

class ACompositeCodec(ABasicCodec):
    @property
    def parameters(self) -> List[Parameter]:
        return []

    @property
    def required_parameters(self) -> List[Parameter]:
        return []

    @property
    def free_parameters(self) -> List[Parameter]:
        return []


a = NotACodec()
b = ABasicCodec()
c = ACompositeCodec()

n = 100_000

def foo():
    for i in range(n):
        isinstance(c, CompositeCodec)

def bar():
    for i in range(n):
        getattr(c, 'parameters')

t1 = timeit.timeit("foo()", number=1, globals=globals())
print(f"isinstance: {t1}")
t2 = timeit.timeit("bar()", number=1, globals=globals())
print(f"getattr: {t2}")
> python3 bla.py 
isinstance: 1.728830883000228
getattr: 0.008215672000005725
```

kayoub5 (approved):

PR #371, opened 2024-12-09
Add Pickle support

nada-ben-ali:
I am working on caching the PDX config parsing result to avoid reparsing it each time, using Pickle for serialization.

During the pickling process, no issues were encountered. However, during the unpickling step, the following exception occurred: "maximum recursion depth exceeded" which originated in the NamedItemList class. 
The recursion occurred due to infinite calls to the NamedItemList._getattr_ method. Specifically, the code attempted to access a property that was uninitialized at the time, causing the _getattr_ to be invoked repeatedly during the reconstruction of the object.
To address this, I added the __reduce__ method which explicitly defines how an object is serialized and reconstructed during pickling and unpickling. After adding this method, the recursion issue was resolved.

After fixing the recursion issue, I got the following exception with some PDX files:

image

This issue arises because the NamedItemList assumes all its elements conform to the OdxNamed interface, including having a short_name attribute. While the PDX files were correctly parsed initially, the issue arose because TableRow objects were reconstructed in a state where their short_name attribute was not properly restored or initialized.
Thus, the solution was to add the __reduce__ method for the TableRow class, ensuring its attributes are correctly restored during unpickling.

odxtools: 8.3.4


andlaus:
I would not recommend pickling or deepcopying large ODX databases directly, because the heavy referencing of ODX screws with the inner workings of these mechanisms, so they often do not work (because of e.g., cyclic references, very deep nesting depths, etc) or -- if they do -- they are often slower than loading the PDX file from scratch. What is beneficial is to only copy the dataclass fields of objects and pickle/unpickle those:

```python
def deepcopy_dataclass(obj: Any, memo: dict[int, Any]) -> Any:
    obj_id = id(obj)
    if (obj_copy := memo.get(obj_id)) is not None:
        return obj_copy

    if isinstance(obj, list):
        result: list[Any] = []
        memo[obj_id] = result
        for x in obj:
            result.append(deepcopy_dataclass(x, memo))
        return result
    elif obj is None:
        return None
    elif isinstance(obj, int | float | str | bytes | bytearray):
        tmp = obj
        memo[obj_id] = tmp
        return tmp
    elif dataclasses.is_dataclass(obj):
        tmp2 = type(obj).__new__(type(obj))  # type: ignore
        memo[obj_id] = tmp2

        kwargs = {}
        for field in dataclasses.fields(obj):
            kwargs[field.name] = deepcopy_dataclass(getattr(obj, field.name), memo)

        tmp2.__init__(**kwargs)  # type: ignore

        memo[obj_id] = tmp2
        return tmp2

    else:
        tmp3 = copy(obj)
        memo[obj_id] = tmp3

        return tmp3

def copy_db(db: Database) -> Database:
    new_db = copy(db)
    new_db._diag_layer_containers = deepcopy_dataclass(db._diag_layer_containers,
                                                            {})
    new_db._comparam_subsets = deepcopy_dataclass(db._comparam_subsets, {})
    new_db._comparam_specs = deepcopy_dataclass(db._comparam_specs, {})

    return new_db   
```

the relevant parts of an existing ODX database can then be duplicated using `db2 = copy_db(db)` and the `db2` object should play nicely with `pickle()`, `deepcopy()`, etc. (before you use it in "normal" code, be sure to call `db2.refresh()` to resolve the references, deal with inheritance, etc.)

nada-ben-ali:
@andlaus, I tried your suggestion and added a comment with the results I got in the new [PR](https://github.com/mercedes-benz/odxtools/pull/372). (This PR is just a mess, so I closed it, sorry :) )

PR #372, opened 2024-12-09
Add Pickle support

nada-ben-ali:
I am working on caching the PDX config parsing result to avoid reparsing it each time, using Pickle for serialization.

During the pickling process, no issues were encountered. However, during the unpickling step, the following exception occurred: "maximum recursion depth exceeded" which originated in the NamedItemList class.
The recursion occurred due to infinite calls to the NamedItemList.\_\_getattr__ method. Specifically, the code attempted to access a property that was uninitialized at the time, causing the \_\_getattr__ to be invoked repeatedly during the reconstruction of the object.
To address this, I added the \_\_reduce__ method which explicitly defines how an object is serialized and reconstructed during pickling and unpickling. After adding this method, the recursion issue was resolved.

After fixing the recursion issue, I got the following exception with some PDX files during the unpickling:

image

This issue arises because the NamedItemList assumes all its elements conform to the OdxNamed interface, including having a short_name attribute. While the PDX files were correctly parsed initially, the issue arose because TableRow objects were reconstructed in a state where their short_name attribute was not properly restored or initialized.
Thus, the solution was to add the \_\_reduce__ method for the TableRow class, ensuring its attributes are correctly restored during unpickling.

odxtools: 8.3.4

### DIFF
--- a/odxtools/nameditemlist.py
+++ b/odxtools/nameditemlist.py
@@ -189,6 +189,12 @@ class ItemAttributeList(List[T]):
 
         return result
 
+    def __reduce__(self):
+        reconstruct_function = self.__class__
+        args = (list(self),)
+        state = {"_item_dict": self._item_dict}
+        return reconstruct_function, args, state
+
 
 class NamedItemList(ItemAttributeList[T]):
 
--- a/odxtools/nameditemlist.py
+++ b/odxtools/nameditemlist.py
@@ -189,7 +189,7 @@ class ItemAttributeList(List[T]):
 
         return result
 
-    def __reduce__(self):
+    def __reduce__(self) -> Tuple[Any, ...]:
         reconstruct_function = self.__class__
         args = (list(self),)
         state = {"_item_dict": self._item_dict}
--- a/odxtools/nameditemlist.py
+++ b/odxtools/nameditemlist.py
@@ -190,10 +190,14 @@ class ItemAttributeList(List[T]):
         return result
 
     def __reduce__(self) -> Tuple[Any, ...]:
-        reconstruct_function = self.__class__
-        args = (list(self),)
-        state = {"_item_dict": self._item_dict}
-        return reconstruct_function, args, state
+        """Support for Python's pickle protocol.
+        This method ensures that the object can be reconstructed with its current state,
+        using its class and the list of items it contains.
+        It returns a tuple containing the reconstruction function (the class)
+        and its arguments necessary to recreate the object.
+
+        """
+        return self.__class__, (list(self),)
 
 
 class NamedItemList(ItemAttributeList[T]):
--- a/odxtools/tablerow.py
+++ b/odxtools/tablerow.py
@@ -157,3 +157,10 @@ class TableRow(IdentifiableElement):
     def dop(self) -> Optional[DataObjectProperty]:
         """The data object property object resolved by dop_ref."""
         return self._dop
+
+    def __reduce__(self):
+        state = self.__dict__.copy()
+        return (self.__class__, (self.short_name, self.long_name, self.description, self.odx_id,
+                                 self.oid, self.table_ref, self.key_raw, self.structure_ref,
+                                 self.structure_snref, self.dop_ref, self.dop_snref, self.semantic,
+                                 self.sdgs), state)
--- a/odxtools/tablerow.py
+++ b/odxtools/tablerow.py
@@ -1,5 +1,5 @@
 # SPDX-License-Identifier: MIT
-from dataclasses import dataclass
+from dataclasses import dataclass, fields
 from typing import TYPE_CHECKING, Any, Dict, List, Optional, Tuple
 from xml.etree import ElementTree
 
@@ -159,8 +159,6 @@ class TableRow(IdentifiableElement):
         return self._dop
 
     def __reduce__(self) -> Tuple[Any, ...]:
+        """This ensures that the object can be correctly reconstructed during unpickling."""
         state = self.__dict__.copy()
-        return (self.__class__,
-                (self.short_name, self.long_name, self.description, self.odx_id, self.oid,
-                 self.table_ref, self.key_raw, self.structure_ref, self.structure_snref,
-                 self.dop_ref, self.dop_snref, self.semantic, self.sdgs), state)
+        return self.__class__, tuple([getattr(self, x.name) for x in fields(self)]), state
--- a/odxtools/tablerow.py
+++ b/odxtools/tablerow.py
@@ -1,6 +1,6 @@
 # SPDX-License-Identifier: MIT
 from dataclasses import dataclass
-from typing import TYPE_CHECKING, Any, Dict, List, Optional
+from typing import TYPE_CHECKING, Any, Dict, List, Optional, Tuple
 from xml.etree import ElementTree
 
 from .basicstructure import BasicStructure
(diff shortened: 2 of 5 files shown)


nada-ben-ali:
@andlaus I tried your [suggestion](https://github.com/mercedes-benz/odxtools/pull/371#issuecomment-2528273680), which is to copy only the dataclass fields of the objects and then pickle/unpickle the object returned by “copy_db”.
I got an exception saying that “cannot pickle ‘zlib.Decompress’ object”:

image

I therefore excluded .jar files from serialization by adding \_\_getstate__ and \_\_setstate__ methods to the Database class.
Then, I got the same exception regarding the recursion:

image

nada-ben-ali:
@kayoub5 what do you think? Any suggestion?

kayoub5:
> @kayoub5 what do you think? Any suggestion?

The approach taken by the PR looks valid at first glance.

A unit test will be needed to ensure future versions don't break pickling support.

For the suggestion from @andlaus, pickling should be in theory a faster than re-reading the pdx, and work fine on circular references of normal classes.

* Trying to get the whole database to work with pickling is not the target here.
* We are not pickling the whole pdx, just the request/response we are interested in (so unused inherited dops and structures are pruned from serialization) 
* the NamedItemList overriding the builtin list type is causing the side effects on pickling we are trying to fix here.


@nada-ben-ali do you have numbers on how much pickling is faster on a large PDX?

andlaus:
> (so unused inherited dops and structures are pruned from serialization)

I guess it's better to overload the `__getstate__()` and `__setstate__()` methods in classes that should be pickleable: https://docs.python.org/3/library/pickle.html#object.__getstate__

(as a general approach, a `OdxDataObject` class could be introduced which could handle serialization of dataclass fields. then, it must just be ensured that all dataclasses that correspond to ODX XML tags are derived from that class.)

kayoub5:
> > (so unused inherited dops and structures are pruned from serialization)
> 
> I guess it's better to overload the `__getstate__()` and `__setstate__()` methods in classes that should be pickleable: https://docs.python.org/3/library/pickle.html#object.__getstate__
> 
> (as a general approach, a `OdxDataObject` class could be introduced which could handle serialization of dataclass fields. then, it must just be ensured that all dataclasses that correspond to ODX XML tags are derived from that class.)

we tried that, doesn't work on a subclass of the builtin type list.

dataclasses are in general pickable, tablerow was the exception that we had to fix.

andlaus:
> dataclasses are in general pickable

sure, but you probably only want to serialize the fields of the dataclass, not also the resolved references, etc?!

kayoub5:
> > dataclasses are in general pickable
> 
> sure, but you probably only want to serialize the fields of the dataclass, not also the resolved references, etc?!

actually, we do, saves time on computing them again

nada-ben-ali:
> @nada-ben-ali do you have numbers on how much pickling is faster on a large PDX?

@kayoub5 yes, pickling has saved us a lot of time. For example, parsing a PDX file without pickling took **14 sec 296 ms**, while using the pickling reduced it to just **846 ms**. Similarly, with another PDX file, the time dropped from **2 sec 791 ms** to **406 ms**.

>I guess it's better to overload the __getstate__() and __setstate__() methods in classes that should be pickleable: https://docs.python.org/3/library/pickle.html#object.__getstate__

@andlaus we tried that, but for list subclasses, the \_\_setstate__() method is not invoked.

odxtools/nameditemlist.py
```diff
@@ -189,6 +189,12 @@ def __deepcopy__(self, memo: Dict[int, Any]) -> Any:
 
         return result
 
+    def __reduce__(self) -> Tuple[Any, ...]:
```
andlaus, line 192: some docstring explaining the purpose of this method would be great

odxtools/nameditemlist.py
```diff
@@ -189,6 +189,12 @@ def __deepcopy__(self, memo: Dict[int, Any]) -> Any:
...
+        reconstruct_function = self.__class__
+        args = (list(self),)
+        state = {"_item_dict": self._item_dict}
```
andlaus, line 195: Is `._item_dict` really required to restore a `ItemAttributeList` object?

odxtools/tablerow.py
```diff
@@ -157,3 +157,10 @@ def structure(self) -> Optional[BasicStructure]:
     def dop(self) -> Optional[DataObjectProperty]:
         """The data object property object resolved by dop_ref."""
         return self._dop
+
+    def __reduce__(self) -> Tuple[Any, ...]:
+        state = self.__dict__.copy()
+        return (self.__class__,
+                (self.short_name, self.long_name, self.description, self.odx_id, self.oid,
+                 self.table_ref, self.key_raw, self.structure_ref, self.structure_snref,
+                 self.dop_ref, self.dop_snref, self.semantic, self.sdgs), state)
```
andlaus, line 166: this seem to be the dataclass fields? if yes, this can written down as
```python
[self.__class__] + [ getattr(self, x.name for x in dataclasses.fields(self) ]
```

another question that I have if this will still work once we switch to python 3.10 and add `kw_only=True` to the `@dataclass` decorators...

kayoub5: for python 3.10, we will figure out that bridge once we cross it

odxtools/nameditemlist.py
nada-ben-ali: No, it's not required, because during the unpickling, the ItemAttributeList constructor is invoked, which initializes the list and populates _item_dict via the "append" calls.

tests/test_diag_data_dictionary_spec.py
```diff
@@ -428,6 +429,11 @@ def test_initialization(self) -> None:
         decoded = mux.decode_from_pdu(decode_state)
         self.assertEqual(decoded, ("default_case", {}))
 
+        # check that the object can be properly serialized and deserialized using pickle
```
kayoub5, line 432: move pickle asserts to their own test function (as much as possible)

tests/test_odxtools.py
```diff
@@ -90,16 +91,28 @@ def test_encode_response_with_matching_request_param_and_structure(self) -> None
...
+
+
 class TestNamedItemList(unittest.TestCase):
 
+    def setUp(self) -> None:
+        self.foo_obj = NamedItemList([X("hello", 0), X("world", 1)])
+
     def test_NamedItemList(self) -> None:
+        # Test with the original object
+        foo = self.foo_obj.__class__(self.foo_obj)
```
kayoub5, line 107: why are you doing this?

nada-ben-ali: I updated the existing NamedItemList unit test to run on both the original object and the pickled one to ensure that the object is correctly reconstructed. I agree that copying the object is unnecessary, so I'll update it.

kayoub5 (approved):

nada-ben-ali:
@andlaus could you please review the changes after the updates I've made and if you approve them, then we can merge them?

odxtools/tablerow.py
```diff
@@ -157,3 +157,8 @@ def structure(self) -> Optional[BasicStructure]:
     def dop(self) -> Optional[DataObjectProperty]:
         """The data object property object resolved by dop_ref."""
         return self._dop
+
+    def __reduce__(self) -> Tuple[Any, ...]:
+        """This ensures that the object can be correctly reconstructed during unpickling."""
+        state = self.__dict__.copy()
+        return self.__class__, tuple([getattr(self, x.name) for x in fields(self)]), state
```
andlaus, line 164: is `state` required here?

andlaus (approved):
@nada-ben-ali: press the green button once you're happy with it...

odxtools/tablerow.py
kayoub5: @andlaus I assume for the properties computed through odxlinks

nada-ben-ali: @andlaus yes, it's necessary to capture the non-dataclass attributes, such us \_structure, \_key, ...So without including the state, these attributes would not be restored.

andlaus: okay, but this makes serialization of individual objects (i.e., not the whole database) quite a bit larger than strictly necessary because a lot of tangentially related objects (i.e., all referenced objects) will also be serialized? (to be clear: I'm fine with that.)

nada-ben-ali: yes you're right, but I think it's better to include the referenced objects in the serialization rather than recomputing them after deserialization.

Issue #368, opened 2024-12-02
Wrong parameter handling with multiple odx files


mafei1982:
I have a pdx file which contains multiple odx files. One of them define a ecu shared data which contains some diag services, and one of them define a basevariant which only contains some definitions of tables/table rows.  in ecu shared data file, there is a param definition like below:
```
<PARAM ID="DLC.ES_UDS_Services.ESD.ES_UDS_Services.DOP.STRUCT_Service86_03_ServiceToRespondToRecord.PA.ServiceToRespondToDID"
                  xsi:type="TABLE-KEY">
                  <SHORT-NAME>ServiceToRespondToDID</SHORT-NAME>
                  <LONG-NAME>ServiceToRespondToDID</LONG-NAME>
                  <TABLE-SNREF SHORT-NAME="TABLE_Service22_DataID"/>
</PARAM>
```
and in this file, there is a dummy table which short name is "TABLE_Service22_DataID":
```
<TABLE ID="id-59bd0afb-b95a-4e18-acca-162f2ea460cb" OID="59bd0afb-b95a-4e18-acca-162f2ea460cb" SEMANTIC="DATA-READ">
              <SHORT-NAME>TABLE_Service22_DataID</SHORT-NAME>
              <LONG-NAME>TABLE Service22 DataID</LONG-NAME>
              <KEY-DOP-REF ID-REF="id-206979a7-32f8-412c-b2f8-93f8bfb74db6"/>
              <TABLE-ROW ID="id-631ab6af-fd93-43d1-911b-a3b02d50db7e" OID="13fe66c9-9f6a-4f78-8bde-fba011a51cb7">
                <SHORT-NAME>Dummy_Row</SHORT-NAME>
                <LONG-NAME>Dummy Row</LONG-NAME>
                <KEY>0</KEY>
              </TABLE-ROW>
            </TABLE>
```
with odxtools, this param would be parsed as a tablekey with a dummy table row in ecu shared data layer.
But in the file which contains basevariant, there is a new definition of table "TABLE_Service22_DataID" which contains real table rows. But with odxtools, the related param of base variant object is still using the dummy table.

I checked the code, and found that when resolving objects from parent, the odxtools will just get the objects from parent and will not do resolve snrefs for those objects.  And as those inherited objects are the same objects as the ones of parent, doing resolve snrefs based on current context (with base variant) will also modify the objects of the parent, which shall be a new issue. 

[...]

PR #373, opened 2024-12-10
resolve snrefs for child items

mafei1982:
for issue #368, current solution is to resolve snrefs based on current variant context before trying to access it. To get the correct result, the snrefs of child items  should be resolved when calling its parent's _resolve_snrefs. So I add code to do the work for some missing items

### DIFF
--- a/odxtools/diagservice.py
+++ b/odxtools/diagservice.py
@@ -187,7 +187,12 @@ class DiagService(DiagComm):
         # here because ComparamInstance.short_name is only valid after
         # reference resolution
         self._comparams = NamedItemList(self.comparam_refs)
-
+        if self._request:
+            self._request._resolve_snrefs(context)
+        for response in self._positive_responses:
+            response._resolve_snrefs(context)
+        for response in self._negative_responses:
+            response._resolve_snrefs(context)
         context.diag_service = None
 
     def decode_message(self, raw_message: bytes) -> Message:
--- a/odxtools/parameters/tablekeyparameter.py
+++ b/odxtools/parameters/tablekeyparameter.py
@@ -91,6 +91,7 @@ class TableKeyParameter(Parameter):
                 self._table_row = odxlinks.resolve(self.table_row_ref, TableRow)
             else:
                 self._table_row = odxlinks.resolve(self.table_row_ref)
+            self._table_row._resolve_odxlinks(odxlinks)
             self._table = self._table_row.table
 
     @override
--- a/odxtools/field.py
+++ b/odxtools/field.py
@@ -89,3 +89,7 @@ class Field(ComplexDop):
         if self.env_data_desc_snref is not None:
             self._env_data_desc = resolve_snref(self.env_data_desc_snref, ddds.env_data_descs,
                                                 EnvironmentDataDescription)
+        if self._structure:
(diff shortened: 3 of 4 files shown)


CLA-Mercedes-Benz:
[CLA assistant check](https://cla.opensource.mercedes-benz.com/mercedes-benz/odxtools?pullRequest=373) <br/>Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.opensource.mercedes-benz.com/mercedes-benz/odxtools?pullRequest=373) before we can accept your contribution.<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.opensource.mercedes-benz.com/check/mercedes-benz/odxtools?pullRequest=373) it.</sub>

odxtools/parameters/tablekeyparameter.py
```diff
@@ -91,6 +91,7 @@ def _resolve_odxlinks(self, odxlinks: OdxLinkDatabase) -> None:
                 self._table_row = odxlinks.resolve(self.table_row_ref, TableRow)
             else:
                 self._table_row = odxlinks.resolve(self.table_row_ref)
+            self._table_row._resolve_odxlinks(odxlinks)
```
mafei1982, line 94: Without this line, the parser will raise exception on below line, said that the "table" of tablerow does not exist.

andlaus: I see. That said, `_resolve_odxlinks()` must not be called for objects that are the result of a reference resolution lest there be cyclic reference resolution hell. as an alternative to this patch line 94 should better be replaced by
```suggestion(
            self._table = odxlinks.resolve(self._table_row.table_ref, Table)
```

odxtools/diagservice.py
```diff
@@ -187,7 +187,12 @@ def _resolve_snrefs(self, context: SnRefContext) -> None:
         # here because ComparamInstance.short_name is only valid after
         # reference resolution
         self._comparams = NamedItemList(self.comparam_refs)
-
+        if self._request:
+            self._request._resolve_snrefs(context)
+        for response in self._positive_responses:
+            response._resolve_snrefs(context)
+        for response in self._negative_responses:
+            response._resolve_snrefs(context)
```
andlaus, line 195: while this might fix the specific problem which you've encountered, this is pretty much a coincidence: this causes the referenced objects to be retargeted to basically the last ECU variant encountered during the reference resolution procedure. If you're looking at another diagnostic layer, e.g. the base variant, the snrefs will be incorrect...

mafei1982: yes, u are correct. So with this change, I need to call _resolve_snrefs again before I try to visit other variants. Since the child variant would use parent objects (References) directly, I cannot find other solutions for this problem.

andlaus: I guess #374 might make your life considerably easier?!

odxtools/parameters/tablekeyparameter.py
andlaus: do you want you change the PR to only fix this issue (and remove the SNREF changes) or do you prefer if I do it?

andlaus:
I suppose that this has become obsolete with #374. I'll close this for now...

PR #374, opened 2024-12-13
Add a `retarget_snrefs()` utility function

andlaus:
The problem is that accessing objects specified via short name reference always requires a context that is specific to the path via these objects are accessed; IOW, strictly speaking, resolving SNREFs cannot be done ahead of time (without copying everything for each possible context). Since odxtools is build on the assumption that all references can be resolved when a database is loaded (i.e., ahead of time) and copying everything for each possible context would be quickly running into RAM limitations, we provide `retarget_snrefs()`, a utility function that re-resolves the SNREFs reachable by a specified diagnostic layer using that diagnostic layer. While this is not a "100% solution" because sometimes contexts of SNREFs are more fine-grained than the diagnostic layer that contains them, it should be a workable solution for about 95% of people who run into this issue.

Usage:

```python
import odxtools
from odxtools.utils import retarget_snrefs

db = odxtools.load_file("MyEcu.pdx")

retarget_snrefs(db, db.ecu_variants.Variant2)
```

After this, all objects specified via SNREFs reachable by the diagnostic layer "Variant2" will be resolved using the point of view of Variant2. Be aware that only a single variant can be "active" for a given database, i.e., in the example above, accessing a resolved object via a diagnostic layer other than "Variant2" will still yield the same result as for Variant2...


andlaus:
note: while this does not "solve" #368, it should reduce the pain considerably...

andlaus:
@mafei1982: does retargeting the SNREFs to the ECU of interest fix the problem for you?

odxtools/utils.py
```diff
@@ -1,7 +1,48 @@
...
+    to another one.
+
+    """
+    from .snrefcontext import SnRefContext
+
+    if context is None:
+        context = SnRefContext()
+
+    if context.diag_layer is None:
+        context.database = database
```
kayoub5, line 36: why only assign database in the check of diag layer?

odxtools/utils.py
```diff
@@ -1,7 +1,48 @@
...
-from typing import Any, Dict
+from typing import TYPE_CHECKING, Any, Dict, Optional
+
+if TYPE_CHECKING:
+    from .database import Database
+    from .diaglayers.diaglayer import DiagLayer
+    from .snrefcontext import SnRefContext
+
+
+def retarget_snrefs(database: "Database",
```
kayoub5, line 12: missing unit test/coverage

odxtools/utils.py
andlaus: fixed: https://github.com/mercedes-benz/odxtools/pull/374/commits/9a9cc0bb0f2b04a7e4d44ea46fbef83c343fcc21

andlaus:
@kayoub5: do you see any issues which prevents merging this one?

kayoub5:
> @kayoub5: do you see any issues which prevents merging this one?

I don't see code issues, but the concept itself feels.... wrong

andlaus:
> I don't see code issues, but the concept itself feels.... wrong

If you mean the concept of short name refs, I agree ;). Anyway, since it is okay with you and this functionality comes with zero side effects for people who don't have "Schrödinger's cat in a box"-type ECUs, I'll merge this...

PR #379, opened 2025-01-14
fix en- and decoding of environment data descriptions

andlaus:
it turns out that the specification only mandates that the base data type of the DOP used by the referenced parameter must be A_UINT32, but the DOP does not need to be a DTC-DOP.

### DIFF
--- a/odxtools/environmentdatadescription.py
+++ b/odxtools/environmentdatadescription.py
@@ -15,9 +15,13 @@ from .exceptions import odxraise, odxrequire
 from .nameditemlist import NamedItemList
 from .odxlink import OdxDocFragment, OdxLinkDatabase, OdxLinkId, OdxLinkRef
 from .odxtypes import DataType, ParameterValue, ParameterValueDict
+from .parameters.codedconstparameter import CodedConstParameter
 from .parameters.parameter import Parameter
+from .parameters.parameterwithdop import ParameterWithDOP
+from .parameters.physicalconstantparameter import PhysicalConstantParameter
 from .parameters.valueparameter import ValueParameter
 from .snrefcontext import SnRefContext
+from .standardlengthtype import StandardLengthType
 from .utils import dataclass_fields_asdict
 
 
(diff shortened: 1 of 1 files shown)


odxtools/environmentdatadescription.py
```diff
@@ -116,40 +118,64 @@ def encode_into_pdu(self, physical_value: Optional[ParameterValue],
...
-        dtc_param_value: Optional[ParameterValue] = None
+        numerical_dtc_value: Optional[ParameterValue] = None
         for prev_param, prev_param_value in reversed(encode_state.journal):
             if prev_param.short_name == self.param_snref:
-                dtc_param = prev_param
-                prev_dop = getattr(prev_param, "dop", None)
-                if not isinstance(prev_dop, DtcDop):
-                    odxraise(f"The DOP of the parameter referenced by environment data "
-                             f"descriptions must be a DTC-DOP (is '{type(prev_dop).__name__}')")
+                if not isinstance(prev_param, ValueParameter):
```
kayoub5, line 132: it could be a constant parameter? no?

odxtools/environmentdatadescription.py
```diff
@@ -177,40 +203,64 @@ def decode_from_pdu(self, decode_state: DecodeState) -> ParameterValue:
...
+                        odxraise(f"The data type used by the DOP of the parameter referenced "
+                                 f"by environment data descriptions must be A_UINT32 "
+                                 f"(encountered '{prev_dop.diag_coded_type.base_data_type.value}')")
+                        return
+
+                    if prev_param_value is None:
+                        odxraise()
+                        return
+
+                    numerical_dtc_value = prev_dop.convert_to_numerical_trouble_code(
```
kayoub5, line 236: this block feels copy-pasted

odxtools/environmentdatadescription.py
andlaus: I don't think so: environment data descriptions are a multiplexer mechanism to attach additional information to a DTC depending on the value of the trouble code (basically this data describes the circumstances under which the DTC was encountered). mandating the key of the multiplexer to be a constant defies the purpose? (haven't checked the spec about this, though..

odxtools/environmentdatadescription.py
```diff
@@ -116,40 +118,64 @@ def encode_into_pdu(self, physical_value: Optional[ParameterValue],
...
+                        prev_param_value)
+                elif isinstance(prev_dop, DataObjectProperty):
+                    if prev_dop.diag_coded_type.base_data_type != DataType.A_UINT32:
+                        odxraise(f"The data type used by the DOP of the parameter referenced "
+                                 f"by environment data descriptions must be A_UINT32 "
+                                 f"(encountered '{prev_dop.diag_coded_type.base_data_type.value}')")
+                        return
+
+                    if not isinstance(prev_param_value, (int, str)):
+                        odxraise()
```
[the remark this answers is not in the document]
andlaus: basically the only reason for this construct is to make `mypy` happpy: on line 154 we already check that we get a `A_UINT32`...

odxtools/environmentdatadescription.py
kayoub5: if the dtc have only one environment data, then it could be in theory a constant.

the main idea, is that the check you are doing either needs to backed by a specification clause, or to prevent an error for something not supported by the rest of the code.

the check here does not seem to fit any of those conditions.

odxtools/environmentdatadescription.py
kayoub5: Introduce a function, that takes `prev_dop` and `prev_param_value` as input and return `numerical_dtc_value`

andlaus: good point. Done: https://github.com/mercedes-benz/odxtools/pull/379/commits/a9a791f804540f070914559fb391bb0dca4dda4e

odxtools/environmentdatadescription.py
```diff
@@ -177,40 +169,26 @@ def decode_from_pdu(self, decode_state: DecodeState) -> ParameterValue:
...
-                             f"descriptions must be a DTC-DOP (is '{type(prev_dop).__name__}')")
-                    return
-                dtc_dop = prev_dop
-                dtc_param_value = prev_param_value
+                numerical_dtc_value = self._get_numerical_dtc_from_parameter(
+                    prev_param, prev_param_value)
                 break
 
-        if dtc_param is None:
+        if numerical_dtc_value is None:
```
kayoub5, line 187: Check could be moved inside `_get_numerical_dtc_from_parameter`, since it's done for each call

odxtools/environmentdatadescription.py
```diff
@@ -228,11 +206,62 @@ def decode_from_pdu(self, decode_state: DecodeState) -> ParameterValue:
...
+                         f"must use either be StandardLengthType or a DtcDop (encountered "
+                         f"{type(param).__name__} for parameter '{self.param.short_name}' "
+                         f"of ENV-DATA-DESC '{self.short_name}')")
+                return
+
+            if dop.diag_coded_type.base_data_type != DataType.A_UINT32:
+                odxraise(f"The data type used by the DOP of the parameter referenced "
+                         f"by environment data descriptions must be A_UINT32 "
+                         f"(encountered '{dop.diag_coded_type.base_data_type.value}')")
+                return
```
kayoub5, line 233: The check is only to be spec compliant, no need to halt here as far as code goes
```suggestion
```

odxtools/environmentdatadescription.py
```diff
@@ -228,11 +206,62 @@ def decode_from_pdu(self, decode_state: DecodeState) -> ParameterValue:
...
+                return dop.compu_method.convert_physical_to_internal(param_value)
+
+            odxraise()  # not reachable
+
+        elif isinstance(param, CodedConstParameter):
+            if param.diag_coded_type.base_data_type != DataType.A_UINT32:
+                odxraise(f"The data type used by the parameter referenced "
+                         f"by environment data descriptions must be A_UINT32 "
+                         f"(encountered '{param.diag_coded_type.base_data_type.value}')")
+                return
```
kayoub5, line 256: The check is only to be spec compliant, no need to halt here as far as code goes
```suggestion
```

odxtools/environmentdatadescription.py
andlaus: if we don't return here, it is basically assured that a different exception like `AttributeError` will be raised by the remainder of the method...

odxtools/environmentdatadescription.py
andlaus: the check here is only triggered if no parameter with the specified short name has been found (`_get_numerical_dtc_from_parameter()` always returns an integer)

kayoub5 (approved):

PR #381, opened 2025-01-17
Implement all encodings of base data type

andlaus:
For integers, this means packed and unpacked BCD, two-complement, one-complement and sign-magnitude. Strings can be represented using UTF-8, UTF-16 (big or little endian), ISO 8859-1 (latin1), ISO 8859-2 (latin2) and Windows codepage 1252. Finally, floating point values may be encoded using the 32 and 64 bit ISO 754 representation.

Be aware that I'm not sure if the bit mask is handled correctly for negative integers (or if there even is a "correct" handling of bit masks for negative integers). Also, I suspect that most ODX implementations encountered in the wild behave slightly differently for non byte-aligned negative integers...

Besides this, the PR includes a small fix to make the `.odx-d` file of the somersault ECU compliant with the ODX XSD: It turns out that for whatever reason only empty DESC tags can be specified for parameters (this either is an issue with the XSD or a bug in the XML validator which I like to use (`xmllint`)), so, as a work-around, the description of `schroedinger_param` is moved back to `.long_name`.


odxtools/decodestate.py
```diff
@@ -87,32 +93,125 @@ def extract_atomic_value(
...
+            elif base_type_encoding == Encoding.ISO_8859_1:
+                internal_value = raw_value.decode("iso-8859-1", errors=text_errors)
+            elif base_type_encoding == Encoding.ISO_8859_2:
+                internal_value = raw_value.decode("iso-8859-2", errors=text_errors)
+            elif base_type_encoding == Encoding.WINDOWS_1252:
+                internal_value = raw_value.decode("cp1252", errors=text_errors)
+            else:
+                odxassert(
+                    base_type_encoding in (None, Encoding.NONE),
+                    f"Specified illegal encoding {base_type_encoding} for string object")
```
kayoub5, line 126: Specs explicitly forbid this?

odxtools/decodestate.py
```diff
@@ -87,32 +93,125 @@ def extract_atomic_value(
...
+                    # values 0-127.  In practice, vendors use
+                    # iso-8859-1, aka latin-1, because iso-8859-1
+                    # never fails since it has a valid character
+                    # mapping for every possible value
+                    internal_value = raw_value.decode("iso-8859-1", errors=text_errors)
+
+        # ... integers, ...
+        elif base_data_type in (DataType.A_INT32, DataType.A_UINT32):
+            if not isinstance(raw_value, int):
+                odxraise(f"Raw value must be of integer type, not {type(raw_value).__name__}")
```
kayoub5, line 147: Could be a byte array, in case for BCP for example

odxtools/decodestate.py
```diff
@@ -87,32 +93,125 @@ def extract_atomic_value(
...
+                    # python defines the bitwise inversion of a
+                    # positive integer value x as ~x = -(x + 1).
+                    internal_value = -((1 << bit_length) - raw_value - 1)
+            elif base_type_encoding == Encoding.TWOC:
+                # two-complement
+                sign_bit = 1 << (bit_length - 1)
+                if raw_value < sign_bit:
+                    internal_value = raw_value
+                else:
+                    internal_value = -((1 << bit_length) - raw_value)
```
kayoub5, line 183: There is a special case of 2C you need to check, if bit length = 1, and 2C is used, then the encoding should map 0 to 0, and 1 to -1

odxtools/decodestate.py
```diff
@@ -87,32 +93,125 @@ def extract_atomic_value(
...
+                internal_value = raw_value
+
+        # ... and others (floating point values)
+        else:
+            odxassert(base_data_type in (DataType.A_FLOAT32, DataType.A_FLOAT64))
+            odxassert(
+                base_type_encoding in (None, Encoding.NONE),
+                f"Specified illegal encoding '{base_type_encoding}' for float object")
+
+            if base_data_type == DataType.A_FLOAT32 and bit_length != 32:
```
kayoub5, line 205: Already checked above

odxtools/encodestate.py
```diff
@@ -86,66 +87,141 @@ def emplace_atomic_value(
...
+                raw_value = internal_value.encode("utf-8")
+            elif base_type_encoding == Encoding.UCS2:
+                text_encoding = "utf-16-be" if is_highlow_byte_order else "utf-16-le"
+                raw_value = internal_value.encode(text_encoding)
+            elif base_type_encoding == Encoding.ISO_8859_1:
+                raw_value = internal_value.encode("iso-8859-1")
+            elif base_type_encoding == Encoding.ISO_8859_2:
+                raw_value = internal_value.encode("iso-8859-2")
+            elif base_type_encoding == Encoding.WINDOWS_1252:
+                raw_value = internal_value.encode("cp1252")
```
kayoub5, line 124: Copy-paste, extract logic into a function

odxtools/odxtypes.py
```diff
@@ -157,7 +157,7 @@ def compare_odx_values(a: AtomicOdxType, b: AtomicOdxType) -> int:
 
 # format specifiers for the data type using the bitstruct module
 _BITSTRUCT_FORMAT_LETTER_MAP__ = {
-    "A_INT32": "s",
+    "A_INT32": "u",  # the sign must be handled separately
```
kayoub5, line 160: Does the new code, handle the case base_type_encoding is None and A_INT32 used?

odxtools/templates/macros/printDOP.xml.jinja2
```diff
@@ -12,7 +12,7 @@
 
 {%- macro printDiagCodedType(dct) -%}
 <DIAG-CODED-TYPE {{-make_xml_attrib("BASE-DATA-TYPE", dct.base_data_type.value)}}
-                 {{-make_xml_attrib("BASE-TYPE-ENCODING", dct.base_type_encoding)}}
+                 {{-make_xml_attrib("BASE-TYPE-ENCODING", dct.base_type_encoding and dct.base_type_encoding.value)}}
```
kayoub5, line 15: Why?

tests/test_encoding.py
```diff
@@ -88,6 +89,300 @@ def test_encode_coded_const_infer_order(self) -> None:
         )
         self.assertEqual(req.encode(), bytearray([0x7D, 0xAB]))
 
+    def test_string_encodings(self) -> None:
+        # ISO 8859-1 (latin1)
+        encode_state = EncodeState()
+        encode_state.emplace_atomic_value(
+            internal_value="ä?a",
```
kayoub5, line 96: What's this?

tests/test_encoding.py
```diff
@@ -88,6 +89,300 @@ def test_encode_coded_const_infer_order(self) -> None:
...
+            base_type_encoding=Encoding.SM,
+            is_highlow_byte_order=True)
+        self.assertEqual(decoded, -0x1234)
+
+    def test_float_encodings(self) -> None:
+        # FLOAT32
+        encode_state = EncodeState()
+        with self.assertRaises(OdxError):
+            encode_state.emplace_atomic_value(
+                internal_value=-1.234,
```
kayoub5, line 324: Add test for magic float values, like NaN and infinity

tests/test_encoding.py
```diff
@@ -88,6 +89,300 @@ def test_encode_coded_const_infer_order(self) -> None:
...
+                bit_length=24,
+                base_data_type=DataType.A_FLOAT64,
+                base_type_encoding=Encoding.NONE,
+                is_highlow_byte_order=True,
+                used_mask=None)
+        encode_state.emplace_atomic_value(
```
kayoub5, line 363: Copy paste, maybe convert to parametrized test case?

tests/test_encoding.py
```diff
@@ -88,6 +89,300 @@ def test_encode_coded_const_infer_order(self) -> None:
...
+                bit_length=24,
+                base_data_type=DataType.A_FLOAT64,
+                base_type_encoding=Encoding.NONE,
+                is_highlow_byte_order=True)
+        decoded = decode_state.extract_atomic_value(
+            bit_length=64,
+            base_data_type=DataType.A_FLOAT64,
+            base_type_encoding=Encoding.NONE,
+            is_highlow_byte_order=True)
+        self.assertTrue(abs(float(decoded) - (-1.234)) < 1e-9)
```
kayoub5, line 384: What's this?

odxtools/decodestate.py
andlaus: good point. In my data sets this is the case, but after checking the spec, the default encoding for strings depends on the base data type (cf table 8) and so do the default encodings of integers. fixed: https://github.com/mercedes-benz/odxtools/pull/381/commits/c84241bad312d0e59f8d63060ba8e2a1a0178c27

odxtools/decodestate.py
andlaus: in that case, `bitstruct` would extract a bytes object because we use the format letter `"r"`. since mypy can obviously not understand such semantic behavior, we help it out a bit here...

odxtools/decodestate.py
andlaus: I think this already is what happens here:

```
raw_value = 1
-> - (2 -1) = -1
raw value  = 0
-> 0
```

odxtools/encodestate.py
andlaus: this is only _almost_ copy-and-paste: `DecodeState` needs to call `.decode()` instead of `.encode()`

odxtools/minmaxlengthtype.py
```diff
@@ -78,20 +78,33 @@ def encode_into_pdu(self, internal_value: AtomicOdxType, encode_state: EncodeSta
...
 
-        if self.max_length is not None:
-            data_length = min(len(internal_value), self.max_length)
-        else:
-            data_length = len(internal_value)
+        # TODO: This assumes that each character of a string is
+        #       encoded into a single byte. This is never the case for
+        #       UTF-16 encoded strings and does not always hold for
+        #       UTF-8. We ignore this issue for now...
+        data_length = len(internal_value)
```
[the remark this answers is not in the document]
andlaus: sure, but I dealing with the string encodings sucks, so let's assume that these values are either byte arrays or ASCII strings.

odxtools/odxtypes.py
andlaus: yes. It needs to deal with negative numbers manually, though. (before this, negative numbers always used the two-complement.)

odxtools/templates/macros/printDOP.xml.jinja2
andlaus: because before this PR, `.base_type_encoding` was of type `Optional[str]`, now it is `Optional[Encoding]`.

tests/test_encoding.py
andlaus: this is a test case containing a non-ASCII character (`ä`. mainly because it was conveniently accessible on my keyboard.)

tests/test_encoding.py
andlaus: done: https://github.com/mercedes-benz/odxtools/pull/381/commits/6a6a37d7380c31eca698d3bafde4a1362dc63818

tests/test_encoding.py
andlaus: I don't think it would be worth it: the parameters for that test case would basically be the same as for `.emplace_atomic_value()`

tests/test_encoding.py
andlaus: this is a potential rounding error. (I don't want to rely on the fact that python uses 64 bit floating point values internally...)

odxtools/encodestate.py
kayoub5: create function that takes the coded type as input, and return the correct string encoding name that python uses

tests/test_encoding.py
kayoub5: put that info as code comment

odxtools/odxtypes.py
andlaus: in case your question was what happens if the base type is not specified: specifying a base type is mandatory for all occasions where a base type encoding can be specified. (i.e., for DIAG-CODED-TYPE)

odxtools/decodestate.py
andlaus: ok, you meant *way* above. fixed. (this was due to the usual code churn...)

odxtools/decodestate.py
kayoub5: and raw_value would bytes instead of int, no?

odxtools/encodestate.py
```diff
@@ -86,66 +87,145 @@ def emplace_atomic_value(
...
+                    EncodeError)
 
-        elif base_data_type == DataType.A_UNICODE2STRING:
-            if not isinstance(internal_value, str):
-                odxraise()
+        # ... signed integers, ...
+        elif base_data_type == DataType.A_INT32:
+            if not isinstance(internal_value, int):
+                odxraise(
+                    f"Internal value must be of integer type, not {type(internal_value).__name__}",
```
kayoub5, line 152: How would this happen?

odxtools/encodestate.py
```diff
@@ -86,66 +87,145 @@ def emplace_atomic_value(
...
+                    raw_value = (1 << (bit_length - 1)) + abs(internal_value)
+            else:
+                odxraise(f"Illegal encoding ({base_type_encoding}) specified for "
+                         f"{base_data_type.value}")
+
+                internal_value = raw_value
+
+        # ... unsigned integers, ...
+        elif base_data_type == DataType.A_UINT32:
+            if not isinstance(internal_value, int) or internal_value < 0:
```
kayoub5, line 183: Didn't you put `u` in bitstruct logic, so that would not happen?

odxtools/encodestate.py
```diff
@@ -86,66 +87,145 @@ def emplace_atomic_value(
...
+                internal_value = raw_value
 
-            if 8 * len(raw_value) > bit_length:
-                raise EncodeError(f"The string {repr(internal_value)} is too large."
-                                  f" The maximum number of characters is {bit_length//16}.")
+        # ... and others (floating point values)
         else:
-            raw_value = internal_value
+            odxassert(base_data_type in (DataType.A_FLOAT32, DataType.A_FLOAT64))
+            odxassert(base_type_encoding in (None, Encoding.NONE))
```
kayoub5, line 216: `foo in (None, bar)` works?

odxtools/encodestate.py
```diff
@@ -161,25 +241,21 @@ def emplace_atomic_value(
...
-        if base_data_type in [DataType.A_INT32, DataType.A_UINT32
-                             ] and (self.cursor_bit_position != 0 or
-                                    (self.cursor_bit_position + bit_length) % 8 != 0):
-            if used_mask is None:
-                tmp = (1 << bit_length) - 1
-            else:
-                tmp = int.from_bytes(used_mask, "big")
-            tmp <<= self.cursor_bit_position
 
+        if used_mask_raw is None:
```
kayoub5, line 245: Should the mask be applied before or after base_type_encoding ?

odxtools/minmaxlengthtype.py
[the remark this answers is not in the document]
kayoub5: After all the effort you spent on the topic in the other functions, you will just half here?

tests/test_encoding.py
```diff
@@ -88,6 +90,330 @@ def test_encode_coded_const_infer_order(self) -> None:
...
+        # potentially using a different representation
+        self.assertTrue(abs(float(decoded) - (-1.234)) < 1e-6)
+
+        encode_state = EncodeState()
+        encode_state.emplace_atomic_value(
+            internal_value=float("nan"),
+            bit_length=32,
+            base_data_type=DataType.A_FLOAT32,
+            base_type_encoding=Encoding.NONE,
+            is_highlow_byte_order=True,
```
kayoub5, line 370: Not going to test little endian?

odxtools/decodestate.py
andlaus: currently, we always extract an unsigned integer and have to deal with the bit fiddling ourselfs. The reason is that I don't think there is a good pythonic way of converting a bytes string using an exotic encoding (1C, BCD-P, etc) to an integer, even less so if that number is not byte-aligned...

odxtools/encodestate.py
andlaus: okay, I budge (also for the `MinMaxLengthType` issue below: https://github.com/mercedes-benz/odxtools/pull/381/commits/7de4a81490072c1e6a6287aef84a7e7a6d60477b . Thanks for the persistent nagging :)

odxtools/minmaxlengthtype.py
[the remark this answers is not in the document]
andlaus: I budge: https://github.com/mercedes-benz/odxtools/pull/381/commits/7de4a81490072c1e6a6287aef84a7e7a6d60477b . Thanks for the insistence.

odxtools/decodestate.py
```diff
@@ -87,32 +95,125 @@ def extract_atomic_value(
...
+            else:
+                odxraise(f"Illegal encoding ({base_type_encoding}) specified for "
+                         f"{base_data_type.value}")
+
+                internal_value = raw_value
+
+        # ... unsigned integers, ...
+        elif base_data_type == DataType.A_UINT32:
+            if not isinstance(raw_value, int) or raw_value < 0:
+                odxraise(f"Raw value must be a positive integer, not {raw_value}")
```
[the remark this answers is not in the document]
andlaus: Currently it should not be possible because we always extract unsigned integers using bitstruct. This is mainly to prevent things breaking if this gets changed in the future...

odxtools/encodestate.py
andlaus: for `.encode()`, `internal_value` is usually specified by the user?!

odxtools/encodestate.py
andlaus: this is a user-specified value. for `.decode()` this is correct...

odxtools/encodestate.py
andlaus: yes:
```python
>>> foo = None
>>> foo in (None, 123)
True
>>> foo in (123,)
False
```

odxtools/encodestate.py
andlaus: I think the bit mask screws things up for everything other than unsigned integers without an encoding, so I guess it does not really matter?

odxtools/decodestate.py
kayoub5: That means `if not isinstance(raw_value, int):` will never be true

odxtools/decodestate.py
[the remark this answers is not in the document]
kayoub5: > This is mainly to prevent things from breaking if this gets changed in the future

That's what unit tests are for, I would not mind a few checks, but the function now have more checks that encoding/decoding logic

odxtools/encodestate.py
kayoub5: User specify physical value, not internal one

odxtools/encodestate.py
kayoub5: User specify physical value, not internal one

odxtools/decodestate.py
```diff
@@ -87,32 +95,109 @@ def extract_atomic_value(
...
-            # bytes to a string
-            text_encoding = "utf-16-be" if is_highlow_byte_order else "utf-16-le"
-            internal_value = internal_value.decode(encoding=text_encoding, errors=text_errors)
+        internal_value: AtomicOdxType
+
+        # Deal with raw byte fields, ...
+        if base_data_type == DataType.A_BYTEFIELD:
+            odxassert(
+                base_type_encoding is None or
+                base_type_encoding in (Encoding.NONE, Encoding.BCD_P, Encoding.BCD_UP))
```
kayoub5, line 108: No need for extra ` base_type_encoding is None or`
```suggestion
                base_type_encoding in (None, Encoding.NONE, Encoding.BCD_P, Encoding.BCD_UP))
```

odxtools/decodestate.py
```diff
@@ -87,32 +95,109 @@ def extract_atomic_value(
...
+            if base_type_encoding == Encoding.ONEC:
+                # one-complement
+                sign_bit = 1 << (bit_length - 1)
+                if raw_value < sign_bit:
+                    internal_value = raw_value
+                else:
+                    # python defines the bitwise inversion of a
+                    # positive integer value x as ~x = -(x + 1).
+                    internal_value = -((1 << bit_length) - raw_value - 1)
+            elif base_type_encoding == Encoding.TWOC or base_type_encoding is None:
```
kayoub5, line 139:
```suggestion
            elif base_type_encoding in (None, Encoding.TWOC):
```

odxtools/encodestate.py
```diff
@@ -86,66 +87,128 @@ def emplace_atomic_value(
...
+                return
 
-            # The spec says ASCII, meaning only byte values 0-127.
-            # But in practice, vendors use iso-8859-1, aka latin-1
-            # reason being iso-8859-1 never fails since it has a valid
-            # character mapping for every possible byte sequence.
-            raw_value = internal_value.encode("iso-8859-1")
+            odxassert(
+                base_type_encoding is None or
+                base_type_encoding in (Encoding.NONE, Encoding.BCD_P, Encoding.BCD_UP))
```
kayoub5, line 106:
```suggestion
                base_type_encoding in (None, Encoding.NONE, Encoding.BCD_P, Encoding.BCD_UP))
```

odxtools/encodestate.py
kayoub5: I am asking about what the specs say

odxtools/encoding.py
```diff
@@ -0,0 +1,56 @@
...
+    UTF8 = "UTF-8"
+    UCS2 = "UCS-2"
+    ISO_8859_1 = "ISO-8859-1"
+    ISO_8859_2 = "ISO-8859-2"
+    WINDOWS_1252 = "WINDOWS-1252"
+
+    NONE = "NONE"
+
+
+def get_string_encoding(base_data_type: DataType, base_type_encoding: Optional[Encoding],
```
kayoub5, line 26: Should this be moved to DiagCodedType ?

odxtools/encodestate.py
andlaus: yes, but e.g., if the compu method is IDENTICAL and the user specifies a negative value, it still arrives here...

odxtools/encodestate.py
andlaus: I might have missed something, but IIRC it basically says "mask out the specified bits" which IMO does not make sense for negative numbers or BCD coded ones...

odxtools/encoding.py
andlaus: unfortuntely, this is not possible because the `DiagCodedType` object is not available in `EncodeState` and `DecodeState`

odxtools/decodestate.py
andlaus: done: https://github.com/mercedes-benz/odxtools/pull/381/commits/356702a5dffca623bc881d4be1d2d443d148fe07

odxtools/encodestate.py
andlaus: done

odxtools/decodestate.py
andlaus: done

odxtools/decodestate.py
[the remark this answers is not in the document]
andlaus: okay, I'm rather indifferent about this. changed: https://github.com/mercedes-benz/odxtools/pull/381/commits/5c698b549f4dbbb29223b7396bc77d680322c3c1

odxtools/encodestate.py
andlaus: (see my comment below)

odxtools/decodestate.py
andlaus: yes, maybe this was a bit too defensive. changed in https://github.com/mercedes-benz/odxtools/pull/381/commits/5c698b549f4dbbb29223b7396bc77d680322c3c1...

kayoub5 (approved):

Issue #286, opened 2024-04-06
increase minimum python version to 3.10


andlaus:
most importantly, this will allow us to use `kw_only` for dataclasses and thus significantly simplify "manual" object creation 
(because with `kw_only` dataclasses, it is possible to use defaulted attributes and class inheritance simultaneously)

currently, this is blocked by the fact that we aim for all maintained Ubuntu LTS releases to be supported out of the box. The first release that ships with python >= 3.10 is 22.04, i.e., until Ubuntu 20.04 reaches its end of life in April 2025, we are stuck with python 3.8...

andlaus:
- use the build-in container types (`dict`, `list`, ...) instead of the ones imported from `typing` for type annotations
- for type annotations, use `|` instead of `typing.Union` to express multiple possibilities

andlaus:
- revert https://github.com/mercedes-benz/odxtools/pull/382/commits/b56f13d79fd412fa0641cac2065e585427bbc5a7

andlaus:
with #406, this has been implemented.

PR #382, opened 2025-01-21
implement support for condensed bit masks

andlaus:
IMO this is a somewhat exotic feature, but since it is in the spec there probably is somebody out there who uses it...

### DIFF
--- a/odxtools/standardlengthtype.py
+++ b/odxtools/standardlengthtype.py
@@ -3,7 +3,7 @@ from dataclasses import dataclass
 from typing import List, Literal, Optional
 from xml.etree import ElementTree
 
-from typing_extensions import override
+from typing_extensions import SupportsBytes, override
 
 from .decodestate import DecodeState
 from .diagcodedtype import DctType, DiagCodedType
@@ -91,8 +91,8 @@ class StandardLengthType(DiagCodedType):
             return used_mask.to_bytes((bit_sz + 7) // 8, endianness)
 
         sz: int
-        if isinstance(internal_value, (bytes, bytearray)):
-            sz = len(internal_value)
+        if isinstance(internal_value, (bytearray, SupportsBytes)):
+            sz = len(bytes(internal_value))
         else:
             sz = (odxrequire(self.get_static_bit_length()) + 7) // 8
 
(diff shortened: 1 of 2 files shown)


odxtools/standardlengthtype.py
```diff
@@ -82,53 +78,132 @@ def __get_raw_mask(self, internal_value: AtomicOdxType) -> Optional[bytes]:
             # been anticipated by the standard, though...
             endianness = "little"
 
+        if self.is_condensed:
+            # if a condensed bitmask is specified, the number of bits
+            # set to one in the bit mask are used in the PDU
+
+            # TODO: this is pretty slow. replace it by
+            # `self.bit_mask.bit_count()` once we require python >=
+            # 3.10.
```
kayoub5, line 87: we are not that far away from 3.9 end of life, you may drop it now

odxtools/standardlengthtype.py
```diff
@@ -82,53 +78,132 @@ def __get_raw_mask(self, internal_value: AtomicOdxType) -> Optional[bytes]:
...
 
     def __apply_mask(self, internal_value: AtomicOdxType) -> AtomicOdxType:
         if self.bit_mask is None:
             return internal_value
+
         if self.is_condensed:
-            odxraise("Serialization of condensed bit mask is not supported", NotImplementedError)
-            return
+            int_value: int
+            if isinstance(internal_value, bytes):
```
kayoub5, line 110: what about bytearray ?

odxtools/standardlengthtype.py
```diff
@@ -82,53 +78,132 @@ def __get_raw_mask(self, internal_value: AtomicOdxType) -> Optional[bytes]:
...
+                int_value = int.from_bytes(internal_value, 'big')
```
kayoub5, line 111: why always big?

odxtools/standardlengthtype.py
andlaus: I promised myself to support python 3.8 until ubuntu 20.04 is EOL. Fortunately, this is not too far off anymore (April), then we can do the same for Ubuntu 22.04 / python 3.10. I've added an item to the TODO list, though: https://github.com/mercedes-benz/odxtools/issues/286 . (That said, I suspect that the performance impact of converting the integer to a string and counting the one characters is lower than the interpreter overhead of doing two instead of one function calls. For compiled languages this is a completely different story...)

odxtools/standardlengthtype.py
andlaus: good point. fixed: https://github.com/mercedes-benz/odxtools/pull/382/commits/e4d7724855e77e89d0c259c5e4906358b4189a43

I'm not sure about the performance implications of `isinstance(..., SupportsBytes)` because I suppose that `SupportsBytes` is a type protocol and checking these is slow, but the vast majority of DOPs should not specify a bitmask, this should not matter too much...

kayoub5 (approved):

PR #386, opened 2025-02-12
Miscellaneous fixes

andlaus:
This PR contains a large number of patches for smaller issues which IMO do not merit separate pull requests. Since all patches of this series are self-contained, please let me know if you prefer it to be split into two or three PRs...


examples/somersaultecu.py
```diff
@@ -268,6 +269,7 @@ class SomersaultSID(IntEnum):
             short_name="session",
             long_name="Session",
             description=None,
+            admin_data=None,
```
kayoub5, line 272: This is why I like trailing comma

examples/somersaultecu.py
```diff
@@ -1094,7 +1096,8 @@ class SomersaultSID(IntEnum):
             oid=None,
             odx_id=OdxLinkId("somersault.struct.recall.forward_flips_grudgingly_done", doc_frags),
             parameters=somersault_positive_responses["forward_flips_grudgingly_done"].parameters,
-            byte_size=None),
+            byte_size=None,
+            is_visible_raw=None),
```
kayoub5, line 1100: And this is why I hate when they are not used

odxtools/diagcomm.py
```diff
@@ -61,8 +61,8 @@ class DiagComm(IdentifiableElement):
     audience: Optional[Audience]
     protocol_snrefs: List[str]
     related_diag_comm_refs: List[RelatedDiagCommRef]
-    pre_condition_state_refs: Iterable[OdxLinkRef]
-    state_transition_refs: Iterable[OdxLinkRef]
+    pre_condition_state_refs: List[OdxLinkRef]
+    state_transition_refs: List[OdxLinkRef]
```
kayoub5, line 65:
```suggestion
    state_transition_refs: list[OdxLinkRef]
```

odxtools/diagnostictroublecode.py
```diff
@@ -15,7 +15,7 @@
 @dataclass
 class DiagnosticTroubleCode(IdentifiableElement):
     trouble_code: int
-    text: Optional[str]
+    text: str
```
kayoub5, line 18: Sure?

odxtools/diagnostictroublecode.py
```diff
@@ -46,7 +42,7 @@ def from_et(et_element: ElementTree.Element,
 
         return DiagnosticTroubleCode(
             trouble_code=int(odxrequire(et_element.findtext("TROUBLE-CODE"))),
-            text=et_element.findtext("TEXT"),
+            text=odxrequire(et_element.findtext("TEXT")),
```
kayoub5, line 45: Sure? even ODX 2.0 ?

odxtools/diagvariable.py
```diff
@@ -32,15 +34,30 @@ class DiagVariable(IdentifiableElement):
...
-    def variable_group(self) -> VariableGroup:
+    def table(self) -> Optional[Table]:
+        return self._table
+
+    @property
+    def table_row(self) -> Optional[TableRow]:
+        return self._table_row
+
+    @property
+    def variable_group(self) -> Optional[VariableGroup]:
```
kayoub5, line 60: New feature, not a bug fix, merit its own PR

odxtools/diagvariable.py
```diff
@@ -62,6 +78,16 @@ def from_et(et_element: ElementTree.Element, doc_frags: List[OdxDocFragment]) ->
             CommRelation.from_et(cr_elem, doc_frags)
             for cr_elem in et_element.iterfind("COMM-RELATIONS/COMM-RELATION")
         ]
+
+        table_snref = None
+        table_row_snref = None
+        if (snref_to_tablerow_elem := et_element.find("SNREF-TO-TABLEROW")) is not None:
```
kayoub5, line 84: New feature?

odxtools/inputparam.py
```diff
@@ -38,12 +38,12 @@ def _build_odxlinks(self) -> Dict[OdxLinkId, Any]:
...
     def _resolve_odxlinks(self, odxlinks: OdxLinkDatabase) -> None:
-        self._dop_base = odxlinks.resolve(self.dop_base_ref, DopBase)
+        self._dop = odxlinks.resolve(self.dop_base_ref, DopBase)
 
     def _resolve_snrefs(self, context: SnRefContext) -> None:
         pass
 
     @property
-    def dop_base(self) -> DopBase:
+    def dop(self) -> DopBase:
```
kayoub5, line 47: Breaking change?

odxtools/parameters/tablekeyparameter.py
```diff
@@ -23,10 +23,16 @@
...
+    # the spec mandates that exactly one of the two attributes must
+    # be non-None
     table_ref: Optional[OdxLinkRef]
     table_snref: Optional[str]
-    table_row_snref: Optional[str]
+
+    # the spec mandates that exactly one of the two attributes must
+    # be non-None
     table_row_ref: Optional[OdxLinkRef]
+    table_row_snref: Optional[str]
```
kayoub5, line 35: New feature?

odxtools/scaleconstr.py
```diff
@@ -25,8 +25,8 @@ class ScaleConstr:
 
     short_label: Optional[str]
     description: Optional[Description]
-    lower_limit: Optional[Limit]
-    upper_limit: Optional[Limit]
+    lower_limit: Limit
+    upper_limit: Limit
```
kayoub5, line 29: Sure?

tests/test_diag_data_dictionary_spec.py
```diff
@@ -264,6 +264,7 @@ def test_initialization(self) -> None:
             short_name="mux_case1_struct",
             long_name=None,
             description=None,
+            is_visible_raw=None,
```
kayoub5, line 267: Didn't you remove `is_visible_raw`  in a previous PR?

examples/somersaultecu.py
andlaus: no particular reason. probably because it makes it possible to paste this line at any given position (odxtools is inconsistent when it comes to this...)

examples/somersaultecu.py
andlaus: I can add them to whole codebase if you want, but that's for another PR...

odxtools/comparaminstance.py
```diff
@@ -138,9 +138,4 @@ def get_subvalue(self, subparam_name: str) -> Optional[str]:
 
     @property
     def short_name(self) -> str:
-        if self.spec:
-            return self.spec.short_name
-
-        # ODXLINK IDs allow dots and hyphens, but short names do not.
-        # (This should not happen anyway in a correct PDX...)
-        return self.spec_ref.ref_id.replace(".", "__").replace("-", "_")
```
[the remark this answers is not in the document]
andlaus: because accessing the short name before reference resolution IMO ought not to be supported. In particular because the short name would change depending on references have been resolved or not, so this IMO constitutes a medium-sized footgun...

odxtools/diagcomm.py
andlaus: yes. once we can rely on python 3.10, cf #286. The blocker is Ubuntu 20.04, which goes EOL in [April](https://ubuntu.com/about/release-cycle). I think we should be patient enough to wait the remaining three months...

odxtools/diagnostictroublecode.py
andlaus: quite: 
Screenshot_20250212_180129

odxtools/diagnostictroublecode.py
andlaus: I think so:

Screenshot_20250212_180328

odxtools/inputparam.py
andlaus: thanks! fixed: https://github.com/mercedes-benz/odxtools/pull/386/commits/517ab51d8eb57e8689e1bdd635e6e64cd493cdc3

odxtools/modification.py
```diff
@@ -3,18 +3,19 @@
...
 @dataclass
 class Modification:
-    change: Optional[str]
+    change: str
     reason: Optional[str]
 
     @staticmethod
     def from_et(et_element: ElementTree.Element, doc_frags: List[OdxDocFragment]) -> "Modification":
-        change = et_element.findtext("CHANGE")
+        change = odxrequire(et_element.findtext("CHANGE"))
```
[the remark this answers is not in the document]
andlaus: yes:

Screenshot_20250212_181434

(same for ODX 2.0)

odxtools/parameters/tablekeyparameter.py
andlaus: not really. table rows can be referenced using a reference to a table plus a SNREF to a row in addition to a ODXLINK reference to the table row. the ODX XSD is pretty twisted here, so I probably missed this before...

odxtools/scaleconstr.py
andlaus: yes: 
Screenshot_20250212_181809

(that's for ODX 2.0. 2.2 is the same)

tests/test_diag_data_dictionary_spec.py
andlaus: yes, #240 (I suppose). seems like I made a mistake back then. (IS-VISIBLE is only part of STRUCTURE, but not BASIC-STRUCTURE...)

odxtools/comparaminstance.py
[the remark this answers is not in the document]
kayoub5: spec is not available during object creation, anf short name is used during object creation to populate the named list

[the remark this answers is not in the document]
andlaus: this is what I mean with "footgun": if this was done this way, the `NamedList` would feature incorrect names because after reference resolution, the `short_name` attributes change whilst the list stays the  same. This is the reason why the `NamedItemList` of `ComparamInstance` objects is only [created after ODXLINK resolution](https://github.com/mercedes-benz/odxtools/blob/main/odxtools/diaglayers/hierarchyelement.py#L215-L222). My point is that is better to get an exception if the code is incorrect rather than incorrect results...

odxtools/diagvariable.py
andlaus: ok, I spinned it off: #387

andlaus:
@kayoub5: do you see any issues which should prevent merging this?

kayoub5:
@nada-ben-ali could you cross this PR against the files we have?

I have a feeling one of those many small changes that are standard compliant may not align with real world files.

nada-ben-ali:
> @nada-ben-ali could you cross this PR against the files we have?
> 
> I have a feeling one of those many small changes that are standard compliant may not align with real world files.

@kayoub5 the tests failed for three DIDs. I'm checking the root cause...

andlaus:
any news on this front @nada-ben-ali?

nada-ben-ali:
> any news on this front @nada-ben-ali?

I got the following error while decoding:

image

nada-ben-ali:
> > any news on this front @nada-ben-ali?
> 
> I got the following error while encoding:
> 
> image

I think it's:
`raw_value = internal_value`
instead of:
`internal_value = raw_value`

right?

kayoub5:
@nada-ben-ali what was the "base_type_encoding" used by the ODX file?

andlaus:
> @andlaus I think it's:
> raw_value = internal_value
> instead of:
> internal_value = raw_value
>
> right?

right. fixed: https://github.com/mercedes-benz/odxtools/pull/386/commits/3b807ebed7e8bc6d68bdbac209905f670fcfcd29

nada-ben-ali:
> > @andlaus I think it's:
> > raw_value = internal_value
> > instead of:
> > internal_value = raw_value
> > right?
> 
> right. fixed: [3b807eb](https://github.com/mercedes-benz/odxtools/commit/3b807ebed7e8bc6d68bdbac209905f670fcfcd29)

I think we need to fix it here too https://github.com/mercedes-benz/odxtools/blob/3b369b5d8e30477b86a150d4b3bfb28021191fa2/odxtools/encodestate.py#L160

andlaus:
oops, right. Fixed: https://github.com/mercedes-benz/odxtools/pull/386/commits/a8edd4bfb9fede1770925ab8cc2c5361e874eebe (I hope that I did not miss another instance of this...)

nada-ben-ali:
> these encoding issues are probably due to #381, not this PR!? Do you have any idea what actually goes wrong? IIRC, I added comprehensive unit tests for all base type encodings (cf https://github.com/mercedes-benz/odxtools/blob/main/tests/test_encoding.py#L93-L437), but possibly I made a few mistakes there...

yes it must be due to https://github.com/mercedes-benz/odxtools/pull/381 , we didn't get this issue using release 9.3.0.

andlaus:
> yes it must be due to https://github.com/mercedes-benz/odxtools/pull/381 , we didn't get this issue using release 9.3.0.

let's get it fixed ASAP...

andlaus:
I agree.

Maybe we can allow the BCD encodings for INT32. The spec disallows this, probably because BCD cannot represent negative values, but I have a feeling that negative values are rarely transmitted over the wire anyway...

UINT32 using two-complement encoding obviously does not make any sense because UINT32 obviously only works for positive numbers. The question is what ought to happen in such a case if a value where the most significant bit is set is encountered.

andlaus:
>  My feeling on this is that someone is not likely to put 2C as encoding by accident (since it's optional), but they may put UINT32 instead of INT32 by accident.

sounds reasonable. That said, what do the existing commercial tools do in that case and do all that claim to be compliant behave the same? Since this is undefined behaviour, I think odxtools should continue to raise a `DecodeError` in strict mode, while in non-strict mode we should probably always switch to "no encoding" if the base type is unsigned but the specified encoding is 1C or 2C. what do you think?

kayoub5:
> > My feeling on this is that someone is not likely to put 2C as encoding by accident (since it's optional), but they may put UINT32 instead of INT32 by accident.
> 
> sounds reasonable. That said, what do the existing commercial tools do in that case and do all that claim to be compliant behave the same? Since this is undefined behaviour, I think odxtools should continue to raise a `DecodeError` in strict mode, while in non-strict mode we should probably always switch to "no encoding" if the base type is unsigned but the specified encoding is 1C or 2C. what do you think?

that's ok as well

kayoub5 (approved):

PR #389, opened 2025-02-18
DataObjectProperty.decode_from_pdu: support returning the default value when the internal value is invalid

nada-ben-ali:
This merge request adds support for returning the default value when the internal value is invalid. If the internal value is not defined, the method will check for a defined default value and return it; otherwise, it will raise an exception.

odxtools version: 9.2.0

### DIFF
--- a/odxtools/dataobjectproperty.py
+++ b/odxtools/dataobjectproperty.py
@@ -143,7 +143,6 @@ class DataObjectProperty(DopBase):
         if default_value is not None:
             return str(default_value)
 
-        # TODO: How to prevent this?
         odxraise(f"DOP {self.short_name} could not convert the coded value "
                  f"{repr(internal)} to physical type {self.physical_type.base_data_type}.")
 
--- a/odxtools/dataobjectproperty.py
+++ b/odxtools/dataobjectproperty.py
@@ -137,15 +137,14 @@ class DataObjectProperty(DopBase):
         if self.compu_method.is_valid_internal_value(internal):
             return self.compu_method.convert_internal_to_physical(internal)
 
-        default_value = self.compu_method.compu_internal_to_phys.compu_default_value
+        default_value = getattr(self.compu_method.compu_internal_to_phys, "compu_default_value",
+                                None)
         if default_value is not None:
-            return default_value.value
+            return str(default_value)
 
         # TODO: How to prevent this?
-        raise DecodeError(
-            f"DOP {self.short_name} could not convert the coded value "
-            f"{repr(internal)} to physical type {self.physical_type.base_data_type}."
-        )
+        raise DecodeError(f"DOP {self.short_name} could not convert the coded value "
+                          f"{repr(internal)} to physical type {self.physical_type.base_data_type}.")
 
     def is_valid_physical_value(self, physical_value: ParameterValue) -> bool:
         return self.compu_method.is_valid_physical_value(cast(AtomicOdxType, physical_value))
--- a/odxtools/dataobjectproperty.py
+++ b/odxtools/dataobjectproperty.py
@@ -143,8 +143,10 @@ class DataObjectProperty(DopBase):
             return str(default_value)
 
         # TODO: How to prevent this?
-        raise DecodeError(f"DOP {self.short_name} could not convert the coded value "
+        odxraise(f"DOP {self.short_name} could not convert the coded value "
                           f"{repr(internal)} to physical type {self.physical_type.base_data_type}.")
+
+        return None
 
     def is_valid_physical_value(self, physical_value: ParameterValue) -> bool:
         return self.compu_method.is_valid_physical_value(cast(AtomicOdxType, physical_value))
--- a/odxtools/dataobjectproperty.py
+++ b/odxtools/dataobjectproperty.py
@@ -137,8 +137,9 @@ class DataObjectProperty(DopBase):
         if self.compu_method.is_valid_internal_value(internal):
             return self.compu_method.convert_internal_to_physical(internal)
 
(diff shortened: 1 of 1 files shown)


andlaus:
note: if you want to get rid of these pesky complaints about the coding style, just install yapf and isort via pip and run the [`reformatSource.sh`](https://github.com/mercedes-benz/odxtools/blob/main/reformat-source.sh) script in the top level directory of your working copy...

odxtools/dataobjectproperty.py
```diff
@@ -136,11 +136,16 @@ def decode_from_pdu(self, decode_state: DecodeState) -> ParameterValue:
...
+
+        default_value = self.compu_method.compu_internal_to_phys.compu_default_value
+        if default_value is not None:
+            return default_value.value
+
+        # TODO: How to prevent this?
+        raise DecodeError(
+            f"DOP {self.short_name} could not convert the coded value "
+            f"{repr(internal)} to physical type {self.physical_type.base_data_type}."
+        )
```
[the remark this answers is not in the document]
nada-ben-ali: @andlaus, @kayoub5 what do you think?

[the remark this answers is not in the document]
andlaus: ok, if the spec says so (do you happen to know the section?), let's use the variant you proposed but let's only raise the exception in strict mode, i.e., replace the `raise` statement by `odxraise()` followed by some non-strict mode fallback code.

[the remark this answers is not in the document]
nada-ben-ali: It's the section 7.3.6.6.7.
Ok, I'll update it.

odxtools/dataobjectproperty.py
```diff
@@ -136,11 +136,15 @@ def decode_from_pdu(self, decode_state: DecodeState) -> ParameterValue:
 
         if self.compu_method.is_valid_internal_value(internal):
             return self.compu_method.convert_internal_to_physical(internal)
-        else:
-            # TODO: How to prevent this?
-            raise DecodeError(
-                f"DOP {self.short_name} could not convert the coded value "
-                f" {repr(internal)} to physical type {self.physical_type.base_data_type}.")
+
+        default_value = getattr(self.compu_method.compu_internal_to_phys, "compu_default_value",
```
kayoub5, line 140: why getattr ?

odxtools/dataobjectproperty.py
```diff
@@ -136,11 +136,15 @@ def decode_from_pdu(self, decode_state: DecodeState) -> ParameterValue:
...
+                                None)
+        if default_value is not None:
+            return str(default_value)
+
+        # TODO: How to prevent this?
+        raise DecodeError(f"DOP {self.short_name} could not convert the coded value "
```
kayoub5, line 146: @andlaus what do you think about using `str(internal_value)` when not in strict mode as final fallback ?

andlaus: I don't really know: my hunch is that `None` is better because if the callling code encounters an unexpected type like `str` instead of `int` it almost certainly goes belly-up anyway and it is probably mentally a bit easier to associate `None` with "did not work". On the other hand, the internal value of the parameter would be lost in the warnings...

odxtools/dataobjectproperty.py
nada-ben-ali: because it's optional and could be None, the pipeline's linting checks reported it.

kayoub5:
```
internal_to_phys = self.compu_method.compu_internal_to_phys
default_value = internal_to_phys.compu_default_value if internal_to_phys else None
```

andlaus: That would be a perfect use case for Dart-style `None` aware operators in python (https://peps.python.org/pep-0505/). Unfortunately, the feature seems to have been stuck in a quackmire for the last decade...

(with such operators, the code would boil down to `self.compu_method.compu_internal_to_phys?.compu_default_value.value`)

odxtools/dataobjectproperty.py
```diff
@@ -136,11 +136,18 @@ def decode_from_pdu(self, decode_state: DecodeState) -> ParameterValue:
...
-            # TODO: How to prevent this?
-            raise DecodeError(
-                f"DOP {self.short_name} could not convert the coded value "
-                f" {repr(internal)} to physical type {self.physical_type.base_data_type}.")
+
+        internal_to_phys = self.compu_method.compu_internal_to_phys
+        default_value = internal_to_phys.compu_default_value if internal_to_phys else None
+
+        if default_value is not None:
+            return str(default_value)
```
andlaus, line 144: do you have a reference to the spec for this? It still seems fishy to me to fall back to a default value when receiving an undecodable value. (encoding a PDU where the parameter in question is left unspecified is a different story IMO.)

odxtools/dataobjectproperty.py
```diff
@@ -136,11 +136,18 @@ def decode_from_pdu(self, decode_state: DecodeState) -> ParameterValue:
...
+
+        # TODO: How to prevent this?
```
andlaus, line 146: that comment can be removed?

kayoub5 (approved):

odxtools/dataobjectproperty.py
nada-ben-ali: removed

odxtools/dataobjectproperty.py
andlaus: ok, thanks. That's a build-in footgun IMO, but...

andlaus (approved):
merge at will

nada-ben-ali:
@andlaus is it possible to get a new release today? Thank you!

PR #390, opened 2025-02-20
Small fix 

nada-ben-ali:
Fix for https://github.com/mercedes-benz/odxtools/pull/389
I initially used `str(default_value)`, but I didn't access its 'value' attribute. It should be `str(default_value.value)` .

### DIFF
--- a/odxtools/dataobjectproperty.py
+++ b/odxtools/dataobjectproperty.py
@@ -141,7 +141,7 @@ class DataObjectProperty(DopBase):
         default_value = internal_to_phys.compu_default_value if internal_to_phys else None
 
         if default_value is not None:
-            return str(default_value)
+            return str(default_value.value)
 
         odxraise(f"DOP {self.short_name} could not convert the coded value "
                  f"{repr(internal)} to physical type {self.physical_type.base_data_type}.")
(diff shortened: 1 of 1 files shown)


odxtools/dataobjectproperty.py
```diff
@@ -141,7 +141,7 @@ def decode_from_pdu(self, decode_state: DecodeState) -> ParameterValue:
         default_value = internal_to_phys.compu_default_value if internal_to_phys else None
 
         if default_value is not None:
-            return str(default_value)
+            return str(default_value.value)
```
kayoub5, line 144: Are you sure `str` needed?

nada-ben-ali: it was reported by the pipeline:
image

and as the default value is string, I added `str`.

kayoub5: Change the if condition to

`if default_value and default_value.value is not None:`

and remove `str`

the default value could be an int from what I see in the error

nada-ben-ali: fixed.

kayoub5 (approved):

andlaus (approved):

PR #391, opened 2025-02-20
improve the fallback code for some undefined encodings in non-strict mode

andlaus:
This emerged during the discussion for https://github.com/mercedes-benz/odxtools/pull/386: For A_INT32, we now handle BCD  encodings in non-strict mode (but use the absolute of the provided value), for A_UINT32 we now allow signed encodings in non-strict mode but simply pass through the raw number. Also, we now complain if the value that shall be encoded does not fit into the specified bit length.

### DIFF
--- a/odxtools/decodestate.py
+++ b/odxtools/decodestate.py
@@ -103,7 +103,9 @@ class DecodeState:
 
         # Deal with raw byte fields, ...
         if base_data_type == DataType.A_BYTEFIELD:
-            odxassert(base_type_encoding in (None, Encoding.NONE, Encoding.BCD_P, Encoding.BCD_UP))
+            odxassert(
+                base_type_encoding in (None, Encoding.NONE, Encoding.BCD_P, Encoding.BCD_UP),
+                f"Illegal encoding '{base_type_encoding}' for A_BYTEFIELD")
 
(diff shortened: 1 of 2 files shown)


odxtools/decodestate.py
```diff
@@ -103,7 +103,9 @@ def extract_atomic_value(
 
         # Deal with raw byte fields, ...
         if base_data_type == DataType.A_BYTEFIELD:
-            odxassert(base_type_encoding in (None, Encoding.NONE, Encoding.BCD_P, Encoding.BCD_UP))
+            odxassert(
+                base_type_encoding in (None, Encoding.NONE, Encoding.BCD_P, Encoding.BCD_UP),
+                f"Illegal encoding '{base_type_encoding}' for A_BYTEFIELD")
```
kayoub5, line 108: @andlaus BCD_P and BCD_UP are valid for A_BYTEFIELD

See table 8 of the specification, page 99

andlaus: yes, that's why we assert it. (the condition must be true for assertations ;))

kayoub5 (approved):

nada-ben-ali (approved):

PR #392, opened 2025-02-21
comparaminstance.spec property is not available during object creation

nada-ben-ali:
Although I tested the changes introduced in https://github.com/mercedes-benz/odxtools/pull/386, after upgrading to the latest odxtools version 9.4.0, we got the issue:

image

image

I believe not encountering the issue during testing was due to a Git branching issue :( .

### DIFF
--- a/odxtools/comparaminstance.py
+++ b/odxtools/comparaminstance.py
@@ -138,4 +138,9 @@ class ComparamInstance:
 
     @property
     def short_name(self) -> str:
-        return self.spec.short_name
+        if self.spec:
+            return self.spec.short_name
+
+        # ODXLINK IDs allow dots and hyphens, but short names do not.
+        # (This should not happen anyway in a correct PDX...)
+        return self.spec_ref.ref_id.replace(".", "__").replace("-", "_")


kayoub5 (approved):

andlaus:
can you provide a traceback? (does the exception happen in your own code or somewhere within odxtools.)

The point with this is that using `.short_name` before references have been resolved will almost certainly lead to incorrect results, and IMVHO it is better to get an issue explicitly flagged by an exception than it is to introduce subtle bugs in the code...

kayoub5:
> can you provide a traceback? (does the exception happen in your own code or somewhere within odxtools.)
> 
> The point with this is that using `.short_name` before references have been resolved will almost certainly lead to incorrect results, and IMVHO it is better to get an issue explicitly flagged by an exception than it is to introduce subtle bugs in the code...

The traceback is already included (screenshot)

This PR just reverts the change from https://github.com/mercedes-benz/odxtools/pull/386/files#r1952936788 to unblock the situation. you can do a follow up PR with a more proper fix.

andlaus:
> The traceback is already included (screenshot)

Yes, but I'm a bit confused: for me, loading a PDX file works fine using the current `main` version and the ecu_variants in these files all exhibit plenty comparams...

kayoub5:
@andlaus In https://github.com/mercedes-benz/odxtools/blob/1e5f030f8839647bdeff4fdebd06bb30ea6d3425/odxtools/database.py#L154

`dlc._finalize_init(self, self._odxlinks)` is called before `dlc._resolve_snrefs(context)`, thus at that point of time `spec` has not been resolved yet, `dlc._finalize_init` creates the NamedItemList here https://github.com/mercedes-benz/odxtools/blob/1e5f030f8839647bdeff4fdebd06bb30ea6d3425/odxtools/diaglayers/hierarchyelement.py#L224C9-L224C97

Creating the NamedItemList causes short_name getter to be called

andlaus:
I understand. What I don't get is why this is a problem: the spec_ref of the comparam_refs is exclusively resolved in [_resolve_odxlinks()](https://github.com/mercedes-benz/odxtools/blob/main/odxtools/comparaminstance.py#L66-L67) and [`._resolve_odxlinks()`](https://github.com/mercedes-benz/odxtools/blob/main/odxtools/database.py#L140-L141) is called before [`.finalize_init()`](https://github.com/mercedes-benz/odxtools/blob/main/odxtools/database.py#L153-L154). Also, why does it work fine for me but obviously not for @nada-ben-ali?

nada-ben-ali:
@andlaus the files that reported this issue have model version "2.0.1". For the files with model version "2.2.0," the issue is not reproducible.

andlaus:
okay, thanks for the info. can you dig a deeper into what causes this?

I still cannot explain it because regardless of the model version, the odxlink references should be resolved when this `NamedItemList` gets created. Also, I suspect that the names of the items exposed by the ComparamInstance NamedItemList do not correspond to the short names of their comparam specs... (maybe ´._resolve_odxlinks()` is not called at all?)

nada-ben-ali:
@andlaus the issue is that while resolving the odxlinks, the comparams were not resolved because:

 `doc_frag_db = self._db.get(ref_frag)` was None, which is because the `ref_frag ` object has doc_type=COMPARAM_SPEC.
However, the OdxDocFragment object with the same doc_name in `self._db` has doc_type=COMPARAM_SUBSET.

`Database.comparam_specs` is empty, while `Database.subsets` contains the required comparam specs, which were tagged as COMPARAM-SPEC in the PDX file.
In the PDX file, there is no COMPARAM-SUBSET tag, only the COMPARAM-SPEC tag exists.

image

andlaus:
@nada-ben-ali: that sounds a lot like an unrelated issue with the dataset which was uncovered by that change? what do you think about changing the fallback code for ODXLINK resolution if the reference was not found in the specified document fragment to "use the first document fragment which has an object with the specified ID"?

nada-ben-ali:
@kayoub @andlaus could you please review the lastest changes?

andlaus:
cool, now I'm happy! thanks for your patience :)

odxtools/comparamsubset.py
```diff
@@ -29,12 +29,16 @@ class ComparamSubset(OdxCategory):
     def from_et(et_element: ElementTree.Element,
                 doc_frags: List[OdxDocFragment]) -> "ComparamSubset":
 
-        cat = OdxCategory.category_from_et(et_element, doc_frags, doc_type=DocType.COMPARAM_SUBSET)
+        category = et_element.get("CATEGORY")
+
+        # In ODX 2.0, COMPARAM-SPEC is used, whereas in ODX 2.2, it refers to something else and has been replaced by COMPARAM-SUBSET.
+        # - If 'category' is missing (ODX 2.0), use COMPARAM_SPEC,
+        # - else (ODX 2.2), use COMPARAM_SUBSET.
+        doc_type = DocType.COMPARAM_SUBSET if category else DocType.COMPARAM_SPEC
```
andlaus, line 37: small question  (now irrelevant because the PR is already merged): isn't it possible to simply look at `et_element.tag`, i.e.
```python
doc_type = DocType.COMPARAM_SUBSET if et_element.tag == "COMPARAM_SUBSET" else DocType.COMPARAM_SPEC
```

kayoub5: possible

nada-ben-ali: yess, it works as well (the tag is "COMPARAM-SUBSET" instead of "COMPARAM_SUBSET").

andlaus: I supppose that would be less hacky, but I don't mind if you want to leave it as-is...

PR #394, opened 2025-02-25
Implement base variant patterns

andlaus:
Base variant patterns are almost identical to ECU variant patterns except that each base variant can specify at most one base variant pattern and that for some reason functional addressing can be specified to be used to communicate with the ECU which ought to be identified.

This PR first refactors the ECU variant matching code with an eye on generalizing it followed by adding the actual base variant matching functionality.


odxtools/ecuvariantmatcher.py
```diff
@@ -1,171 +1,32 @@
...
-from .matchingparameter import MatchingParameter
-from .odxtypes import ParameterValue
-from .response import Response
+from .variantmatcher import VariantMatcher
 
 
-class EcuVariantMatcher:
-    """EcuVariantMatcher implements the matching algorithm of ecu variants according to their
-    ECU-VARIANT-PATTERN according to ISO 22901-1.
+class EcuVariantMatcher(VariantMatcher):
```
dmholtz, line 8: Not sure whether this class is still needed, since all behavior is defined in the base class. The docstring is useful, though.

odxtools/basevariantmatcher.py
```diff
@@ -0,0 +1,35 @@
...
+        matcher.evaluate(resp)
+
+    # result
+    if matcher.has_match()
+        matching_bv = matcher.matching_variant
+        assert isinstance(matching_bv, BaseVariant)
+    ```
+    """
+
+    def __init__(self, base_variant_candidates: List[BaseVariant], use_cache: bool = True):
```
kayoub5, line 34: Really need the prefix?
```suggestion
    def __init__(self, candidates: List[BaseVariant], use_cache: bool = True):
```

odxtools/basevariantpattern.py
```diff
@@ -0,0 +1,38 @@
...
+
+@dataclass
+class BaseVariantPattern(VariantPattern):
+    """Base variant patterns are variant patterns used to identify the
+    base variant of an ECU.
+    """
+    matching_base_variant_parameters: List[MatchingBaseVariantParameter]
+
+    @override
+    def get_matching_parameters(
```
kayoub5, line 23: Why a getter? it takes no arguments

odxtools/diaglayers/ecuvariant.py
```diff
@@ -38,6 +39,14 @@ def dyn_defined_spec(self) -> Optional[DynDefinedSpec]:
     def parent_refs(self) -> List[ParentRef]:
         return self.ecu_variant_raw.parent_refs
 
+    @property
+    def base_variant(self) -> Optional[BaseVariant]:
+        for pr in self.ecu_variant_raw.parent_refs:
+            if isinstance(pr.layer, BaseVariant):
```
kayoub5, line 45: Could be multiple or not?

odxtools/ecuvariantmatcher.py
```diff
@@ -1,171 +1,32 @@
...
 
-    # initialize the matcher with a list of ecu variants,
-    # i.e., DiagLayer instances of variant_type == DiagLayerType.ECU-VARIANT
+    # initialize the matcher with a list of ecu variants
     matcher = EcuVariantMatcher(ecu_variant_candidates=[...], use_cache=use_cache)
 
     # run the request loop to obtain responses for every request
-    for req in matcher.request_loop():
-        resp = ... # make a UDS request
+    for _, encoded_request in matcher.request_loop():
```
kayoub5, line 20: Why return something if it's going to be discarded?

odxtools/basevariantpattern.py
andlaus: because `matching_parameters` conflicts with the `MATCHING-PARAMETERS` dataclass field of ECU variant patterns...

andlaus: (better solution appreciated)

odxtools/diaglayers/ecuvariant.py
andlaus: the spec only at most a single base variant parent, see checker rule 50 at the bottom. I've added a docstring. (sorry for the "drive-by change", BTW.)

odxtools/basevariantmatcher.py
```diff
@@ -0,0 +1,35 @@
+# SPDX-License-Identifier: MIT
+from typing import List
+
+from .diaglayers.basevariant import BaseVariant
+from .variantmatcher import VariantMatcher
+
+
+class BaseVariantMatcher(VariantMatcher):
```
kayoub5, line 8: What's the point of this class? it has no logic, VariantMatcher could be used directly

odxtools/ecuvariantmatcher.py
andlaus: because this allows to use the same code for ECU variant matching and base variant matching: for whatever reason, functional addressing can only be specified for the latter (oversight of the spec?)

odxtools/basevariantpattern.py
kayoub5: You already have `matching_base_variant_parameters`

odxtools/ecuvariantmatcher.py
kayoub5: I agree with @dmholtz you don't need a logicless wrapper for that

odxtools/basevariantmatcher.py
andlaus: fixed: https://github.com/mercedes-benz/odxtools/pull/394/commits/98d90caa4efd9f14603fad9175da279406da0b89

odxtools/ecuvariantmatcher.py
kayoub5: Where is the `isinstance` check that you are referring to?

andlaus: currently there's none, but if you want, I can modify the test... (as said below: this is exactly the same situation as with `ComplexDop`). Also, I think it is more intuitive to use `EcuVariantMatcher` if you want to do ECU variant matching...

kayoub5: so the class exists to cover a case that doesn't exist?

andlaus: ok, I budge about the subclasses issue: https://github.com/mercedes-benz/odxtools/pull/394/commits/d0b7481b26f64f0a172bd1a53e940202eacf191c .

for accessing the matching parameters, I'll stay firm on using a method (`.get_matching_parameters()`) instead of a property (`.matching_variant_parameters`) because reading the former IMO has less potential for confusion because nobody will try to look this up in the XML data...

kayoub5 (approved):

PR #395, opened 2025-02-25
fix the ODXLINK reference resolution fallback to ignore the DOCREF attributes

andlaus:
This is only relevant for invalid datasets which specify references with incorrect DOCREF and/or DOCTYPE attributes. This patch should make ODXLINK reference resolution work in non-strict mode for such datasets.

### DIFF
--- a/odxtools/odxlink.py
+++ b/odxtools/odxlink.py
@@ -199,6 +199,7 @@ class OdxLinkDatabase:
                     OdxWarning,
                     stacklevel=1,
                 )
+
                 continue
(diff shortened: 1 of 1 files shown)


kayoub5:
> cool, thanks. it is probably a good idea to prop up the reference resolution fallback code anyway, right?

I never saw an OEM file where that would be needed, only exception was the ODX 2.0 files, and that's fixed already

andlaus:
fair enough. let's close this...

PR #396, opened 2025-02-28
Fix SupportsBytes Behavior Change in Python 3.12

nada-ben-ali:
The `SupportsBytes` protocol only recognizes types that explicitly implement `__bytes__()`.
Python 3.12 implements `__bytes__()`, but Python 3.9 does not.

image  image

This change affects the behavior of `isinstance(obj, SupportsBytes)`, which previously returned `False` for bytes in Python 3.9 but now returns `True` in Python 3.12.

**### Impact `on standardlengthtype.__unapply_mask`:**
This change has impacted the behavior of the `standardlengthtype.__unapply_mask(self, raw_value: AtomicOdxType)` method.

- When `raw_value` is of type bytes, Python 3.9 raises an odxraise and returns raw_value as-is (in bytes) in non-strict mode.

- In Python 3.12, the condition at [this line](https://github.com/mercedes-benz/odxtools/blob/5cc2c95ab238d482fc1ffc3819d45814f007871d/odxtools/standardlengthtype.py#L173) evaluates to True, causing the method to return an integer instead of bytes.

- Returning an integer in this case is incorrect, leading to the following exception:

image

**### Fix Implemented in This Merge Request:**
To ensure compatibility across Python versions, the following changes have been made:

- Introduced `BytesTypes` in utils.py, which includes `(bytearray, bytes, SupportsBytes)`.

- Updated `standardlengthtype.__unapply_mask` to correctly return bytes when` isinstance(raw_value, BytesTypes) == True`.

odxtools==9.4.1

### DIFF
--- a/odxtools/dataobjectproperty.py
+++ b/odxtools/dataobjectproperty.py
@@ -17,7 +17,7 @@ from .odxtypes import AtomicOdxType, ParameterValue
 from .physicaltype import PhysicalType
 from .snrefcontext import SnRefContext
 from .unit import Unit
-from .utils import dataclass_fields_asdict
+from .utils import BytesTypes, dataclass_fields_asdict
 
 
 @dataclass
@@ -120,7 +120,7 @@ class DataObjectProperty(DopBase):
                 f"The value {repr(physical_value)} of type {type(physical_value).__name__}"
                 f" is not a valid.")
 
-        if not isinstance(physical_value, (int, float, str, bytes, bytearray)):
+        if not isinstance(physical_value, (int, float, str, BytesTypes)):
             odxraise(f"Invalid type '{type(physical_value).__name__}' for physical value. "
                      f"(Expect atomic type!)")
         internal_value = self.compu_method.convert_physical_to_internal(physical_value)
--- a/odxtools/utils.py
+++ b/odxtools/utils.py
@@ -3,10 +3,14 @@ import dataclasses
 import re
 from typing import TYPE_CHECKING, Any, Dict, Optional
 
+from typing_extensions import SupportsBytes
+
 if TYPE_CHECKING:
     from .database import Database
     from .diaglayers.diaglayer import DiagLayer
     from .snrefcontext import SnRefContext
+
+BytesTypes = (bytearray, bytes, SupportsBytes)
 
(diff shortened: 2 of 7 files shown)


odxtools/dataobjectproperty.py
```diff
@@ -120,7 +120,7 @@ def encode_into_pdu(self, physical_value: ParameterValue, encode_state: EncodeSt
                 f"The value {repr(physical_value)} of type {type(physical_value).__name__}"
                 f" is not a valid.")
 
-        if not isinstance(physical_value, (int, float, str, bytes, bytearray)):
+        if not isinstance(physical_value, (int, float, str, BytesTypes)):
```
kayoub5, line 123: Are you sure `isinstance` supports nested tuples?

kayoub5: It seems it does https://docs.python.org/3/library/functions.html#isinstance

kayoub5 (approved):

odxtools/dataobjectproperty.py
nada-ben-ali: yes it works:
image

odxtools/utils.py
```diff
@@ -3,11 +3,15 @@
...
 from typing import TYPE_CHECKING, Any, Dict, Optional
 
+from typing_extensions import SupportsBytes
+
 if TYPE_CHECKING:
     from .database import Database
     from .diaglayers.diaglayer import DiagLayer
     from .snrefcontext import SnRefContext
 
+BytesTypes = (bytearray, bytes, SupportsBytes)
```
andlaus, line 13: it's probably better to change this to
```suggestion
BytesTypes = Union[bytearray, bytes, SupportsBytes]
```

kayoub5: isinstance supports union starting from python 3.10

andlaus: do tuples work on 3.8?

kayoub5: yes, https://docs.python.org/3.8/library/functions.html#isinstance

andlaus: great! let's merge...

PR #398, opened 2025-03-14
Implement state machines

andlaus:
This implements the runtime functionality of state charts. Note that the specification strongly implies deterministic state machines, but does not explicitly forbid non-determinism and also does not specify what should happen if more than one state transition ought to be taken.

Also, IMO it is rather inconvenient that no values for state transition references which specify constant parameters can be specified because the parameters relevant for the state transitions are (probably, the spec is also unclear here) the ones of the response, and these can vary depending on the concrete data received. Finally, it is IMO rather limiting that value parameters can only be tested for equivalence, and not for scale-constraints...


examples/somersaultecu.py
```diff
@@ -2445,7 +2518,210 @@ class SomersaultSID(IntEnum):
...
     import_refs=[],
-    state_charts=NamedItemList(),
+    state_charts=NamedItemList([
+        StateChart(
+            semantic="annoyed",
+            odx_id=OdxLinkId("charts.annoyed.chart", doc_frags),
+            oid=None,
+            short_name="annoyed_chart",
+            long_name=None,
+            description=None,  #Description("State chart for a day where the ECU is grumpy",
```
andlaus, line 2528: be aware that this is due to some weirdness: if a description is added here, the resulting ODX-D file does not validate anymore although I'm 100% sure that the XSD includes an ELEMENT-ID group (and thus DESC) for STATE-CHART...

examples/somersaultecu.py
```diff
@@ -1518,6 +1523,19 @@ class SomersaultSID(IntEnum):
                     bit_position=None,
                     sdgs=[],
                 ),
+                ValueParameter(
+                    oid=None,
+                    short_name="bribe",
```
kayoub5, line 1528: Why the extra param?

examples/somersaultecu.py
kayoub5: Could be the order of elements in the XML

examples/somersaultecu.py
```diff
@@ -2445,7 +2518,210 @@ class SomersaultSID(IntEnum):
...
+                    short_name="in_bed",
+                    long_name=None,
+                    description=None,
+                ),
+                State(
+                    odx_id=OdxLinkId("charts.annoyed.states.on_street", doc_frags),
+                    oid=None,
+                    short_name="on_street",
+                    long_name=None,
+                    description=None,
```
kayoub5, line 2544: Looking at the python code, wouldn't it be easier and more sane to commit the XML file instead?

kayoub5:
Suggestion: commit XML files instead of PDX files, to make it easier for review for `examples/somersault_modified.pdx`

You can add a CI step that creates the zip pdx during tests

odxtools/preconditionstateref.py
```diff
@@ -0,0 +1,142 @@
...
+        Note that the specification is unclear about what the
+        parameters are: It says "The optional VALUE together with the
+        also optional IN-PARAM-IF snref at STATE-TRANSITION-REF and
+        PRE-CONDITION-STATE-REF can be used if the STATE-TRANSITIONs
+        and pre-condition STATEs are dependent on the values of the
+        referenced PARAMs.", but it does not specify what the
+        "referenced PARAMs" are. For the state transition refs, let's
+        assume that they are the parameters of a positive response
+        received from the ECU, whilst we assume that they are the
+        parameters of the request for PRE-CONDITION-STATE-REFs.
```
kayoub5, line 87: Is there a reason why not to include both request and response parameters?

odxtools/preconditionstateref.py
```diff
@@ -0,0 +1,142 @@
...
+
+        if self.in_param_if_snref is None and self.in_param_if_snpathref is None:
+            # no parameter given -> only the specified state is relevant
+            return True
+
+        param, param_value = \
+            _resolve_in_param(self.in_param_if_snref, self.in_param_if_snpathref, params, param_value_dict)
+
+        if param is None:
+            # The referenced parameter does not exist. TODO: is this an error?
```
kayoub5, line 103: No reason for it to be
```suggestion
            # The referenced parameter does not exist.
```

odxtools/preconditionstateref.py
```diff
@@ -0,0 +1,142 @@
...
+            (CodedConstParameter, PhysicalConstantParameter, TableKeyParameter, ValueParameter)):
+            # see checker rule 194 in section B.2 of the spec
+            odxraise(f"Parameter referenced by state transition ref is of "
+                     f"invalid type {type(param).__name__}")
+            return False
+        elif isinstance(param, (CodedConstParameter, PhysicalConstantParameter,
+                                TableKeyParameter)) and self.value is not None:
+            # see checker rule 193 in section B.2 of the spec. Why can
+            # no values for constant parameters be specified? (This
+            # seems to be rather inconvenient...)
```
kayoub5, line 116: Because that would mean that the state will always apply, or always won't apply, it's a logic constraint, no need to hardcode it in the code

odxtools/preconditionstateref.py
```diff
@@ -0,0 +1,142 @@
...
+            odxraise(f"No value may be specified for state transitions referencing "
+                     f"parameters of type {type(param).__name__}")
+            return False
+        elif isinstance(param, ValueParameter):
+            # see checker rule 193 in section B.2 of the spec
+            if self.value is None:
+                odxraise(f"An expected VALUE must be specified for preconditions "
```
kayoub5, line 123: Already checked in `__post_init__`

odxtools/preconditionstateref.py
```diff
@@ -0,0 +1,142 @@
...
+                return False
+
+            if param_value is None:
+                param_value = param.physical_default_value
+                if param_value is None:
+                    odxraise(f"No value can be determined for parameter {param.short_name}")
+                    return False
+
+            if param.dop is None or not isinstance(param.dop, DataObjectProperty):
+                odxraise(f"Parameter {param.short_name} uses a non-simple DOP")
```
kayoub5, line 134: Is there a spec clause about this?

odxtools/statemachine.py
```diff
@@ -0,0 +1,174 @@
...
+        # the service can be executed if any of the specified
+        # precondition states is fulfilled. (TODO: correct?)
+        if service.pre_condition_state_refs is not None:
+            if not any(
+                    x.applies(self, service.request.parameters, service_params)
+                    for x in service.pre_condition_state_refs):
+                # if all preconditions which are applicable are
+                # invalid (i.e., they evaluate to False), we must not
+                # execute the service.
+                self.succeeded = False
```
kayoub5, line 141: Why not raise?

odxtools/statemachine.py
```diff
@@ -0,0 +1,174 @@
...
+        else:
+            resp_object = service.positive_responses[0]
+            for stransref in service.state_transition_refs:
+                if stransref.execute(self, resp_object.parameters, raw_resp):
+                    self.succeeded = True
+                    yield b''
+                    return
+
+        self.succeeded = True
+        yield b''
```
kayoub5, line 173: Why yield empty bytes?

odxtools/statetransitionref.py
```diff
@@ -0,0 +1,219 @@
...
+        path_chunks = [in_param_if_snref]
+    elif in_param_if_snpathref is not None:
+        path_chunks = in_param_if_snpathref.split(".")
+    else:
+        return None, None
+
+    return _resolve_in_param_helper(params, param_dict, path_chunks)
+
+
+def _resolve_in_param_helper(
```
kayoub5, line 44: Isn't this info already in the encode state/ decode state ?

odxtools/statetransitionref.py
```diff
@@ -0,0 +1,219 @@
...
+    def _build_odxlinks(self) -> Dict[OdxLinkId, Any]:
+        return {}
+
+    def _resolve_odxlinks(self, odxlinks: OdxLinkDatabase) -> None:
+        self._state_transition = odxlinks.resolve(self, StateTransition)
+
+    def _resolve_snrefs(self, context: SnRefContext) -> None:
+        pass
+
+    def execute(self, state_machine: StateMachine, params: List[Parameter],
```
kayoub5, line 143: Looks like same code from `applies`

examples/somersaultecu.py
andlaus: I don't think so: The ELEMENT-ID subtags are written by the [`printElementIdSubtags()`](https://github.com/mercedes-benz/odxtools/blob/main/odxtools/templates/macros/printElementId.xml.jinja2#L13-L19) macro and it works fine for everything else. the error produced by xmllint also does not look like this is the case:
```
> xmllint --schema BaseStandard/Schema/odx.xsd --noout somersault.odx-d
somersault.odx-d:1220: Schemas validity error : Element 'DESC': Character content other than whitespace is not allowed because the content type is 'element-only'.
somersault.odx-d fails to validate
```

examples/somersaultecu.py
andlaus: because I wanted a state chart which has a pre condition state ref that depends on the value of a parameter. Doing it this way seemed to be a somewhat intuitive scenario to me...

examples/somersaultecu.py
andlaus: yeah. at least we could/should create some of these objects from XML some time in the future. That said, in about a month's time we will switch to python 3.10 and `kw-only` dataclasses so we only need to explicitly specify the non-optional attributes...

andlaus:
> Suggestion: commit XML files instead of PDX files, to make it easier for review for `examples/somersault_modified.pdx`

maybe we should remove these PDX files entirely as they are auto-generated anyway and their unzipped contents are so unwieldy that you probably don't want to review them either. If this is done, we need to ensure that they are generated before running any of the unit tests then, though...

odxtools/preconditionstateref.py
andlaus: yes, but for state transitions multiple responses could apply, so if they use a constant exhibiting the same name for e.g. their error code, they cannot be distinguished...

odxtools/preconditionstateref.py
```diff
@@ -0,0 +1,142 @@
...
+                return False
+
+            expected_value = param.dop.physical_type.base_data_type.from_string(self.value)
+
+            if expected_value != param_value:
+                return False
+
+        return True
```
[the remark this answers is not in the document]
andlaus: The spec disallows this (cf checker rule 194). I agree that this is overly strict (e.g., PHYS-CONST parameters should be allowed), but I've got the impression that the state-chart/state machine functionality as described by the spec is only half-baked...

odxtools/statemachine.py
andlaus: because it is IMO not really an error if the service cannot be run and it would complicate the calling code. (the for loop described in the docstring would need to be made aware of the exception.)

odxtools/statemachine.py
andlaus: I was surprised about this as well, but returning outright causes a `StopIteration` to be raised here. instead of terminating the loop after a single iteration. For whatever reason, the second yield does not get passed to the calling code, see the [unit test](https://github.com/mercedes-benz/odxtools/pull/398/files#diff-032f4e3b4207885cc9f9a556d079a94d23c497c353da952e672357c10b12de65R59-R66).

odxtools/statetransitionref.py
andlaus: not really. there already is a close analog in the [variant matching code](https://github.com/mercedes-benz/odxtools/blob/main/odxtools/matchingparameter.py#L80-L138), though. (the traversal of the pathref is basically identical, but what is done is different...)

odxtools/statetransitionref.py
andlaus: true. the bones are the same, but the meat attached to them is slightly different. I did not see a simple way to unify these two...

odxtools/preconditionstateref.py
andlaus: removing this here makes mypy trip:
```
odxtools/preconditionstateref.py:131: error: Argument 1 to "from_string" of "DataType" has incompatible type "str | None"; expected "str"  [arg-type]
Found 1 error in 1 file (checked 201 source files)
```

examples/somersaultecu.py
kayoub5: You need to wrap it in `<p></p>`

odxtools/preconditionstateref.py
kayoub5: Adding request won't hurt, extra params simply won't be used

examples/somersaultecu.py
andlaus: ah, I completely forgot. thanks the info :)

odxtools/preconditionstateref.py
andlaus: I agree, [changed](https://github.com/mercedes-benz/odxtools/pull/398/commits/85661e7b03e3a7fc44e4e34a0665da44c931ba15). (Note that this is a bit more complicated than it seems at first...)

odxtools/preconditionstateref.py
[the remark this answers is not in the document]
andlaus: modulo bugs it is: This means that all preconditions for executing the diag comm are satisfied: active state of state machine is correct plus the referenced parameter exists is of a valid type and exhibits the expected value...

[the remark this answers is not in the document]
andlaus: the question is if we want to risk raising an exception if we encounter a response that exhibits an illegal parameter type (because there might be ones with parameters that exhibit a valid type.) since the spec also does not mention such details, I propose to interpret it as "if the referenced parameter exists it must always exhibit one of the specified types".

odxtools/statemachine.py
```diff
@@ -145,6 +145,18 @@ def execute(self, service: "DiagService", **service_params: Any
         # ask the calling code to send the request to the ECU and
         # report back the reply
         raw_resp = yield raw_req
+
+        decoded_req_params = service.request.decode(raw_req)
```
kayoub5, line 149: Why decode request if you already have `service_params` ?

odxtools/odxlink.py
```diff
@@ -277,24 +277,40 @@ def update(self, new_entries: Dict[OdxLinkId, Any], *, overwrite: bool = True) -
...
+                  *,
+                  lenient: bool = True) -> Optional[TNamed]:
     ...
 
 
 def resolve_snref(target_short_name: str,
                   items: Iterable[OdxNamed],
-                  expected_type: Any = None) -> Any:
+                  expected_type: Any = None,
+                  lenient: Optional[bool] = None) -> Any:
```
kayoub5, line 308: Why lenient is `Optional[bool]` ?

odxtools/preconditionstateref.py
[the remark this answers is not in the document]
kayoub5: Then you need an extra else check on the parameter type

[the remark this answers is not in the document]
andlaus: I probably misunderstood what you mean here, but I don't think so: we also want to return `True` after we checked that the parameter exhibits the expected value. We could also return `True` for each individual `if`-branch, but IMO the filtering approach (only returning if there the state transition ought not to be taken) is more intuitive and it allows for more conditions to be added later.

odxtools/statetransitionref.py
kayoub5: No callbacks needed

` def get_next_state(self, state_machine: StateMachine, params,  param_value_dict) -> Optional[State]:`

odxtools/statemachine.py
andlaus: because optional parameters don't need to be specified for `service_params`. (this is a kludge to ensure that the same parameters as seen by the target ECU are used.)

odxtools/odxlink.py
andlaus: because I have not found a way to specialize a function signature depending on an argument's value, only for its type. (Note that this means that mypy uses the incorrect function signature if `resolve_snrefs(..., lenient=False)` is used, but IMO this is pretty much a corner case which also does not do any harm.)

odxtools/statemachine.py
kayoub5: but optional (constant) parameters are not allowed in transitions as well.

odxtools/statetransitionref.py
andlaus: for the precondition state references, there is no "next state", though.

odxtools/statemachine.py
andlaus: as far as I understood this, they are allowed. What's not allowed is specifying a value for them, but they can be tested for their presence. Also, value parameters can be defaulted and thus do not need to be specified for `encode()`. (Defaulted parameters that have not been specified for `encode()` could be handled here, but keeping this exactly equivalent to the encoding code is probably quite a headache which can be avoided by simply decoding the raw message, in particular since this code is not performance critical...)

odxtools/statetransitionref.py
andlaus: okay, I've unified that boiler plate code: https://github.com/mercedes-benz/odxtools/pull/398/commits/93e57cc0c3702911e6f4e634cdaab1b1b5ff61ed . do you see any further holdups for getting this merged?

kayoub5 (approved):

PR #399, opened 2025-03-25
implement the TEXT tag

andlaus:
This is a tag containing an ordinary string but the tag itself has an additional `TI` (text identifier) attribute. The only user is the `DTC` tag and it also seems to be very unusual to specify a text identifier there.

This is the last missing tag type that can be specified inside `DIAG-LAYER-CONTAINER`, `COMPARAM-SUBSET` or `COMPARAM-SPEC`, i.e., after this commit the diagnostics related functionality of the ODX standard should be fully covered by odxtools... (modulo bugs)

### DIFF
--- a/odxtools/text.py
+++ b/odxtools/text.py
@@ -1,0 +1,35 @@
+from dataclasses import dataclass
+from typing import List, Optional
+from xml.etree import ElementTree
+
+from .odxlink import OdxDocFragment
+
+
+@dataclass
+class Text:
+    text: str
+    text_identifier: Optional[str]
+
+    @staticmethod
+    def from_et(et_element: ElementTree.Element, doc_frags: List[OdxDocFragment]) -> "Text":
+        # Extract the contents of the tag as a string.
+        raw_string = et_element.text or ""
+        for e in et_element:
+            raw_string += ElementTree.tostring(e, encoding="unicode")
+
+        # remove white spaces at the beginning and at the end of all
+        # extracted lines
+        stripped_lines = [x.strip() for x in raw_string.split("\n")]
+
+        text = "\n".join(stripped_lines).strip()
+
+        text_identifier = et_element.get("TI")
+
+        return Text(text=text, text_identifier=text_identifier)
+
+    @staticmethod
+    def from_string(text: str) -> "Text":
+        return Text(text=text, text_identifier=None)
+
+    def __str__(self) -> str:
+        return self.text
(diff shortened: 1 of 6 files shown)


odxtools/text.py
```diff
@@ -0,0 +1,35 @@
...
+    text_identifier: Optional[str]
+
+    @staticmethod
+    def from_et(et_element: ElementTree.Element, doc_frags: List[OdxDocFragment]) -> "Text":
+        # Extract the contents of the tag as a string.
+        raw_string = et_element.text or ""
+        for e in et_element:
+            raw_string += ElementTree.tostring(e, encoding="unicode")
+
+        # remove white spaces at the beginning and at the end of all
```
kayoub5, line 20: Why?

odxtools/text.py
```diff
@@ -0,0 +1,35 @@
...
+from typing import List, Optional
+from xml.etree import ElementTree
+
+from .odxlink import OdxDocFragment
+
+
+@dataclass
+class Text:
+    text: str
+    text_identifier: Optional[str]
```
kayoub5, line 11:
```suggestion
   identifier: Optional[str]
```

odxtools/text.py
```diff
@@ -0,0 +1,35 @@
...
+
+        text = "\n".join(stripped_lines).strip()
+
+        text_identifier = et_element.get("TI")
+
+        return Text(text=text, text_identifier=text_identifier)
+
+    @staticmethod
+    def from_string(text: str) -> "Text":
+        return Text(text=text, text_identifier=None)
```
kayoub5, line 32: Feels like a useless wrapper, why not just directly call `Text()` ?

andlaus: because this requires `Text(text="foo", text_identifier=None)`, which is a bit less convenient. The main reason is to keep it consistent with `Description`, though. since the latter is used a ton in the somersault example, always having to explicitly specify all arguments is a bit clunky IMO. (That said, I don't know why the standard does not simply use the DESC type for DTC/TEXT...)

odxtools/text.py
andlaus: I don't mind changing this in principle, but the XML attribute is called "TI" and the spec says that it stands for "text identifier" (same for DESCRIPTION tags, BTW)...

odxtools/text.py
```diff
@@ -0,0 +1,35 @@
...
+
+@dataclass
+class Text:
+    text: str
+    text_identifier: Optional[str]
+
+    @staticmethod
+    def from_et(et_element: ElementTree.Element, doc_frags: List[OdxDocFragment]) -> "Text":
+        # Extract the contents of the tag as a string.
+        raw_string = et_element.text or ""
```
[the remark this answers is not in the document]
andlaus: a while ago, in the context of `Description`, I was surprised that this works, too: It turns out that [`ElementTree.tostring()`](https://docs.python.org/3/library/xml.etree.elementtree.html#xml.etree.ElementTree.tostring) includes the element's `.tail` attribute in its output and `.tail` contains the text following immediately after the tag...

PR #400, opened 2025-03-26
Use the same order of dataclass fields as the XSD uses for subtags

andlaus:
This patch firmly belongs into the "long and fairly boring" category: It aims to make the python codebase consistent with the XSD, eliminate a few discrepancies to the spec and some genereral janitorial work. The main changes are to change (almost) all dataclasses to a consistent shape:

1) dataclass fields in the same order as in the XSD
2) `@property` methods in the same order as in the XSD (if applicable)
3) the `from_et()` constructor, followed by `__post_init__()`
4) auxiliary initialization functions (`_build_odxlinks()`,
  `_resolve_odxlinks()`, `_resolve_snrefs()`)
5) everything else

Andreas Lauser &lt;andreas.lauser@mercedes-benz.com&gt;, on behalf of [MBition GmbH](https://mbitionio/).
[Provider Information](https://github.com/mercedes-benz/foss/blob/master/PROVIDER_INFORMATION.md)

PR #401, opened 2025-03-28
SnRefContext: Handle `ComparamSpec`s and `ComparamSubset`s like diaglayers

andlaus:
This is a spin-off PR of #400 proposed by [at]kayoub5: in principle, this allows to resolve SNREFs that are relative to the
`COMPARAM-{SPEC,SUBSET}`. There are currently no users for this, but since this is a pretty simple change and `COMPARAM-{SPEC,SUBSET}` are equivalent to the DIAG-LAYERs from a topological point of view, let's introduce this...

### DIFF
--- a/odxtools/comparamspec.py
+++ b/odxtools/comparamspec.py
@@ -51,7 +51,11 @@ class ComparamSpec(OdxCategory):
         super()._finalize_init(database, odxlinks)
 
     def _resolve_snrefs(self, context: SnRefContext) -> None:
+        context.comparam_spec = self
+
         super()._resolve_snrefs(context)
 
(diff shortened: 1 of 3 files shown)


kayoub5:
> There are currently no users for this

The property can be introduced when the users for it are actually introduced, no?

andlaus:
> The property can be introduced when the users for it are actually introduced, no?

in principle yes, bit I don't like the asymmetry between the different top-level objects that is currently present...

andlaus:
I slept a night over this and I now think that this is not a hill I want to die on (rather something like  a picknick). Closing.

PR #402, opened 2025-03-29
Fix tag name for PosResponseSuppressible.bit_mask

nada-ben-ali:
According to the specification, the bit mask property for `PosResponseSuppressible` is defined using the tag name **BIT-MASK**, not **BITMASK**.

odxtools version: 9.5.0

### DIFF
--- a/odxtools/diagservice.py
+++ b/odxtools/diagservice.py
@@ -53,7 +53,7 @@ class PosResponseSuppressible:
     def from_et(et_element: ElementTree.Element,
                 doc_frags: List[OdxDocFragment]) -> "PosResponseSuppressible":
 
-        bit_mask = int(odxrequire(et_element.findtext("BITMASK")))
+        bit_mask = int(odxrequire(et_element.findtext("BIT-MASK")))
 
         coded_const_snref = None
         if (cc_snref_elem := et_element.find("CODED-CONST-SNREF")) is not None:


andlaus:
oops. that's correct, thanks!

odxtools/diagservice.py
```diff
@@ -53,7 +53,7 @@ class PosResponseSuppressible:
     def from_et(et_element: ElementTree.Element,
                 doc_frags: List[OdxDocFragment]) -> "PosResponseSuppressible":
 
-        bit_mask = int(odxrequire(et_element.findtext("BITMASK")))
+        bit_mask = int(odxrequire(et_element.findtext("BIT-MASK")))
```
andlaus, line 56: I noticed that the BIT-MASK tag is `xsd:hexBinary`, so this should probably become
```suggestion
        bit_mask_str := odxrequire(et_element.findtext("BIT-MASK")))
        # The XSD uses the type xsd:hexBinary
        # xsd:hexBinary allows for leading/trailing whitespace, empty strings, and it only allows an even
        # number of hex digits, while some of the examples shown in the  ODX specification exhibit an
        # odd number of hex digits.
        # This causes a validation paradox, so we try to be flexible
        bit_mask_str = bit_mask_str.strip()
        if len(bit_mask_str):
            bit_mask = int(bit_mask_str, 16)
        else:
            bit_mask = 0
```

right? (this makes it consistent with `StandardLengthType`)

andlaus: I don't disagree, but I'd like to point out the fact that most lines above are comments...

PR #404, opened 2025-03-31
centralize parsing of `xsd:hexBinary` values

andlaus:
This is a follow-up of https://github.com/mercedes-benz/odxtools/pull/402 proposed by @kayoub5. Besides eliminating a few lines of copy-and-pasted code, a surprising number of things can go wrong here, so it makes sense to centralize this.

### DIFF
--- a/odxtools/utils.py
+++ b/odxtools/utils.py
@@ -2,11 +2,38 @@
 import dataclasses
 import re
 from typing import TYPE_CHECKING, Any, Dict, Optional
+from xml.etree import ElementTree
+
+from .exceptions import odxraise
 
 if TYPE_CHECKING:
     from .database import Database
     from .diaglayers.diaglayer import DiagLayer
     from .snrefcontext import SnRefContext
+
+
+def read_hex_binary(et_element: Optional[ElementTree.Element]) -> Optional[int]:
+    """Convert the contents of an xsd:hexBinary to a `bytes` object"""
+    if et_element is None:
+        return None
+
+    if (bytes_str := et_element.text) is None:
+        odxraise(f"Element {et_element.tag} has no content")
+        return None
+
+    # The XSD uses the type xsd:hexBinary and xsd:hexBinary allows for
+    # leading/trailing whitespace and empty strings whilst `int(x,
+    # 16)` raises an exception if one of these things happen.
+    bytes_str = bytes_str.strip()
+    if len(bytes_str) == 0:
+        return 0
+
+    try:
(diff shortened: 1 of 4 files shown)


odxtools/utils.py
```diff
@@ -2,13 +2,40 @@
...
     from .snrefcontext import SnRefContext
 
 
+def read_hex_binary(et_element: Optional[ElementTree.Element]) -> Optional[int]:
+    """Convert the contents of an xsd:hexBinary to a `bytes` object"""
+    if et_element is None:
+        return None
+
+    if (bytes_str := et_element.text) is None:
+        odxraise(f"Element {et_element.tag} has no content")
```
kayoub5, line 21: Empty text is valid, it means zero length byte array

odxtools/utils.py
```diff
@@ -2,13 +2,40 @@
...
+
+from .exceptions import odxraise
 
 if TYPE_CHECKING:
     from .database import Database
     from .diaglayers.diaglayer import DiagLayer
     from .snrefcontext import SnRefContext
 
 
+def read_hex_binary(et_element: Optional[ElementTree.Element]) -> Optional[int]:
```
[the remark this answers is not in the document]
andlaus: yes, but we always use it as a bitmask.

odxtools/utils.py
kayoub5: you can try it with an xsd validator, `<BIT-MASK />` is a valid `xsd:hexBinary`, most libraries don't differentiate inline termination from empty string, and consider them the same thing.

andlaus: good points. fixed, thanks: https://github.com/mercedes-benz/odxtools/pull/404/commits/866b5434e670fdc7623e0efc8e248461e80ddb8d

odxtools/utils.py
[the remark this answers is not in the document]
kayoub5: Nitpick, but using int, will strip any leading zeros when writing the hex binary back

Did you check that writing hex binary is correct?

andlaus:
I simplified and fixed the output code (the part for POS-RESPONSE-SUPPRESSABLE used `hex` which is not available in jinja, and even if it was it adds a `0x` prefix which makes the XML validator fail.)

@kayoub5: do you see anything which should be fixed before merging?

kayoub5:
An odd number of nibbles is not actually valid, this is enforced by XSD, the examples in the spec are actually wrong.

kayoub5 (approved):

PR #405, opened 2025-04-01
Pass model version down for version dependent parsing

zariiii9003:
To enable version dependent parsing and in hope of better ODX 2.1.0 support, i added the `version` parameter to each `from_et` method.

Sorry for the big diff, but it's the same simple change in every file.

### DIFF
--- a/examples/somersaultecu.py
+++ b/examples/somersaultecu.py
@@ -8,6 +8,8 @@ from itertools import chain
 from typing import Any
 from xml.etree import ElementTree
 
+from packaging.version import Version
+
 import odxtools.uds as uds
 from odxtools.additionalaudience import AdditionalAudience
 from odxtools.admindata import AdminData
@@ -44,6 +46,7 @@ from odxtools.exceptions import odxrequire
 from odxtools.functionalclass import FunctionalClass
 from odxtools.modification import Modification
 from odxtools.nameditemlist import NamedItemList
+from odxtools.odxdoccontext import OdxDocContext
 from odxtools.odxlink import DocType, OdxDocFragment, OdxLinkId, OdxLinkRef
 from odxtools.odxtypes import DataType
 from odxtools.parameters.codedconstparameter import CodedConstParameter
@@ -75,6 +78,8 @@ from odxtools.unitgroup import UnitGroup
 from odxtools.unitgroupcategory import UnitGroupCategory
 from odxtools.unitspec import UnitSpec
 from odxtools.xdoc import XDoc
+
+ODX_VERSION = Version("2.2.0")
 
 
 class SomersaultSID(IntEnum):
@@ -3020,14 +3025,14 @@ for odx_cs_filename in (
     odx_cs_root = ElementTree.parse(odx_cs_dir / odx_cs_filename).getroot()
     subset = odx_cs_root.find("COMPARAM-SUBSET")
     if subset is not None:
-        comparam_subsets.append(ComparamSubset.from_et(subset, []))
+        comparam_subsets.append(ComparamSubset.from_et(subset, OdxDocContext(ODX_VERSION, [])))
 
 comparam_specs = []
 for odx_c_filename in ("UDSOnCAN_CPS.odx-c",):
     odx_c_root = ElementTree.parse(odx_cs_dir / odx_c_filename).getroot()
     subset = odx_c_root.find("COMPARAM-SPEC")
     if subset is not None:
-        comparam_specs.append(ComparamSpec.from_et(subset, []))
+        comparam_specs.append(ComparamSpec.from_et(subset, OdxDocContext(ODX_VERSION, [])))
 
 # create a database object
 database = Database()
--- a/odxtools/additionalaudience.py
+++ b/odxtools/additionalaudience.py
@@ -4,7 +4,8 @@ from typing import Any
 from xml.etree import ElementTree
 
 from .element import IdentifiableElement
-from .odxlink import OdxDocFragment, OdxLinkDatabase, OdxLinkId
+from .odxdoccontext import OdxDocContext
+from .odxlink import OdxLinkDatabase, OdxLinkId
 from .snrefcontext import SnRefContext
 from .utils import dataclass_fields_asdict
 
@@ -16,9 +17,8 @@ class AdditionalAudience(IdentifiableElement):
     """
 
     @staticmethod
-    def from_et(et_element: ElementTree.Element,
-                doc_frags: list[OdxDocFragment]) -> "AdditionalAudience":
-        kwargs = dataclass_fields_asdict(IdentifiableElement.from_et(et_element, doc_frags))
+    def from_et(et_element: ElementTree.Element, context: OdxDocContext) -> "AdditionalAudience":
+        kwargs = dataclass_fields_asdict(IdentifiableElement.from_et(et_element, context))
 
         return AdditionalAudience(**kwargs)
 
--- a/odxtools/diaglayers/diaglayerraw.py
+++ b/odxtools/diaglayers/diaglayerraw.py
@@ -1,5 +1,4 @@
 # SPDX-License-Identifier: MIT
-from copy import copy
 from dataclasses import dataclass
 from typing import Any, cast
 from xml.etree import ElementTree
@@ -15,6 +14,7 @@ from ..exceptions import odxassert, odxraise, odxrequire
 from ..functionalclass import FunctionalClass
 from ..library import Library
 from ..nameditemlist import NamedItemList
+from ..odxdoccontext import OdxDocContext
 from ..odxlink import DocType, OdxDocFragment, OdxLinkDatabase, OdxLinkId, OdxLinkRef
 from ..request import Request
 from ..response import Response
@@ -70,7 +70,7 @@ class DiagLayerRaw(IdentifiableElement):
         return self._single_ecu_jobs
 
     @staticmethod
-    def from_et(et_element: ElementTree.Element, doc_frags: list[OdxDocFragment]) -> "DiagLayerRaw":
+    def from_et(et_element: ElementTree.Element, context: OdxDocContext) -> "DiagLayerRaw":
         try:
             variant_type = DiagLayerType(et_element.tag)
         except ValueError:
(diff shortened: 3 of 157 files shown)


andlaus (commented):
welcome! I'm not sure if this is the right approach: currently odx 2.0 and 2.2 are supported by simply accepting the superset of the two (plus some comments explaining the discrepancies to the 2.2 XSD). This is obviously not always possible, but so far it worked out well enough. did you encounter a feature where the ODX version needed to be distinguished but it cannot not be done so by looking at the XML?

If this is indeed required and we're uprooting the `.from_et()` constructor parameters anyway, it would possibly be a good idea to take the chance to switch the `.from_et()`s to a single `ParseContext` argument analogous to `SnRefContext` and `._resolve_snrefs()`. any opinions?

kayoub5:
Only a handful of cases require the version, those could get the version from `et_element.getroot()`, a helper function could be added that takes the `et_element` as input, and return the version.

You don't need to pass the version around in every single fucntion

andlaus:
> [...] those could get the version from et_element.getroot() [...]

I just checked this out because it would be nice. Unfortunately, the `.getroot()` method seems to only be available for `ElementTree.ElementTree` objects, not `ElementTree.Element`...

odxtools/additionalaudience.py
```diff
@@ -16,9 +19,10 @@ class AdditionalAudience(IdentifiableElement):
     """
 
     @staticmethod
-    def from_et(et_element: ElementTree.Element,
-                doc_frags: List[OdxDocFragment]) -> "AdditionalAudience":
-        kwargs = dataclass_fields_asdict(IdentifiableElement.from_et(et_element, doc_frags))
+    def from_et(et_element: ElementTree.Element, doc_frags: List[OdxDocFragment],
+                version: "Version") -> "AdditionalAudience":
```
kayoub5, line 23: To make it more future proof, and avoid adding new arguments each time, maybe provide a "context" object, that has version and doc_frags as properties

andlaus:
Thanks for fixing the conflicts due to #406 and changing to the context approach. I think code-wise this is now fine, but you have not yet told us which feature requires this. (i.e., for which feature the "superset approach" breaks down.)

zariiii9003:
I created a context NamedTuple, that is passed down with `version` and `doc_frags` attributes as suggested. 

I also changed another thing which may be wrong since my understanding of the code is very limited: `OdxLinkId` and `OdxLinkRef` now keep an immutable snapshot of the doc_frags. The tests are passing and my PDX file with version 2.0.1 fails with the same traceback, so i hope that does not break anything.

examples/somersaultecu.py
```diff
@@ -97,12 +101,12 @@ class SomersaultSID(IntEnum):
 dlc_short_name = "somersault"
 
 # document fragment for everything except the communication parameters
-doc_frags = [OdxDocFragment(dlc_short_name, DocType.CONTAINER)]
+doc_frags = (OdxDocFragment(dlc_short_name, DocType.CONTAINER),)
```
kayoub5, line 104: why change a list to tuple ?

odxtools/odxlink.py
```diff
@@ -45,7 +52,7 @@ class OdxLinkId:
 
     #: The name and type of the document fragment to which the
     #: `local_id` is relative to
-    doc_fragments: list[OdxDocFragment]
+    doc_fragments: tuple[OdxDocFragment, ...]
```
andlaus, line 55: is there any advantage of changing this to a tuple? I always mentally associate a tuple with a sequence of constant length. While it is true that the document fragments of an individual `OdxLinkId` can be considered constant, between different objects their size can vary...

zariiii9003:
> Thanks for fixing the conflicts due to #406 and changing to the context approach. I think code-wise this is now fine, but you have not yet told us which feature requires this. (i.e., for which feature the "superset approach" breaks down.)

My 2.0.1 file is failing to resolve a COMPARAM-SUBSET or COMPARAM-SPEC. i'm not sure. I think there's some version dependent  doctype detection in the `ComparamSubset` code which is not robust.

odxtools/odxlink.py
zariiii9003: I generally prefer immutable types, if mutability is not necessary. But as i said, that may be wrong 😄

odxtools/odxlink.py
```diff
@@ -28,6 +30,11 @@ class OdxDocFragment:
     doc_type: DocType
 
 
+class OdxDocContext(NamedTuple):
```
andlaus, line 33: I prefer to have this class in a separate `odxdoccontext.py` file. In my mind it is not directly connected to the ODXLINK reference mechanism...

odxtools/statetransitionref.py
```diff
@@ -108,13 +108,13 @@ def _check_applies(ref: Union["StateTransitionRef",
         return False
     elif not isinstance(
             param,
-            CodedConstParameter | PhysicalConstantParameter | TableKeyParameter | ValueParameter):
```
andlaus, line 111: weird: when I was doing #406, ruff changed this from a tuple to the union type. do you have any idea why (and also why it does not complain now)

zariiii9003: I added it to the ignore list in pyproject.toml.[ That rule is deprecated](https://docs.astral.sh/ruff/rules/non-pep604-isinstance/).

odxtools/odxlink.py
zariiii9003: done

odxtools/odxlink.py
andlaus: I suppose it is a matter of taste: manipulating these objects is kind of a pain IMO (c.f. for example [mksomersaultmodifiedpdx.py](https://github.com/mercedes-benz/odxtools/blob/main/examples/mksomersaultmodifiedpdx.py)), but I see your point from a conceptual perspective. That said, I think this change should factored out into a separate PR.

besides that shouldn't these classes be marked as `frozen=True` if you want to see their objects as immutable?

andlaus:
> My 2.0.1 file is failing to resolve a COMPARAM-SUBSET or COMPARAM-SPEC. i'm not sure. I think there's some version dependent doctype detection in the ComparamSubset code which is not robust.

yeah, lacking any 2.0 datassets, I could not test this part. The `ComparamSpec` / `ComparamSubset` stuff seems to have changed in fishy ways between the versions, see [here](https://github.com/mercedes-benz/odxtools/blob/main/odxtools/comparamsubset.py#L34-L43). Would be great if you could give this code a shake...

odxtools/statetransitionref.py
andlaus: cool; but could you move that into a separate PR?

odxtools/diaglayers/diaglayerraw.py
```diff
@@ -80,88 +80,83 @@ def from_et(et_element: ElementTree.Element, doc_frags: list[OdxDocFragment]) ->
         short_name = odxrequire(et_element.findtext("SHORT-NAME"))
 
         # extend the applicable ODX "document fragments" for the diag layer objects
-        doc_frags = copy(doc_frags)
```
zariiii9003, line 83: Before i undo the change from list to tuple: is there a reason, why this the the only place where a copy of the doc_frags list is created? Can i just reuse the same context.doc_frags as everywhere else?

andlaus: if you make `.doc_frags` a tuple in the `OdxLinkId` and `OdxLinkRef` objects, copying it is not necessary here. in the current version, due to python using references, the `.doc_frag` attributes changing the original list would automagically change all these attributes if no new object was created here...

zariiii9003: Yes, i understand that, but my question was: is the copying necessary here? The other classes do not copy.

andlaus: I think I only introduced this after something broke down. maybe this simply did not happen for the other classes which introduce a new document fragment. (if this is the case, this is a bug. Possibly it is also the reason for the issues which you've encountered recently.)

I think the list-to-tuple (and maybe freezing the Id and Ref classes) should be factored out into a separate PR and be discussed there, besides this the present PR is mergable IMO...

zariiii9003: I rebased onto main and removed the tuple change

andlaus:
great, let's merge.

please open a PR for the tuple change, you may be on to something there: at least, there seem to be a few missing  `copy()` statements for the list of applicable document fragments in the parsing code...

andlaus:
hm: I just noticed that this caused a _massive_ performance regression: `odxtools list --dump-database` now takes 50% longer for one of my fairly large files (70 vs 47 seconds). @zariiii9003: do you have an idea what causes this?

andlaus:
I noticed that changing `OdxDocContext` from a `NamedTuple` to a dataclass recovers some of the performance (it goes from 70 to 60 seconds. odxtools 9.7.0 only needs 47 seconds, though...)

andlaus:
all of this is using python 3.13.2:

parsing benchmark script:
```python
import sys
import time
import odxtools

t0 = time.time()
db = odxtools.load_file(sys.argv[1])
t1 = time.time()

print(f"{t1 - t0}")
```



latest main branch: 
```
main > time python3 bench.py $MY_PDX_FILE 
41.90346693992615

real    0m47.000s
user    0m45.240s
sys     0m1.752s
|main > time odxtools list --dump-database $MY_PDX_FILE > /dev/null 

real    1m7.570s
user    1m3.563s
sys     0m3.997s
```

#410:
```
|PR410 > time python3 bench.py $MY_PDX_FILE 
41.244316816329956

real    0m46.330s
user    0m44.553s
sys     0m1.771s
|PR410 > time odxtools list --dump-database $MY_PDX_FILE > /dev/null 

real    0m57.097s
user    0m54.050s
sys     0m3.040s
```

9.6.1:
```
|828a4ad... > time python3 bench.py $MY_PDX_FILE 
41.82666873931885

real    0m47.056s
user    0m45.140s
sys     0m1.910s
|828a4ad... > time odxtools list --dump-database $MY_PDX_FILE > /dev/null 

real    0m57.377s
user    0m54.299s
sys     0m3.070s
```

I'm a bit confused: I'm sure that the performance difference was much larger yesterday and also very reproducible (also, my smoketesting script still shows the performance regression on `main`, but luckily it is gone for #410). That said, for the results above I changed the CPU scaling governor to `performance` which I didn't do yesterday. Also, #410 now seems to fix the issue...

PR #410, opened 2025-04-04
Convert OdxLinkId.doc_fragments and OdxLinkRef.ref_docs to tuple

zariiii9003:
As suggested in #405.

I tried converting OdxLinkRef into a frozen dataclass, but subclasses assign attributes after instantiation.

### DIFF
--- a/odxtools/comparamsubset.py
+++ b/odxtools/comparamsubset.py
@@ -9,7 +9,7 @@ from .dataobjectproperty import DataObjectProperty
 from .nameditemlist import NamedItemList
 from .odxcategory import OdxCategory
 from .odxdoccontext import OdxDocContext
-from .odxlink import DocType, OdxLinkDatabase, OdxLinkId
+from .odxlink import OdxLinkDatabase, OdxLinkId
 from .snrefcontext import SnRefContext
 from .unitspec import UnitSpec
 from .utils import dataclass_fields_asdict
@@ -31,14 +31,7 @@ class ComparamSubset(OdxCategory):
 
         category_attrib = et_element.attrib.get("CATEGORY")
 
-        # In ODX 2.0, COMPARAM-SPEC is used, whereas in ODX 2.2, it
-        # refers to something else and has been replaced by
-        # COMPARAM-SUBSET.
-        # - If the 'CATEGORY' attribute is missing (ODX 2.0), use
-        #   COMPARAM_SPEC,
-        # - else (ODX 2.2), use COMPARAM_SUBSET.
-        doc_type = DocType.COMPARAM_SUBSET if category_attrib is not None else DocType.COMPARAM_SPEC
-        base_obj = OdxCategory.category_from_et(et_element, context, doc_type=doc_type)
+        base_obj = OdxCategory.from_et(et_element, context)
         kwargs = dataclass_fields_asdict(base_obj)
 
         comparams = NamedItemList(
--- a/odxtools/database.py
+++ b/odxtools/database.py
@@ -91,29 +91,29 @@ class Database:
 
         category_et = child_elements[0]
         category_sn = odxrequire(category_et.findtext("SHORT-NAME"))
+        category_tag = category_et.tag
 
-        match category_et.tag:
-            case "DIAG-LAYER-CONTAINER":
-                context = OdxDocContext(model_version,
-                                        (OdxDocFragment(category_sn, DocType.CONTAINER),))
-                self._diag_layer_containers.append(DiagLayerContainer.from_et(category_et, context))
-            case "COMPARAM-SUBSET":
+        if category_tag == "DIAG-LAYER-CONTAINER":
+            context = OdxDocContext(model_version,
+                                    (OdxDocFragment(category_sn, DocType.CONTAINER),))
+            self._diag_layer_containers.append(DiagLayerContainer.from_et(category_et, context))
+        elif category_tag == "COMPARAM-SUBSET":
+            context = OdxDocContext(model_version,
+                                    (OdxDocFragment(category_sn, DocType.COMPARAM_SUBSET),))
+            self._comparam_subsets.append(ComparamSubset.from_et(category_et, context))
+        elif category_tag == "COMPARAM-SPEC":
+            # In ODX 2.0 there was only COMPARAM-SPEC. In ODX 2.2 the
+            # content of COMPARAM-SPEC was moved to COMPARAM-SUBSET
+            # and COMPARAM-SPEC became a container for PROT-STACKS and
+            # a PROT-STACK references a list of COMPARAM-SUBSET
+            if model_version < Version("2.2"):
                 context = OdxDocContext(model_version,
                                         (OdxDocFragment(category_sn, DocType.COMPARAM_SUBSET),))
                 self._comparam_subsets.append(ComparamSubset.from_et(category_et, context))
-            case "COMPARAM-SPEC":
-                # In ODX 2.0 there was only COMPARAM-SPEC. In ODX 2.2 the
-                # content of COMPARAM-SPEC was moved to COMPARAM-SUBSET
-                # and COMPARAM-SPEC became a container for PROT-STACKS and
-                # a PROT-STACK references a list of COMPARAM-SUBSET
-                if model_version < Version("2.2"):
-                    context = OdxDocContext(model_version,
-                                            (OdxDocFragment(category_sn, DocType.COMPARAM_SUBSET),))
-                    self._comparam_subsets.append(ComparamSubset.from_et(category_et, context))
-                else:
-                    context = OdxDocContext(model_version,
-                                            (OdxDocFragment(category_sn, DocType.COMPARAM_SPEC),))
-                    self._comparam_specs.append(ComparamSpec.from_et(category_et, context))
+            else:
+                context = OdxDocContext(model_version,
+                                        (OdxDocFragment(category_sn, DocType.COMPARAM_SPEC),))
+                self._comparam_specs.append(ComparamSpec.from_et(category_et, context))
 
     def refresh(self) -> None:
         # Create wrapper objects
--- a/odxtools/database.py
+++ b/odxtools/database.py
@@ -21,7 +21,7 @@ from .diaglayers.protocol import Protocol
 from .exceptions import odxraise, odxrequire
 from .nameditemlist import NamedItemList
 from .odxdoccontext import OdxDocContext
-from .odxlink import OdxLinkDatabase, OdxLinkId
+from .odxlink import DocType, OdxDocFragment, OdxLinkDatabase, OdxLinkId
 from .snrefcontext import SnRefContext
 
 
@@ -77,10 +77,6 @@ class Database:
         self.auxiliary_files[str(aux_file_name)] = aux_file_obj
 
     def _process_xml_tree(self, root: ElementTree.Element) -> None:
-        dlcs: list[DiagLayerContainer] = []
-        comparam_subsets: list[ComparamSubset] = []
-        comparam_specs: list[ComparamSpec] = []
-
         # ODX spec version
         model_version = Version(root.attrib.get("MODEL-VERSION", "2.0"))
         if self.model_version is not None and self.model_version != model_version:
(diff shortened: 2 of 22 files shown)


zariiii9003:
@andlaus Unrelated: i think there are a lot of mypy issues hidden by the `Noreturn` in the `odxraise` method. It should be `NoReturn | None` and the `TYPE_CHECKING` should be removed.

andlaus:
The `Noreturn` is on purpose: Semantically `odxraise()` should be considered identical to `raise`. The difference is that in non-strict mode it becomes a no-op, i.e. the reason why there is often code after the `odxraise()` call is to provide a sensible fallback for non-strict mode (which does not make any guarantees about undefined behavior or working in general, just that it "does its best"). So `odxraise()` should stay `Noreturn` in the main branch, but for development purposes, it I agree that could be temporarily changed to `None` to uncover the issues which you mentioned (I suppose that you will get a lot of mypy errors then...)

andlaus:
> I tried converting OdxLinkRef into a frozen dataclass, but subclasses assign attributes after instantiation.

IIRC, that was also the issue I stumbled over, `ParentRef` was particularly bad because it does quite a bit of reference resolution...

zariiii9003:
A few more changes after a bit of research. To my understand there are two options: the doc fragments are either just the category, or the category+diag_layer.

i changed the type annotation of OdxDocContext.doc_fragments to reflect that. The category context is now completely defined in `Database._process_xml_tree()`. The DiagLayerContainer then creates a new context for its layers.

odxtools/diaglayercontainer.py
```diff
@@ -43,29 +44,35 @@ def ecus(self) -> NamedItemList[EcuVariant]:
...
-        ])
-        base_variants = NamedItemList([
-            BaseVariant.from_et(dl_element, context)
-            for dl_element in et_element.iterfind("BASE-VARIANTS/BASE-VARIANT")
-        ])
-        ecu_variants = NamedItemList([
-            EcuVariant.from_et(dl_element, context)
-            for dl_element in et_element.iterfind("ECU-VARIANTS/ECU-VARIANT")
-        ])
+        def _get_layer_context(_layer_et: ElementTree.Element) -> OdxDocContext:
```
kayoub5, line 50:
```suggestion
        def _get_layer_context(layer_et: ElementTree.Element) -> OdxDocContext:
```

odxtools/diaglayercontainer.py
```diff
@@ -43,29 +44,35 @@ def ecus(self) -> NamedItemList[EcuVariant]:
...
+            layer_sn = odxrequire(_layer_et.findtext("SHORT-NAME"))
+            layer_docfrag = OdxDocFragment(layer_sn, DocType.LAYER)
+            # add layer doc fragment to container doc fragment
+            return OdxDocContext(context.version, (context.doc_fragments[0], layer_docfrag))
```
kayoub5, line 54: why 0?

odxtools/diaglayercontainer.py
```diff
@@ -43,29 +44,35 @@ def ecus(self) -> NamedItemList[EcuVariant]:
...
+
+        protocols = NamedItemList[Protocol]()
+        for layer_et in et_element.iterfind("PROTOCOLS/PROTOCOL"):
+            protocols.append(Protocol.from_et(layer_et, _get_layer_context(layer_et)))
```
kayoub5, line 58: @andlaus does NamedItemList handle appending and internal name lookup update correctly ?

odxtools/database.py
```diff
@@ -89,31 +85,35 @@ def _process_xml_tree(self, root: ElementTree.Element) -> None:
...
-        self._comparam_subsets.extend(comparam_subsets)
-        self._comparam_specs.extend(comparam_specs)
+        child_elements = list(root)
+        if len(child_elements) != 1:
+            odxraise("Each ODX document must contain exactly one category.")
+
+        category_et = child_elements[0]
+        category_sn = odxrequire(category_et.findtext("SHORT-NAME"))
+
+        match category_et.tag:
```
andlaus, line 95: the `match` statement is for "structural pattern matching" ([pep](https://peps.python.org/pep-0622/), I guess I'm too stupid to understand what that is), and AFAIK it is not more performant than a simple `if ... elif` chain. Please replace...

odxtools/diaglayers/diaglayerraw.py
```diff
@@ -77,10 +77,6 @@ def from_et(et_element: ElementTree.Element, context: OdxDocContext) -> "DiagLay
             variant_type = cast(DiagLayerType, None)
             odxraise(f"Encountered unknown diagnostic layer type '{et_element.tag}'")
 
-        short_name = odxrequire(et_element.findtext("SHORT-NAME"))
-
-        # extend the applicable ODX "document fragments" for the diag layer objects
-        context.doc_fragments.append(OdxDocFragment(short_name, DocType.LAYER))
```
kayoub5, line 83: what happened to this logic ?

odxtools/odxdoccontext.py
```diff
@@ -8,4 +8,6 @@
 
 class OdxDocContext(NamedTuple):
     version: Version
-    doc_fragments: list["OdxDocFragment"]
+
+    # the doc_fragments are either tuple(category,) or tuple(category, diag_layer)
+    doc_fragments: tuple["OdxDocFragment"] | tuple["OdxDocFragment", "OdxDocFragment"]
```
kayoub5, line 13: is it just me or comment and type hint don't match ?

tests/test_variant_patterns.py
```diff
@@ -10,7 +10,7 @@
 from odxtools.odxdoccontext import OdxDocContext
 from odxtools.odxlink import DocType, OdxDocFragment
 
-doc_frags = [OdxDocFragment(doc_name="pytest", doc_type=DocType.CONTAINER)]
+doc_frags = (OdxDocFragment(doc_name="pytest", doc_type=DocType.CONTAINER),)
```
andlaus, line 13: this is not an issue introduced by this PR, but I think that in practice all IDs that are inside a container also contain a document fragment for the layer which they are contained in? (this does not really matter here, though...)

andlaus (commented):
looks pretty good. when the `match` statement is replaced and it is verified that ODX 2.0 does not regress (you might need to help out on that front @kayoub5) I think this should be merged...

odxtools/odxdoccontext.py
andlaus: As I see it, the comment is a bit sloppy. it should probably be something like
```
    # the doc_fragments are either tuple(doc_frag(category),) or tuple(doc_frag(category), doc_frag(diag_layer))
```

odxtools/diaglayercontainer.py
andlaus: this code seems fine to me: the restriction is that after an item has been added to a `NamedItemList`, its `.short_name` attribute must not chang anymore or else accessing the item by its name (`foo_list.bar`) will not be consistent with its `.short_name` anymore...

odxtools/diaglayercontainer.py
```diff
@@ -43,29 +44,35 @@ def ecus(self) -> NamedItemList[EcuVariant]:
...
         kwargs = dataclass_fields_asdict(cat)
 
-        protocols = NamedItemList([
-            Protocol.from_et(dl_element, context)
-            for dl_element in et_element.iterfind("PROTOCOLS/PROTOCOL")
-        ])
-        functional_groups = NamedItemList([
-            FunctionalGroup.from_et(dl_element, context)
-            for dl_element in et_element.iterfind("FUNCTIONAL-GROUPS/FUNCTIONAL-GROUP")
-        ])
```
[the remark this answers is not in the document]
andlaus: to me it is still fine; the title of the PR became a bit outdated and should now be something like "refactor document fragment handling" ;)

odxtools/comparamsubset.py
```diff
@@ -31,14 +31,7 @@ def from_et(et_element: ElementTree.Element, context: OdxDocContext) -> "Compara
...
 
-        # In ODX 2.0, COMPARAM-SPEC is used, whereas in ODX 2.2, it
-        # refers to something else and has been replaced by
-        # COMPARAM-SUBSET.
-        # - If the 'CATEGORY' attribute is missing (ODX 2.0), use
-        #   COMPARAM_SPEC,
-        # - else (ODX 2.2), use COMPARAM_SUBSET.
-        doc_type = DocType.COMPARAM_SUBSET if category_attrib is not None else DocType.COMPARAM_SPEC
-        base_obj = OdxCategory.category_from_et(et_element, context, doc_type=doc_type)
+        base_obj = OdxCategory.from_et(et_element, context)
```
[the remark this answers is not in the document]
zariiii9003: The behavior should be unchanged. The doctype is now set within `Database._process_xml_tree()`.

odxtools/diaglayercontainer.py
zariiii9003: I added the underscore to avoid name collision with `layer_et` from outer scope. I could change it to something else though if needed.

odxtools/diaglayercontainer.py
zariiii9003: The `doc_fragments` tuple only contains one entry at this point (the diag layer container). So i take this first element and create a new tuple with the layer doc fragment.

odxtools/diaglayercontainer.py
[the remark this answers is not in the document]
zariiii9003: It's necessary, because a new context is created for each diag layer instantiation. The new context contains both DLC and LAYER  fragments.

odxtools/diaglayers/diaglayerraw.py
zariiii9003: The LAYER doc fragment is now created in DiagLayerContainer.

tests/test_variant_patterns.py
zariiii9003: Attributes of the ODX category like company data could have just the container though, right?

odxtools/odxdoccontext.py
zariiii9003: I pushed your comment

odxtools/diaglayercontainer.py
kayoub5: underscore is usually used to mark class members as private

variable names should not defer by just underscore between scoops

kayoub5:
@nada-ben-ali could you check that odx 2.0 compatibility is not broken with those changes ?

odxtools/diaglayers/hierarchyelement.py
```diff
@@ -31,7 +31,7 @@
 from .hierarchyelementraw import HierarchyElementRaw
 
 if TYPE_CHECKING:
-    from .database import Database
+    from ..database import Database
```
kayoub5, line 34: mypy didn't catch this ?

tests/test_variant_patterns.py
andlaus: right.

odxtools/diaglayers/hierarchyelement.py
andlaus: @zariiii9003: how did you find this? if using tools, can/should we change the CI system to do the same?

zariiii9003: I noticed red squiggly lines in my IDE. I also wonder why both ruff and mypy seem to ignore it

andlaus: ok. Seems like you should continue to let your IDE look at odxtools. mine (emacs) is not sophisticated enough ;)

andlaus:
sorry for the conflict due to #411. as I see it, the `.doc_frags` attribute should not be defaulted anyway...

andlaus:
thanks for rebasing. as soon as @nada-ben-ali reports back, I'll merge this.

andlaus:
For some strange reason, I ran into some DOCREF-related issue with this. I don't know why it only crept up with this PR, but I made a "meta pull request" for your PR here: https://github.com/zariiii9003/odxtools/pull/1 . would be great if you merged it...

andlaus:
@nada-ben-ali, @kayoub5: If I do not get any objections to this PR, I'll merge it on Thursday...

nada-ben-ali:
@zariiii9003 @kayoub5 @andlaus sorryy I've just seen the email notifications. I'll begin checking now.

nada-ben-ali:
> I got the following exception while loading the pickle file for some pdx files we have: image
> 
> I'm checking this...

the issue is caused by https://github.com/mercedes-benz/odxtools/pull/407.
@kayoub5 @andlaus after removing the changes introduced in https://github.com/mercedes-benz/odxtools/pull/407 and https://github.com/mercedes-benz/odxtools/pull/411/ (which have not been included in a release yet), no issues are observed with the current merge request changes.

I'm currently checking the impact of the dataclass as `Kw_only` on the unpickling operation and how we can fix it. It only occurs with some files, not all of them. For now, I have only identified the MRs that caused the issue.

andlaus:
ok, does this mean that this PR can be merged and I should only prepare the next release once the unpickling issues are fixed? are you on it? (I do not currently use pickling, so I have a hard time reproducing it.)

nada-ben-ali:
> ok, does this mean that this PR can be merged and I should only prepare the next release once the unpickling issues are fixed? are you on it? (I do not currently use pickling, so I have a hard time reproducing it.)

@andlaus yes, we can merge this, and I'll work on fixing the unpickling problem.

andlaus:
ok, cool. merging; thanks everyone!

PR #411, opened 2025-04-07
Specify defaults for all optional dataclass fields

andlaus:
This is another PR that firmly belongs into the "long and boring" category: It specifies default values for all optional fields of dataclasses, i.e. fields that can be `None` or are a list. (We can do this now, because all these classes are now `kw_only`.) In the interest of keeping the scope of this PR as small as possible, removing the parameter value specifications of defaulted fields when such a class is instantiated in the examples and tests will be done in a follow-up PR...

### DIFF
--- a/odxtools/basevariantpattern.py
+++ b/odxtools/basevariantpattern.py
@@ -1,5 +1,5 @@
 # SPDX-License-Identifier: MIT
-from dataclasses import dataclass
+from dataclasses import dataclass, field
 from xml.etree import ElementTree
 
 from typing_extensions import override
@@ -16,7 +16,8 @@ class BaseVariantPattern(VariantPattern):
     """Base variant patterns are variant patterns used to identify the
     base variant of an ECU.
     """
-    matching_base_variant_parameters: list[MatchingBaseVariantParameter]
+    matching_base_variant_parameters: list[MatchingBaseVariantParameter] = field(
+        default_factory=list)
 
     @staticmethod
     def from_et(et_element: ElementTree.Element, context: OdxDocContext) -> "BaseVariantPattern":
--- a/odxtools/comparamsubset.py
+++ b/odxtools/comparamsubset.py
@@ -23,7 +23,7 @@ class ComparamSubset(OdxCategory):
     comparams: NamedItemList[Comparam]
     complex_comparams: NamedItemList[ComplexComparam]
     data_object_props: NamedItemList[DataObjectProperty]
-    unit_spec: UnitSpec | None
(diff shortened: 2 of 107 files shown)


odxtools/basevariantpattern.py
```diff
@@ -16,7 +16,8 @@ class BaseVariantPattern(VariantPattern):
     """Base variant patterns are variant patterns used to identify the
     base variant of an ECU.
     """
-    matching_base_variant_parameters: list[MatchingBaseVariantParameter]
+    matching_base_variant_parameters: list[MatchingBaseVariantParameter] = field(
```
kayoub5, line 19: Not a `NamedItemList` ?

odxtools/comparamsubset.py
```diff
@@ -20,10 +20,10 @@
...
 class ComparamSubset(OdxCategory):
-    comparams: NamedItemList[Comparam]
-    complex_comparams: NamedItemList[ComplexComparam]
-    data_object_props: NamedItemList[DataObjectProperty]
-    unit_spec: UnitSpec | None
+    comparams: NamedItemList[Comparam] = field(default_factory=NamedItemList)
+    complex_comparams: NamedItemList[ComplexComparam] = field(default_factory=NamedItemList)
+    data_object_props: NamedItemList[DataObjectProperty] = field(default_factory=NamedItemList)
+    unit_spec: UnitSpec | None = None
     category: str | None  # mandatory in ODX 2.2, but non-existent in ODX 2.0
```
kayoub5, line 27: No default value here?

odxtools/basevariantpattern.py
andlaus: for whatever reason, MATCHING-PARAMETER does not feature `.short_name`...

odxtools/comparamsubset.py
andlaus: thanks for the catch: these changes where done in a mostly automated manner and the comment at the end of the line seems to have screwed up my regexp. that said, for our default ODX version (2.2) this field is mandatory, so we should probably require to specify it explicitly?

kayoub5: Fine, just wanted to make sure

andlaus: ok, while it probably better not to specify a default value in this specific case, I found a few other instances where this happened: [fixed](https://github.com/mercedes-benz/odxtools/pull/411/commits/cb65db5d475fcfe67948b47e763403f363031881), thanks for bringing this up!

PR #412, opened 2025-04-07
tests, examples: Do not explicitly specify parameters which are set to the default value during instantiation

andlaus:
This PR is the follow-up to #411, and is another long and boring but monotopical PR. This one is probably holding the record for "most lines removed" in odxtools so far :)


examples/somersaultecu.py
```diff
@@ -383,15 +315,13 @@ class SomersaultSID(IntEnum):
             temperature_exp=1,
             molar_amount_exp=0,
             luminous_intensity_exp=0,
```
kayoub5, line 317: All of those really needed?

kayoub5 (approved):

PR #415, opened 2025-04-09
support computing the MUX static length if all the cases have the same size

nada-ben-ali:
This MR introduces support for calculating the static length of the multiplexer when all cases have the same size, enabling the codec length to be considered static.

### DIFF
--- a/odxtools/multiplexer.py
+++ b/odxtools/multiplexer.py
@@ -263,4 +263,5 @@ class Multiplexer(ComplexDop):
         if switch_key_size is None:
             return None
 
-        return max(switch_key_size + self.switch_key.byte_position, reference_size + self.byte_position)
+        return max(switch_key_size + self.switch_key.byte_position,
+                   reference_size + self.byte_position)
--- a/odxtools/multiplexer.py
+++ b/odxtools/multiplexer.py
@@ -247,15 +247,20 @@ class Multiplexer(ComplexDop):
         Otherwise, returns None to indicate that the size is dynamic.
         """
         reference_case = self.default_case if self.default_case else self.cases[0]
+
+        case_bit_length: int | None
         if reference_case.structure is None:
-            return None
-        case_bit_length = reference_case.structure.get_static_bit_length()
+            case_bit_length = 0
+        else:
+            case_bit_length = reference_case.structure.get_static_bit_length()
         if case_bit_length is None:
             return None
+        case_size: int | None
         for mux_case in self.cases:
             if mux_case.structure is None:
-                return None
-            case_size = mux_case.structure.get_static_bit_length()
+                case_size = 0
+            else:
+                case_size = mux_case.structure.get_static_bit_length()
             if case_size != case_bit_length:
                 return None  # Found a case with a different or unknown size
 
--- a/odxtools/multiplexer.py
+++ b/odxtools/multiplexer.py
@@ -265,5 +265,4 @@ class Multiplexer(ComplexDop):
 
         return max(
             switch_key_size + self.switch_key.byte_position * 8 +
-            (self.switch_key.bit_position if self.switch_key.bit_position else 0),
-            case_bit_length + self.byte_position * 8)
+            (self.switch_key.bit_position or 0), case_bit_length + self.byte_position * 8)
--- a/odxtools/multiplexer.py
+++ b/odxtools/multiplexer.py
@@ -236,3 +236,21 @@ class Multiplexer(ComplexDop):
         decode_state.origin_byte_position = orig_origin
 
         return result
+
+    @override
+    def get_static_bit_length(self) -> int | None:
+        """
+        Returns the static bit length of the multiplexer structure, if determinable.
+        If all cases (including the default, if present) have the same static bit length,
+        the codec length is considered static and is returned.
+        Otherwise, returns None to indicate that the size is dynamic.
+        """
+        reference_case = self.default_case if self.default_case else self.cases[0]
+        reference_size = reference_case.structure.get_static_bit_length()
+
+        for mux_case in self.cases:
+            case_size = mux_case.structure.get_static_bit_length()
+            if case_size != reference_size:
+                return None  # Found a case with a different or unknown size
+
+        return reference_size
--- a/odxtools/multiplexer.py
+++ b/odxtools/multiplexer.py
@@ -249,7 +249,7 @@ class Multiplexer(ComplexDop):
         reference_case = self.default_case if self.default_case else self.cases[0]
         if reference_case.structure is None:
             return None
-        reference_size = reference_case.structure.get_static_bit_length()
+        case_bit_length = reference_case.structure.get_static_bit_length()
         if reference_size is None:
             return None
         for mux_case in self.cases:
--- a/odxtools/multiplexer.py
+++ b/odxtools/multiplexer.py
@@ -241,6 +241,7 @@ class Multiplexer(ComplexDop):
     def get_static_bit_length(self) -> int | None:
         """
         Returns the static bit length of the multiplexer structure, if determinable.
+
         If all cases (including the default, if present) have the same static bit length,
         the codec length is considered static and is returned.
         Otherwise, returns None to indicate that the size is dynamic.
@@ -249,7 +250,8 @@ class Multiplexer(ComplexDop):
         if reference_case.structure is None:
             return None
         reference_size = reference_case.structure.get_static_bit_length()
-
+        if reference_size is None:
+            return None
         for mux_case in self.cases:
             if mux_case.structure is None:
                 return None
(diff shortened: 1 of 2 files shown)


odxtools/multiplexer.py
```diff
@@ -236,3 +236,25 @@ def decode_from_pdu(self, decode_state: DecodeState) -> ParameterValue:
...
+        reference_size = reference_case.structure.get_static_bit_length()
+
+        for mux_case in self.cases:
+            if mux_case.structure is None:
+                return None
+            case_size = mux_case.structure.get_static_bit_length()
+            if case_size != reference_size:
+                return None  # Found a case with a different or unknown size
+
+        return reference_size
```
kayoub5, line 260: size of mux = size of case + size of multiplexer field + any possible padding from offsets

odxtools/multiplexer.py
```diff
@@ -236,3 +236,25 @@ def decode_from_pdu(self, decode_state: DecodeState) -> ParameterValue:
         decode_state.origin_byte_position = orig_origin
 
         return result
+
+    @override
+    def get_static_bit_length(self) -> int | None:
+        """
+        Returns the static bit length of the multiplexer structure, if determinable.
```
andlaus, line 243: insert a blank line between the first line and the rest of the description. (Most documentation tools treat the first line as the summary...)

odxtools/multiplexer.py
andlaus: also note that, in theory, the contents of a multiplexer and its switch key can be non-consecutive...

nada-ben-ali: @kayoub5 @andlaus could you please recheck? I added a new commit.

odxtools/multiplexer.py
```diff
@@ -249,12 +250,17 @@ def get_static_bit_length(self) -> int | None:
...
             case_size = mux_case.structure.get_static_bit_length()
             if case_size != reference_size:
                 return None  # Found a case with a different or unknown size
 
-        return reference_size
+        switch_key_size = self.switch_key.dop.get_static_bit_length()
+        if switch_key_size is None:
+            return None
+
+        return max(switch_key_size + self.switch_key.byte_position, reference_size + self.byte_position)
```
kayoub5: * Missing switch_key bit position 
* mixing bits and bytes

kayoub5: Also missing unit test for the new function

tests/test_diag_data_dictionary_spec.py
```diff
@@ -277,12 +277,15 @@ def test_initialization(self) -> None:
         self.assertEqual(decoded, ("default_case", {}))
 
         self.assertEqual(mux.get_static_bit_length(), None)  # mux default case structure is None
-        mux.default_case.structure_ref = OdxLinkRef.from_id(mux_case2_struct.odx_id)
+        if mux.default_case:
+            mux.default_case.structure_ref = OdxLinkRef.from_id(mux_case2_struct.odx_id)
         ecu._resolve_odxlinks(odxlinks)
-        self.assertEqual(mux.get_static_bit_length(), None)  # mux cases doesn't have the same structure size
+        self.assertEqual(mux.get_static_bit_length(),
+                         None)  # mux cases doesn't have the same structure size
```
kayoub5, line 284: Keep comment above line, to avoid line wrapping

tests/test_diag_data_dictionary_spec.py
```diff
@@ -277,12 +277,15 @@ def test_initialization(self) -> None:
         self.assertEqual(decoded, ("default_case", {}))
 
         self.assertEqual(mux.get_static_bit_length(), None)  # mux default case structure is None
-        mux.default_case.structure_ref = OdxLinkRef.from_id(mux_case2_struct.odx_id)
+        if mux.default_case:
```
kayoub5, line 280: Why the if check in a test case?

nada-ben-ali: image

kayoub5: Do `assert mux.default_case` instead

nada-ben-ali: It didn't work. Locally, I didn't get this failure!!

nada-ben-ali: It's reproducible with Python 3.10 on Linux, but not with Python 3.12 on Windows.

it's fixed with adding `assert mux.default_case`, `self.assertIsNotNone(mux.default_case)` raised same issue.

kayoub5 (approved):

odxtools/multiplexer.py
```diff
@@ -236,3 +236,33 @@ def decode_from_pdu(self, decode_state: DecodeState) -> ParameterValue:
...
+            case_size = mux_case.structure.get_static_bit_length()
+            if case_size != case_bit_length:
+                return None  # Found a case with a different or unknown size
+
+        switch_key_size = self.switch_key.dop.get_static_bit_length()
+        if switch_key_size is None:
+            return None
+
+        return max(
+            switch_key_size + self.switch_key.byte_position * 8 +
```
andlaus, line 267: Note that ODX allows to specify atrocities like

- 5 bytes case content @ byte position 0
- 1 byte switch key @ byte position 20
- 4 bytes value parameter @ byte position 6
- 4 bytes value parameter @ byte position 22

what even is the size of the multiplexer in this case? While I don't think that we should handle these cases in the code, I would like to see this fact mentioned in the docstring for the function above.

tests/test_diag_data_dictionary_spec.py
```diff
@@ -276,6 +276,16 @@ def test_initialization(self) -> None:
         decoded = mux.decode_from_pdu(decode_state)
         self.assertEqual(decoded, ("default_case", {}))
 
+        self.assertEqual(mux.get_static_bit_length(), None)  # mux default case structure is None
```
andlaus, line 279: shouldn't the default case just be ignored for the size calculation if it is not present?

odxtools/multiplexer.py
```diff
@@ -236,3 +236,33 @@ def decode_from_pdu(self, decode_state: DecodeState) -> ParameterValue:
...
+        Returns the static bit length of the multiplexer structure, if determinable.
+
+        If all cases (including the default, if present) have the same static bit length,
+        the codec length is considered static and is returned.
+        Otherwise, returns None to indicate that the size is dynamic.
+        """
+        reference_case = self.default_case if self.default_case else self.cases[0]
+        if reference_case.structure is None:
+            return None
+        case_bit_length = reference_case.structure.get_static_bit_length()
```
andlaus, line 252: shouldn't the length of cases that don't define a structure be assumed to be zero, i.e.,
```suggestion
       if reference_case.structure is None:
            case_bit_length = 0
        else:
            case_bit_length = reference_case.structure.get_static_bit_length()
```

(analogous change required in the loop. this is quite an edge case because it will only return 0 if *all* cases are empty thus making the multiplexer pointless, but this would not be the worst thing I have seen in this industry...)

odxtools/multiplexer.py
```diff
@@ -236,3 +236,33 @@ def decode_from_pdu(self, decode_state: DecodeState) -> ParameterValue:
...
+    @override
+    def get_static_bit_length(self) -> int | None:
+        """
+        Returns the static bit length of the multiplexer structure, if determinable.
+
+        If all cases (including the default, if present) have the same static bit length,
+        the codec length is considered static and is returned.
+        Otherwise, returns None to indicate that the size is dynamic.
+        """
+        reference_case = self.default_case if self.default_case else self.cases[0]
```
andlaus, line 249: `self.default_case` should be tested for `is None`. another edge case: `self.cases` could be empty...

odxtools/multiplexer.py
kayoub5: it will be always the max of (byes required for key,  byes required for case)

bytes required = offset + size

current code handles all those "atrocities" just fine

andlaus: yeah, correct. If I think about it, this is an issue with any composite object in ODX, i.e., defining the static size as the minimum size required in a PDU is fine, one just needs to be aware that the sum of all static sizes is not necessarily the static size of the composite object. (IOW, the current code for structures is only "mostly correct", whereas "mostly" means for de-/encoding purposes...)

tests/test_diag_data_dictionary_spec.py
nada-ben-ali: it's present but defines no structure.

andlaus: good point. do you mind changing the comment to reflect that?

nada-ben-ali: this assertion is no longer needed after assuming `case_bit_length` is zero when the structure is undefined. I removed it.

odxtools/multiplexer.py
nada-ben-ali: done

odxtools/multiplexer.py
nada-ben-ali: the test now supports the following scenarios:
- mux cases don't have the same structure size
- mux cases have the same structure size
- only the default case is defined with no cases
- no default case, but cases have the same structure size
- no default case and each case has no associated structure

andlaus:
cool, let's merge. thanks for the contribution!

PR #416, opened 2025-04-12
Keep track of specialization level in inheritance resolution

zariiii9003:
Currently, only the layer type of the parent-ref is considered for resolving inheritance. I changed the return type of `_compute_available_objects()` to return the objects together with their source layer and specialization level.

I also changed the the specialization level of objects imported from ECU-SHARED-DATA according to 7.3.2.4.4.

Motivation: in my PDX file two functional groups FG1 and FG2 inherit from the same protocol PR. FG1 overrides some object. Resolving the objects of base variant BV, which inherits both FG1 and FG2, causes conflicts with the current odxtools implementation.

### DIFF
--- a/odxtools/diaglayers/diaglayer.py
+++ b/odxtools/diaglayers/diaglayer.py
@@ -29,9 +29,11 @@ from ..specialdatagroup import SpecialDataGroup
 from ..subcomponent import SubComponent
 from ..unitgroup import UnitGroup
 from .diaglayerraw import DiagLayerRaw
-from .diaglayertype import DiagLayerType
+from .diaglayertype import DiagLayerType, TInheritancePrio
 
 PrefixTree = dict[int, Union[list[DiagService], "PrefixTree"]]
+TInheritedObject = tuple[TNamed, "DiagLayer", TInheritancePrio]
+TInheritedObjects = Iterable[tuple[TNamed, "DiagLayer", TInheritancePrio]]
 
 
 @dataclass(kw_only=True)
@@ -156,9 +158,9 @@ class DiagLayer:
 
     def _compute_available_objects(
         self,
-        get_local_objects: Callable[["DiagLayer"], Iterable[TNamed]],
+        get_local_objects: Callable[["DiagLayer"], TInheritedObjects[TNamed]],
         get_not_inherited: Callable[[ParentRef], Iterable[str]],
-    ) -> Iterable[TNamed]:
+    ) -> TInheritedObjects[TNamed]:
         """Helper method to compute the set of all objects applicable
         to the DiagLayer if these objects are subject to the value
         inheritance mechanism
--- a/odxtools/diaglayers/diaglayer.py
+++ b/odxtools/diaglayers/diaglayer.py
@@ -1,10 +1,11 @@
 # SPDX-License-Identifier: MIT
+import sys
 from collections.abc import Callable, Iterable
 from copy import copy, deepcopy
 from dataclasses import dataclass
 from functools import cached_property
 from itertools import chain
-from typing import Any, Union, cast
+from typing import Any, Generic, Union, cast
 from xml.etree import ElementTree
 
 from ..admindata import AdminData
@@ -29,11 +30,21 @@ from ..specialdatagroup import SpecialDataGroup
 from ..subcomponent import SubComponent
 from ..unitgroup import UnitGroup
 from .diaglayerraw import DiagLayerRaw
-from .diaglayertype import DiagLayerType, TInheritancePrio
+from .diaglayertype import DiagLayerType, InheritancePriority
+
+# python 3.10 does not support generic namedtuples
+if sys.version_info >= (3, 11):
+    from typing import NamedTuple
+else:
+    from typing_extensions import NamedTuple
 
 PrefixTree = dict[int, Union[list[DiagService], "PrefixTree"]]
-TInheritedObject = tuple[TNamed, "DiagLayer", TInheritancePrio]
-TInheritedObjects = Iterable[tuple[TNamed, "DiagLayer", TInheritancePrio]]
+
+
+class InheritanceTriplet(Generic[TNamed], NamedTuple):
+    object: TNamed
+    source: "DiagLayer"
+    prio: InheritancePriority
 
 
 @dataclass(kw_only=True)
@@ -158,9 +169,9 @@ class DiagLayer:
 
     def _compute_available_objects(
         self,
-        get_local_objects: Callable[["DiagLayer"], TInheritedObjects[TNamed]],
+        get_local_objects: Callable[["DiagLayer"], Iterable[InheritanceTriplet[TNamed]]],
         get_not_inherited: Callable[[ParentRef], Iterable[str]],
-    ) -> TInheritedObjects[TNamed]:
+    ) -> Iterable[InheritanceTriplet[TNamed]]:
         """Helper method to compute the set of all objects applicable
         to the DiagLayer if these objects are subject to the value
         inheritance mechanism
--- a/odxtools/diaglayers/diaglayer.py
+++ b/odxtools/diaglayers/diaglayer.py
@@ -1,5 +1,4 @@
 # SPDX-License-Identifier: MIT
-import sys
 from collections.abc import Callable, Iterable
 from copy import copy, deepcopy
 from dataclasses import dataclass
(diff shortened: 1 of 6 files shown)


odxtools/diaglayers/diaglayertype.py
```diff
@@ -1,5 +1,8 @@
 # SPDX-License-Identifier: MIT
 from enum import Enum
+from typing import NewType
+
+TInheritancePrio = NewType("TInheritancePrio", float)
```
kayoub5, line 5: * why the T prefix ?
* int or float ?

odxtools/diaglayers/hierarchyelement.py
```diff
@@ -306,34 +305,40 @@ def _compute_available_objects(
...
                 # there's no problem if it was inherited from a parent
                 # of different priority than the one currently
                 # considered
-                orig_prio = result_dict[obj.short_name][1].variant_type.inheritance_priority
-                new_prio = parent_dl.variant_type.inheritance_priority
+                orig_prio = result_dict[obj.short_name][2]
+
+                if obj_prio == TInheritancePrio(0):
+                    # special case for inheriting from ECU-SHARED-DATA
+                    new_prio = cast(TInheritancePrio, self.variant_type.inheritance_priority - 0.5)
```
kayoub5, line 328: why 0.5 ?

odxtools/diaglayers/diaglayer.py
```diff
@@ -29,9 +29,11 @@
 from ..subcomponent import SubComponent
 from ..unitgroup import UnitGroup
 from .diaglayerraw import DiagLayerRaw
-from .diaglayertype import DiagLayerType
+from .diaglayertype import DiagLayerType, TInheritancePrio
 
 PrefixTree = dict[int, Union[list[DiagService], "PrefixTree"]]
+TInheritedObject = tuple[TNamed, "DiagLayer", TInheritancePrio]
```
kayoub5, line 35: I think it would be better to create an actual class/named tuple

odxtools/diaglayers/hierarchyelement.py
zariiii9003: Objects that are inherited from ECU-SHARED-DATA get a specialization, that is less than the inheriting layer but higher than the next general layer. That's also why i made it a float and not an int.

odxtools/diaglayers/diaglayertype.py
zariiii9003: T for Type 😄
The float is intentional, see my other comment.

odxtools/diaglayers/diaglayer.py
zariiii9003: I agree, i will change that.

odxtools/diaglayers/hierarchyelement.py
kayoub5: Does this 0.5 accumulate? for example if there are multiple ecu shared data?

odxtools/diaglayers/diaglayertype.py
kayoub5: T prefix is generally used with generic types, I don't see it being one here.

Example

```
from typing import TypeVar, List

TItem = TypeVar('TItem')  # TItem can be any type

def first_element(elements: List[TItem]) -> TItem:
    return elements[0] 
```

odxtools/diaglayers/hierarchyelement.py
zariiii9003: No accumulation, each object gets its own specialization value.

odxtools/diaglayers/diaglayertype.py
zariiii9003: I renamed it to `InheritancePriority`

andlaus (commented):
I think that I have to dissect the spec sentence by sentence when it comes to inheritance: I am fairly sure that the inheritance priorities defined by the spec are not meant to be transitive, i.e., two objects named X present in the functional groups FG1 and FG2 should exhibit the same priority even if the X of FG1 is inherited from a protocol and the X of FG2 is locally defined. (AFAIK, such conflicts need to be resolved using NOT-INHERITED lists.) If it turns out to be an issue in your dataset, would using the non-strict mode for that dataset work out for you?

odxtools/diaglayers/hierarchyelement.py
kayoub5: it would be simpler to put priority at int, each layer have its layer multiplied by 10, (0, 10, 20, etc)

and simply increase it here by 5.

reason is that float does not do equality well, see https://en.m.wikipedia.org/wiki/Floating-point_arithmetic#Accuracy_problems

odxtools/diaglayers/diaglayertype.py
```diff
@@ -1,5 +1,8 @@
 # SPDX-License-Identifier: MIT
 from enum import Enum
+from typing import NewType
+
+InheritancePriority = NewType("InheritancePriority", float)
```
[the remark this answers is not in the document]
kayoub5: I don't see the benefit of aliasing a simple type like float or int, it just complicates the code in this case

odxtools/diaglayers/diaglayer.py
```diff
@@ -29,11 +30,23 @@
...
 from ..unitgroup import UnitGroup
 from .diaglayerraw import DiagLayerRaw
-from .diaglayertype import DiagLayerType
+from .diaglayertype import DiagLayerType, InheritancePriority
+
+# python 3.10 does not support generic namedtuples
+if sys.version_info >= (3, 11):
+    from typing import NamedTuple
+else:
+    from typing_extensions import NamedTuple
```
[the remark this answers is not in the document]
kayoub5: NamedTuple exists even in python 3.7, no need to complicate the import 

https://docs.python.org/3.7/library/typing.html#typing.NamedTuple

odxtools/diaglayers/diaglayer.py
```diff
@@ -29,11 +30,23 @@
...
 
 PrefixTree = dict[int, Union[list[DiagService], "PrefixTree"]]
 
 
+class InheritanceTriplet(Generic[TNamed], NamedTuple):
```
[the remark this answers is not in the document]
zariiii9003: I wasn't sure whether dataclasses can be generic, but i'll try that.

odxtools/diaglayers/diaglayer.py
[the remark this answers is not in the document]
zariiii9003: Before 3.11 NamedTuple raises an Exception, because it does not support multiple inheritance.

I chose that `sys.version_info >= (3, 11)` construct because ruff will automatically remove it once you raise the minimum python version. As mentioned above, i will try a dataclass instead.

odxtools/diaglayers/hierarchyelement.py
```diff
@@ -306,130 +305,155 @@ def _compute_available_objects(
...
                 # considered
-                orig_prio = result_dict[obj.short_name][1].variant_type.inheritance_priority
-                new_prio = parent_dl.variant_type.inheritance_priority
+                orig_prio = result_dict[obj.short_name].prio
+
+                if obj_prio == InheritancePriority(0):
+                    # Special case for inheriting from ECU-SHARED-DATA:
+                    # Objects, that are inherited from ECU-SHARED-DATA, get a specialization,
+                    # that is less than the inheriting layer but higher than the next general layer.
+                    new_prio = InheritancePriority(self.variant_type.inheritance_priority - 0.5)
```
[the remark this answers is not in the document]
zariiii9003: I think that's true for imports from ECU-SHARED-DATA, inheritance is handled differently.

zariiii9003:
> I think that I have to dissect the spec sentence by sentence when it comes to inheritance: I am fairly sure that the inheritance priorities defined by the spec are not meant to be transitive, i.e., two objects named X present in the functional groups FG1 and FG2 should exhibit the same priority even if the X of FG1 is inherited from a protocol and the X of FG2 is locally defined. (AFAIK, such conflicts need to be resolved using NOT-INHERITED lists. Making these priorities transitive will make object inheritance quite unintuitive, e.g. objects might be overridden "laterally", i.e., a functional group might override an object from a base variant if the base variant inherited the object from a protocol and the functional group defined it locally (disregarding for a moment that ECU variants can only derive from a single base variant)). If it turns out to be an issue in your dataset, would using the non-strict mode for that dataset work out for you?

You're right, i just found the paragraph (page 52) which describes exactly the situation from my example. 
The objects in both functional groups have the same specialization **regardless of how they became part of that group**. 

So my implementation is not correct, although it works for my files.
Sorry to waste your time, closing :)

andlaus:
> So my implementation is not correct, although it works for my files.

if this is a problem which you've encounter in the wild, I am still open to accommodate it in non-strict mode. any thoughts?

PR #417, opened 2025-04-17
Fix the TableRow reconstruction during the unpickling

nada-ben-ali:
Following up on the discussion in https://github.com/mercedes-benz/odxtools/pull/410 regarding the issue found during the unpickling process for the `TableRow` object reconstruction.

The issue identified during the unpickling process for the `TableRow` object reconstruction is related to the usage of the dataclass with `kw_only=True`. This causes the generated `__init__()` method to require keyword arguments, which interferes with the default pickling behavior. Specifically, pickle typically uses the tuple returned by `__reduce__()`.

To resolve this, I implemented a custom `_reconstruct_tablerow` function that bypasses the `__init__()` method by using `cls.__new__(cls)`. This allows us to create an instance without invoking `__init__()`. In the `__reduce__()` method, I return this reconstruction function along with the necessary arguments.

As a result, during the unpickling process, Python calls `_reconstruct_tablerow(self.__class__, saved_dict)` to properly reconstruct the object.

### DIFF
--- a/odxtools/tablerow.py
+++ b/odxtools/tablerow.py
@@ -1,6 +1,6 @@
 # SPDX-License-Identifier: MIT
-from dataclasses import dataclass, field, fields
-from typing import TYPE_CHECKING, Any
+from dataclasses import dataclass, field
+from typing import TYPE_CHECKING, Any, cast
 from xml.etree import ElementTree
 
 from .admindata import AdminData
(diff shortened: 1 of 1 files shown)


odxtools/tablerow.py
```diff
@@ -260,4 +260,26 @@ def _resolve_snrefs(self, context: SnRefContext) -> None:
     def __reduce__(self) -> tuple[Any, ...]:
         """This ensures that the object can be correctly reconstructed during unpickling."""
         state = self.__dict__.copy()
-        return self.__class__, tuple([getattr(self, x.name) for x in fields(self)]), state
+        return _reconstruct_tablerow, (self.__class__, state)
```
andlaus, line 263: hm. according to [The Documentation](https://docs.python.org/3/library/pickle.html#object.__reduce__), the fifth item of the tuple returned are the keyword arguments passed to the constructor during unpickling, i.e., `__reduce__` should possibly become something like
```python
    def __reduce__(self) -> tuple[Any, ...]:
        """This ensures that the object can be correctly reconstructed during unpickling."""
        state = self.__dict__.copy()
        kwargs = { x.name: getattr(self, x.name) for x in fields(self) }
        return self.__class__, tuple(), state, None, kwargs.items()
```

(I have not tested it it works, so it would  be great if you checked it out...)

nada-ben-ali: I had already tried that, but it didn’t work. I got: `fifth element of the tuple returned by __reduce__ must be an iterator, not dict_items`. As stated in the documentation, the fifth element must be an iterator, so I converted it to an iterator and got `TypeError: TableRow.__init__() missing 4 required keyword-only arguments: 'short_name', 'odx_id', 'table_ref', and 'key_raw'`.

andlaus: you're right: I misread the documentation as this item as being the keyword arguments passed to the constructor. Let's merge.

BTW: using `copy(self.__dict__)` as the state makes pickled table pretty big. possibly, it should be changed to just the dataclass fields. (unpickled table rows will be not be finalized then, though.)

PR #420, opened 2025-04-29
generalize ODXLINK reference writing

andlaus:
writing ODXLINK references should now always work regardless of in which document fragments the reference and the target objects are located.

this works by providing three new "magical" global functions to the jinja macros for output:

- `set_category_docfrag()`: Specify the document fragment for the ODX category which is applicable for all subsequent macro invocations.
- `set_layer_docfrag()`: Specify the document fragment for the diagnostic layer which is applicable for all subsequent macro invocations (for DIAG-LAYER-CONTAINER).
- `make_ref_attribs(ref: OdxLinkRef)`: produces all necessary XML attributes required for an ODX link reference XML tag. (i.e., ID-REF, DOCREF and DOCTYPE)

internally, these functions require a dictionary parameter for storing "magic" global variables. On the jinja side, this is not necessary because lambda functions are used as wrappers.

### DIFF
--- a/odxtools/writepdxfile.py
+++ b/odxtools/writepdxfile.py
@@ -17,8 +17,6 @@ from .odxtypes import bool_to_odxstr
 
 odxdatabase: Database | None = None
 
-cur_docfrags: list[OdxDocFragment] = []
-
 
 def jinja2_odxraise_helper(msg: str) -> None:
     raise Exception(msg)
--- a/odxtools/writepdxfile.py
+++ b/odxtools/writepdxfile.py
@@ -15,8 +15,6 @@ from .database import Database
 from .odxlink import DocType, OdxDocFragment, OdxLinkRef
 from .odxtypes import bool_to_odxstr
 
-odxdatabase: Database | None = None
-
 
 def jinja2_odxraise_helper(msg: str) -> None:
     raise Exception(msg)
@@ -82,10 +80,6 @@ def write_pdx_file(
     """
     Write an internalized database to a PDX file.
     """
-    global odxdatabase
-
-    odxdatabase = database
-
     file_index = []
     with zipfile.ZipFile(output_file_name, mode="w", compression=zipfile.ZIP_DEFLATED) as zf:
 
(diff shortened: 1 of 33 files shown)


odxtools/writepdxfile.py
```diff
@@ -128,9 +164,14 @@ def write_pdx_file(
...
-        vars: dict[str, Any] = {}
-        vars["odxtools_version"] = odxtools.__version__
-        vars["database"] = database
+        jinja_vars: dict[str, Any] = {}
+        jinja_vars["odxtools_version"] = odxtools.__version__
+        jinja_vars["database"] = database
+
+        jinja_env.globals["set_category_docfrag"] = lambda cname, ctype: set_category_docfrag(
+            jinja_vars, cname, ctype)
+        jinja_env.globals["set_layer_docfrag"] = lambda lname: set_layer_docfrag(jinja_vars, lname)
```
kayoub5, line 173: `from functools import partial`
```suggestion
        jinja_env.globals["set_layer_docfrag"] = partial(set_layer_docfrag, jinja_vars)
```

andlaus: I think this should stay with the `lambda`s because `partial` is much less known (I didn't know it) and it is IMO also less explicit...

odxtools/writepdxfile.py
```diff
@@ -12,10 +12,13 @@
 import odxtools
 
 from .database import Database
+from .odxlink import DocType, OdxDocFragment, OdxLinkRef
 from .odxtypes import bool_to_odxstr
 
 odxdatabase: Database | None = None
 
+cur_docfrags: list[OdxDocFragment] = []
```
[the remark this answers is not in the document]
andlaus: thanks for the catch. that was an artifact from the initial version of the patch until I realized that the global variable is not required. [fixed](https://github.com/mercedes-benz/odxtools/pull/420/commits/dd6a925867d3457712666ed1dd9bc5cb41fb2336)

kayoub5 (approved):

Issue #289, opened 2024-04-09
Support odx-f?


chenyin0126:
Looks like we only support odx-d, odx-c and odx-cs. Do we have any plan to support odx-f?

andlaus:
I currently do not intend to implement this myself, but I'd be happy to review and accept patches for it...

andlaus:
with #422, this should be implemented. closing.

PR #422, opened 2025-05-06
Implement support for flash specifcations (`.odx-f`)

andlaus:
This includes support for internalization and output writing of `FLASH` containers. Also, I tried to make the unit test quite comprehensive.

This PR is pretty large, but it should contain a very small number of "drive-by fixes" and -- in contrast to the diagnostics functionality -- there does not seem to be much logic required, so it should be fairly simple. I also tested this with some of the `.odx-f` files which I have access to, but they do not use all features, so there might be a few bugs lurking in the darker corners...

This implements https://github.com/mercedes-benz/odxtools/issues/289 .

### DIFF
--- a/odxtools/addrdeffilter.py
+++ b/odxtools/addrdeffilter.py
@@ -1,0 +1,33 @@
+# SPDX-License-Identifier: MIT
+from dataclasses import dataclass
+from typing import Any
+from xml.etree import ElementTree
+
+from .exceptions import odxrequire
+from .filter import Filter
+from .odxdoccontext import OdxDocContext
+from .odxlink import OdxLinkDatabase, OdxLinkId
+from .snrefcontext import SnRefContext
+from .utils import dataclass_fields_asdict, read_hex_binary
+
+
+@dataclass(kw_only=True)
+class AddrdefFilter(Filter):
+    filter_end: int
+
+    @staticmethod
+    def from_et(et_element: ElementTree.Element, context: OdxDocContext) -> "AddrdefFilter":
+        kwargs = dataclass_fields_asdict(Filter.from_et(et_element, context))
+
+        filter_end = odxrequire(read_hex_binary(et_element.find("FILTER-END")))
+
+        return AddrdefFilter(filter_end=filter_end, **kwargs)
+
+    def _build_odxlinks(self) -> dict[OdxLinkId, Any]:
+        return super()._build_odxlinks()
+
+    def _resolve_odxlinks(self, odxlinks: OdxLinkDatabase) -> None:
+        super()._resolve_odxlinks(odxlinks)
+
+    def _resolve_snrefs(self, context: SnRefContext) -> None:
+        super()._resolve_snrefs(context)
--- a/odxtools/database.py
+++ b/odxtools/database.py
@@ -19,6 +19,7 @@ from .diaglayers.ecuvariant import EcuVariant
 from .diaglayers.functionalgroup import FunctionalGroup
 from .diaglayers.protocol import Protocol
 from .exceptions import odxraise, odxrequire
+from .flash import Flash
 from .nameditemlist import NamedItemList
 from .odxdoccontext import OdxDocContext
 from .odxlink import DocType, OdxDocFragment, OdxLinkDatabase, OdxLinkId
@@ -39,6 +40,7 @@ class Database:
         self._diag_layer_containers = NamedItemList[DiagLayerContainer]()
         self._comparam_subsets = NamedItemList[ComparamSubset]()
         self._comparam_specs = NamedItemList[ComparamSpec]()
+        self._flashs = NamedItemList[Flash]()
         self._short_name = "odx_database"
 
     def add_pdx_file(self, pdx_file: Union[str, "PathLike[Any]", IO[bytes], ZipFile]) -> None:
@@ -57,7 +59,7 @@ class Database:
             p = Path(zip_member)
             if p.suffix.lower().startswith(".odx"):
                 root = ElementTree.parse(pdx_zip.open(zip_member)).getroot()
-                self._process_xml_tree(root)
+                self.add_xml_tree(root)
             elif p.name.lower() == "index.xml":
                 root = ElementTree.parse(pdx_zip.open(zip_member)).getroot()
                 db_short_name = odxrequire(root.findtext("SHORT-NAME"))
@@ -66,7 +68,7 @@ class Database:
                 self.add_auxiliary_file(zip_member, pdx_zip.open(zip_member))
 
     def add_odx_file(self, odx_file_name: Union[str, "PathLike[Any]"]) -> None:
-        self._process_xml_tree(ElementTree.parse(odx_file_name).getroot())
+        self.add_xml_tree(ElementTree.parse(odx_file_name).getroot())
 
     def add_auxiliary_file(self,
                            aux_file_name: Union[str, "PathLike[Any]"],
@@ -76,7 +78,7 @@ class Database:
 
         self.auxiliary_files[str(aux_file_name)] = aux_file_obj
 
-    def _process_xml_tree(self, root: ElementTree.Element) -> None:
+    def add_xml_tree(self, root: ElementTree.Element) -> None:
         # ODX spec version
         model_version = Version(root.attrib.get("MODEL-VERSION", "2.0"))
         if self.model_version is not None and self.model_version != model_version:
@@ -112,6 +114,9 @@ class Database:
                 self._comparam_subsets.append(ComparamSubset.from_et(category_et, context))
             else:
                 self._comparam_specs.append(ComparamSpec.from_et(category_et, context))
+        elif category_tag == "FLASH":
+            context = OdxDocContext(model_version, (OdxDocFragment(category_sn, DocType.FLASH),))
+            self._flashs.append(Flash.from_et(category_et, context))
 
     def refresh(self) -> None:
         # Create wrapper objects
@@ -143,6 +148,9 @@ class Database:
         for dlc in self.diag_layer_containers:
             dlc._resolve_odxlinks(self._odxlinks)
 
+        for flash in self.flashs:
+            flash._resolve_odxlinks(self._odxlinks)
+
         # resolve short name references for containers which do not do
         # inheritance (we can call directly call _resolve_snrefs())
         context = SnRefContext()
@@ -155,6 +163,8 @@ class Database:
             spec._finalize_init(self, self._odxlinks)
         for dlc in self.diag_layer_containers:
             dlc._finalize_init(self, self._odxlinks)
+        for flash in self.flashs:
+            flash._finalize_init(self, self._odxlinks)
 
         for subset in self.comparam_subsets:
             subset._resolve_snrefs(context)
@@ -162,6 +172,8 @@ class Database:
             spec._resolve_snrefs(context)
         for dlc in self.diag_layer_containers:
             dlc._resolve_snrefs(context)
+        for flash in self.flashs:
+            flash._resolve_snrefs(context)
 
     def _build_odxlinks(self) -> dict[OdxLinkId, Any]:
         result: dict[OdxLinkId, Any] = {}
@@ -174,6 +186,9 @@ class Database:
 
         for dlc in self.diag_layer_containers:
             result.update(dlc._build_odxlinks())
+
+        for flash in self.flashs:
+            result.update(flash._build_odxlinks())
 
         return result
 
@@ -248,10 +263,15 @@ class Database:
     def comparam_specs(self) -> NamedItemList[ComparamSpec]:
         return self._comparam_specs
 
+    @property
+    def flashs(self) -> NamedItemList[Flash]:
+        return self._flashs
+
     def __repr__(self) -> str:
         return f"Database(model_version={self.model_version}, " \
             f"protocols={[x.short_name for x in self.protocols]}, " \
             f"ecus={[x.short_name for x in self.ecus]}, " \
             f"diag_layer_containers={repr(self.diag_layer_containers)}, " \
             f"comparam_subsets={repr(self.comparam_subsets)}, " \
-            f"comparam_specs={repr(self.comparam_specs)})"
+            f"comparam_specs={repr(self.comparam_specs)}, " \
+            f"flashs={repr(self.flashs)})"
--- a/odxtools/diagdatadictionaryspec.py
+++ b/odxtools/diagdatadictionaryspec.py
@@ -46,9 +46,7 @@ class DiagDataDictionarySpec:
     @staticmethod
     def from_et(et_element: ElementTree.Element,
                 context: OdxDocContext) -> "DiagDataDictionarySpec":
-        admin_data = None
-        if (admin_data_elem := et_element.find("ADMIN-DATA")) is not None:
-            admin_data = AdminData.from_et(admin_data_elem, context)
+        admin_data = AdminData.from_et(et_element.find("ADMIN-DATA"), context)
 
         dtc_dops = NamedItemList([
             DtcDop.from_et(dtc_dop_elem, context)
--- a/odxtools/encryptcompressmethodtype.py
+++ b/odxtools/encryptcompressmethodtype.py
@@ -1,10 +1,10 @@
 # SPDX-License-Identifier: MIT
-from enum import StrEnum
+from enum import Enum
 
 from .odxtypes import DataType
 
 
-class EncryptCompressMethodType(StrEnum):
+class EncryptCompressMethodType(Enum):
     A_UINT32 = "A_UINT32"
     A_BYTEFIELD = "A_BYTEFIELD"
 
--- a/odxtools/encryptcompressmethodtype.py
+++ b/odxtools/encryptcompressmethodtype.py
@@ -1,10 +1,10 @@
 # SPDX-License-Identifier: MIT
-from enum import Enum
+from enum import StrEnum
 
 from .odxtypes import DataType
 
 
-class EncryptCompressMethodType(Enum):
+class EncryptCompressMethodType(StrEnum):
     A_UINT32 = "A_UINT32"
     A_BYTEFIELD = "A_BYTEFIELD"
 
--- a/odxtools/encryptcompressmethodtype.py
+++ b/odxtools/encryptcompressmethodtype.py
@@ -1,0 +1,7 @@
+# SPDX-License-Identifier: MIT
+from enum import Enum
+
+
+class EncryptCompressMethodType(Enum):
+    A_UINT32 = "A_UINT32"
+    A_BYTEFIELD = "A_BYTEFIELD"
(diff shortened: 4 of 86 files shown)


odxtools/database.py
```diff
@@ -76,7 +78,7 @@ def add_auxiliary_file(self,
 
         self.auxiliary_files[str(aux_file_name)] = aux_file_obj
 
-    def _process_xml_tree(self, root: ElementTree.Element) -> None:
+    def add_xml_tree(self, root: ElementTree.Element) -> None:
```
kayoub5, line 81: Not private?

odxtools/addrdeffilter.py
```diff
@@ -0,0 +1,33 @@
...
+from .exceptions import odxrequire
+from .filter import Filter
+from .odxdoccontext import OdxDocContext
+from .odxlink import OdxLinkDatabase, OdxLinkId
+from .snrefcontext import SnRefContext
+from .utils import dataclass_fields_asdict, read_hex_binary
+
+
+@dataclass(kw_only=True)
+class AddrdefFilter(Filter):
```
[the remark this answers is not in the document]
andlaus: As far as I can see, the spec does not say explicitly, but I suppose it means "address defined".

odxtools/database.py
andlaus: it was useful for the unit test, so I figured to make it public...

odxtools/diagdatadictionaryspec.py
```diff
@@ -46,9 +46,7 @@ class DiagDataDictionarySpec:
     @staticmethod
     def from_et(et_element: ElementTree.Element,
                 context: OdxDocContext) -> "DiagDataDictionarySpec":
-        admin_data = None
-        if (admin_data_elem := et_element.find("ADMIN-DATA")) is not None:
-            admin_data = AdminData.from_et(admin_data_elem, context)
+        admin_data = AdminData.from_et(et_element.find("ADMIN-DATA"), context)
```
kayoub5, line 49: Why?

odxtools/encryptcompressmethodtype.py
```diff
@@ -0,0 +1,7 @@
+# SPDX-License-Identifier: MIT
+from enum import Enum
+
+
+class EncryptCompressMethodType(Enum):
+    A_UINT32 = "A_UINT32"
+    A_BYTEFIELD = "A_BYTEFIELD"
```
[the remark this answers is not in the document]
kayoub5: Why not re-use existing data type enum?

odxtools/expectedident.py
```diff
@@ -0,0 +1,40 @@
...
+from .odxdoccontext import OdxDocContext
+from .odxlink import OdxLinkDatabase, OdxLinkId
+from .snrefcontext import SnRefContext
+from .utils import dataclass_fields_asdict
+
+
+@dataclass(kw_only=True)
+class ExpectedIdent(IdentifiableElement):
+    """
+    Corresponds to ADDITIONAL-AUDIENCE.
```
kayoub5, line 17: `ADDITIONAL-AUDIENCE` ?

odxtools/diagdatadictionaryspec.py
andlaus: `AdminData.from_et()` accepts `None` for the `et_element` and returns `None` in this case. This simplifies the many places where `AdminData` is required. (Same for SDGs, BTW.)

odxtools/externflashdata.py
```diff
@@ -0,0 +1,34 @@
...
+    @staticmethod
+    def from_et(et_element: ElementTree.Element, context: OdxDocContext) -> "ExternFlashdata":
+        kwargs = dataclass_fields_asdict(Flashdata.from_et(et_element, context))
+
+        datafile = Datafile.from_et(odxrequire(et_element.find("DATAFILE")), context)
+
+        return ExternFlashdata(datafile=datafile, **kwargs)
+
+    def _build_odxlinks(self) -> dict[OdxLinkId, Any]:
+        return super()._build_odxlinks()
```
kayoub5, line 28: Why overload a function if you are just going to call super?

odxtools/filter.py
```diff
@@ -0,0 +1,32 @@
...
+from .exceptions import odxrequire
+from .odxdoccontext import OdxDocContext
+from .odxlink import OdxLinkDatabase, OdxLinkId
+from .snrefcontext import SnRefContext
+from .utils import read_hex_binary
+
+
+@dataclass(kw_only=True)
+class Filter:
+    filter_start: int
```
kayoub5, line 15: a class for just an int?

odxtools/flashclass.py
```diff
@@ -0,0 +1,32 @@
...
+from .odxlink import OdxLinkDatabase, OdxLinkId
+from .snrefcontext import SnRefContext
+from .utils import dataclass_fields_asdict
+
+
+@dataclass(kw_only=True)
+class FlashClass(IdentifiableElement):
+    """
+    Corresponds to FLASH-CLASS.
+    """
```
kayoub5, line 17: No properties?

odxtools/flashdata.py
```diff
@@ -0,0 +1,70 @@
...
+from .odxlink import OdxLinkDatabase, OdxLinkId
+from .snrefcontext import SnRefContext
+from .utils import dataclass_fields_asdict
+
+
+@dataclass(kw_only=True)
+class Flashdata(IdentifiableElement):
+    size_length: int | None = None
+    address_length: int | None = None
+    dataformat: Dataformat
```
kayoub5, line 20:
```suggestion
    data_format: Dataformat
```

odxtools/fwsignature.py
```diff
@@ -0,0 +1,30 @@
...
+
+from .exceptions import odxraise, odxrequire
+from .odxdoccontext import OdxDocContext
+from .sessionsubelemtype import SessionSubElemType
+
+
+@dataclass(kw_only=True)
+class FwSignature:
+    value: str
+    value_type: SessionSubElemType
```
kayoub5, line 14: Duplicate with FwChecksum

odxtools/identdesc.py
```diff
@@ -0,0 +1,54 @@
...
+    def _build_odxlinks(self) -> dict[OdxLinkId, Any]:
+        odxlinks: dict[OdxLinkId, Any] = {}
+
+        return odxlinks
+
+    def _resolve_odxlinks(self, odxlinks: OdxLinkDatabase) -> None:
+        pass
+
+    def _resolve_snrefs(self, context: SnRefContext) -> None:
+        # TODO: resolve the short name references. for IdentDesc this
```
kayoub5, line 52: Planned for this PR? ticket?

odxtools/encryptcompressmethodtype.py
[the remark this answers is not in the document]
andlaus: according to the spec, yes. (`IdentValueType` is another instance of this. IMO this could be changed to any data type via `DataType`)

odxtools/identdesc.py
```diff
@@ -0,0 +1,54 @@
...
+            diag_comm_snref=diag_comm_snref,
+            ident_if_snref=ident_if_snref,
+            out_param_if_snref=out_param_if_snref,
+            out_param_if_snpathref=out_param_if_snpathref,
+        )
+
+    def _build_odxlinks(self) -> dict[OdxLinkId, Any]:
+        odxlinks: dict[OdxLinkId, Any] = {}
+
+        return odxlinks
```
kayoub5, line 46: Objects that don't participate in the link resolution, should not have such methods, they waste CPU and RAM without benefit.

odxtools/identvalue.py
```diff
@@ -0,0 +1,32 @@
...
+class IdentValue:
+    """
+    Corresponds to IDENT-VALUE.
+    """
+
+    value: str
+
+    # note that the spec says this attribute is named "TYPE", but in
+    # python, "type" is a build-in function...
+    value_type: IdentValueType
```
kayoub5, line 20: `type` should not be used as variable name, but could be used as an attribute without side effects
```suggestion
    type: IdentValueType
```

odxtools/identvaluetype.py
```diff
@@ -0,0 +1,8 @@
+# SPDX-License-Identifier: MIT
+from enum import Enum
+
+
+class IdentValueType(Enum):
+    A_UINT32 = "A_UINT32"
+    A_BYTEFIELD = "A_BYTEFIELD"
+    A_ASCIISTRING = "A_ASCIISTRING"
```
kayoub5, line 8: Not other types?

odxtools/ownident.py
```diff
@@ -0,0 +1,38 @@
...
+from .odxdoccontext import OdxDocContext
+from .odxlink import OdxLinkDatabase, OdxLinkId
+from .snrefcontext import SnRefContext
+from .utils import dataclass_fields_asdict
+
+
+@dataclass(kw_only=True)
+class OwnIdent(IdentifiableElement):
+    """
+    Corresponds to ADDITIONAL-AUDIENCE.
```
kayoub5, line 18: ` ADDITIONAL-AUDIENCE` ?

odxtools/physsegment.py
```diff
@@ -0,0 +1,42 @@
...
+from .exceptions import odxrequire
+from .odxdoccontext import OdxDocContext
+from .odxlink import OdxLinkDatabase, OdxLinkId
+from .snrefcontext import SnRefContext
+from .utils import dataclass_fields_asdict, read_hex_binary
+
+
+@dataclass(kw_only=True)
+class PhysSegment(IdentifiableElement):
+    fillbyte: int | None = None
```
kayoub5, line 16:
```suggestion
    fill_byte: int | None = None
```

odxtools/filter.py
andlaus: it looks like it. but FILTER is a base class like PARAMETER, i.e., all instances of FILTER are either `SIZEDEF-FILTER` or `ADDRDEF-FILTER` (which need to be specified via `xsi:type`)...

odxtools/flashclass.py
andlaus: nope, but the spec has this tag type, so we should have it as well. (ADDITIONAL-AUDIENCE is the same, BTW.)

odxtools/externflashdata.py
andlaus: basically to signal that these methods ought to be called by the parent object. That said, this is not consistently done in the code base...

odxtools/flashdata.py
andlaus: I agree, but the spec calls the XML tag `DATAFORMAT`...

odxtools/fwsignature.py
andlaus: yep. the spec does it like this though. maybe this should be made a type alias...

odxtools/identdesc.py
andlaus: some objects which do not have an ODX ID still "own" objects which do. I'd like to always have the full set of the link resolution functions or none at all... (the performance penalty of calling empty functions is negligible IMO.)

andlaus: (in this case the SNREFs might need to be resolved in the future...)

odxtools/identvalue.py
andlaus: I would prefer this as well, but `type()` is a build-in function.

odxtools/identvaluetype.py
andlaus: not according to the XSD...

odxtools/physsegment.py
andlaus: the XSD calls the tag `FILLBYTE`...

odxtools/ownident.py
andlaus: thanks, fixed.

odxtools/expectedident.py
andlaus: copy and paste: best friend, worst enemy. [fixed](https://github.com/mercedes-benz/odxtools/pull/422/commits/cece9ced4375492629941f755d4db1d7a8154931), thanks.

odxtools/identvalue.py
andlaus: I'd like to avoid this anyway. If nothing else, it prevents confusing the syntax highlighting of IDEs. (BTW, the same argument can be made for `.odx_id` vs `.id`.)

odxtools/flashdata.py
kayoub5: it's a clear typo in the spec, not the first or last that we run into, mo need to follow it

odxtools/identdesc.py
kayoub5: A general good practice is to have an issue/ticket for each TODO, this helps people who wants to contribute know what's already on the TODO list.

odxtools/identdesc.py
kayoub5: given enough objects, even calling an empty function have a cost

odxtools/physsegment.py
kayoub5: it's a clear typo in the spec, not the first or last that we run into, no need to follow it

odxtools/encryptcompressmethodtype.py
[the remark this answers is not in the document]
andlaus: possibly, a `.to_datatype() -> DataType` method or a `.datatype` property should be introduced?

odxtools/physsegment.py
andlaus: In contrast to the incorrect pluralizations ("SECURITYS", "FLASH-CLASSS", ...) and definitive typos ("POS-RESPONSE-SUPPRESSABLE"), I'm actually not so sure in this instance: Even in English, words are sometimes amalgamated (e.g., `bytearray`). I tend to think that `DATABLOCK` should better be `DATA-BLOCK`, but even there I am not 100% sure and if in doubt we should IMO stick with what the spec uses...

odxtools/identdesc.py
andlaus: yes, but compared to the cost of parsing and internalizing the XML, it is pretty negligible. I don't think that having to wait a second more for a dataset that takes two minutes to load makes a big difference in practice...

andlaus:
Ok: I added two smallish drive-by fixes that are only tangentially related to flash container support: https://github.com/mercedes-benz/odxtools/pull/422/commits/17ab3fbcf458a6e4b90959b5821e9d5ff4fcec4f and https://github.com/mercedes-benz/odxtools/pull/422/commits/6914624d5380d8a761e84d16d26f757e397ca131 (although I stumbled about them when playing with my `.odx-f` files). If you really want this, I can move these two patches into a separate PR...

odxtools/flashdata.py
andlaus: I'm not so sure: database is used as a single word in 99.5% of all cases...

odxtools/encryptcompressmethodtype.py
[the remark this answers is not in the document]
kayoub5: Maybe `StrEnum` can make it better?

See https://docs.python.org/3/library/enum.html#enum.StrEnum

odxtools/filter.py
kayoub5: So abstract class?

odxtools/encryptcompressmethodtype.py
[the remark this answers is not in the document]
andlaus: good suggestion. I just checked: Whilst it enables to get rid of a few instances of these annoying `.value` calls in the output writing code, the fundamental problem that enum classes cannot be subclassed is present for `StrEnum` as well:
```
> mypy odxtools
odxtools/encryptcompressmethodtype.py:7: error: Cannot extend enum with existing members: "DataType"  [misc]
Found 1 error in 1 file (checked 246 source files)
```
(after making `EncryptCompressMethodType` derive from `DataType`.)

odxtools/filter.py
andlaus: yes. That said, we cannot make it an ABC on the python side because this would break the `.from_et()` mechanism because this requires the base class to be instantiable...

odxtools/flashclass.py
```diff
@@ -0,0 +1,32 @@
...
+
+    @staticmethod
+    def from_et(et_element: ElementTree.Element, context: OdxDocContext) -> "FlashClass":
+        kwargs = dataclass_fields_asdict(IdentifiableElement.from_et(et_element, context))
+
+        return FlashClass(**kwargs)
+
+    def _build_odxlinks(self) -> dict[OdxLinkId, Any]:
+        return {self.odx_id: self}
```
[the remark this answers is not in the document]
andlaus: nope. That's a thing I've had on my TODO list for a while, but I haven't come to it because it is quite a bit of work and so far, there always were more important things...

odxtools/externflashdata.py
kayoub5: I don't get what you mean

odxtools/flashclass.py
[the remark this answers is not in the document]
kayoub5: TODO = open issue ?

odxtools/externflashdata.py
andlaus: to allow easy extension of these classes or changes in the class hierarchy. (IMO, the fact that in OOP all methods of parent classes are implicitly available in derived classes is a course and a benefit at the same time...)

andlaus: (in the original comment, I should have better said "owner object" instead of "parent object"...)

odxtools/flashclass.py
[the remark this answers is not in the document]
andlaus: done: https://github.com/mercedes-benz/odxtools/issues/425

odxtools/fwsignature.py
kayoub5: Other than this, LGTM

andlaus:
@kayoub5: do you see any further issues with this that should be fixed before merging?

odxtools/fwsignature.py
kayoub5: @andlaus this topic is still open, any plans to fix it?

odxtools/encryptcompressmethodtype.py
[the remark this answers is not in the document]
andlaus: Yes, but I assumed you you don't like drive-by fixes in unrelated PRs, so I intended to make a separate one for this. Since this seems not to be the case here, I added the patch: https://github.com/mercedes-benz/odxtools/pull/422/commits/7f0010e293c549315bad31243d2a3df76fced1ea . thanks for proposing this :)

odxtools/fwsignature.py
andlaus: for the same reason as for the `StrEnum` change I intended to do this in a separate PR. Anyway, I added the patch for this: https://github.com/mercedes-benz/odxtools/pull/422/commits/31a8abba2d6ea4e926269a93f2dfad3027bedb04

andlaus: Unfortunately I had to revert this change: `StrEnum` was only introduced in python 3.11...

kayoub5 (approved):

Issue #424, opened 2025-05-08
Property parent_refs missing in Protocol?


svnmic:
Resolving snrefs sometimes comes up with incomplete results, because this property is missing. Something like
`
@property
def parent_refs(self: Protocol) -> list[ParentRef]:
    return self.protocol_raw.parent_refs
`

andlaus:
Thanks for the catch! It seems like the proxy properties for `comparam_spec_refs` and `prot_stack_snref` are also missing. can you make a PR? (in this context, it would be nice if you checked that all properties exposed by the raw  diagnostic layer classes were forwarded by the non-raw ones...)

PR #427, opened 2025-05-19
Some fixes

andlaus:
This PR contains a few smallish fixes: The biggest one is a fix for https://github.com/mercedes-benz/odxtools/issues/424 , while the other two are the inversion of a spurious `if` condition in a relatively exotic part of the code (`SubComponentParamConnector`), and the improvement of the error message one gets if the result of a reference resolution does not exhibit the expected type.

### DIFF
--- a/odxtools/diaglayers/protocol.py
+++ b/odxtools/diaglayers/protocol.py
@@ -7,6 +7,8 @@ from xml.etree import ElementTree
 from ..comparamspec import ComparamSpec
 from ..exceptions import odxassert
 from ..odxdoccontext import OdxDocContext
+from ..odxlink import OdxLinkRef
+from ..parentref import ParentRef
 from ..protstack import ProtStack
 from .hierarchyelement import HierarchyElement
 from .protocolraw import ProtocolRaw
@@ -25,12 +27,24 @@ class Protocol(HierarchyElement):
         return cast(ProtocolRaw, self.diag_layer_raw)
 
     @property
+    def comparam_spec_ref(self) -> OdxLinkRef:
+        return self.protocol_raw.comparam_spec_ref
+
+    @property
     def comparam_spec(self) -> ComparamSpec:
         return self.protocol_raw.comparam_spec
 
     @property
+    def prot_stack_snref(self) -> str | None:
+        return self.protocol_raw.prot_stack_snref
+
+    @property
     def prot_stack(self) -> ProtStack | None:
         return self.protocol_raw.prot_stack
+
+    @property
+    def parent_refs(self) -> list[ParentRef]:
+        return self.protocol_raw.parent_refs
 
     @staticmethod
     def from_et(et_element: ElementTree.Element, context: OdxDocContext) -> "Protocol":
--- a/odxtools/odxlink.py
+++ b/odxtools/odxlink.py
@@ -204,8 +204,10 @@ class OdxLinkDatabase:
             # locate an object exhibiting with the referenced local ID
             # in the ID database for the document fragment
             if (obj := doc_frag_db.get(ref.ref_id)) is not None:
-                if expected_type is not None:
-                    odxassert(isinstance(obj, expected_type))
+                if expected_type is not None and not isinstance(obj, expected_type):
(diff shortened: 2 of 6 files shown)


odxtools/diaglayers/protocol.py
```diff
@@ -24,14 +26,26 @@ class Protocol(HierarchyElement):
...
+    @property
+    def prot_stack_snref(self) -> str | None:
+        return self.protocol_raw.prot_stack_snref
+
     @property
     def prot_stack(self) -> ProtStack | None:
         return self.protocol_raw.prot_stack
 
+    @property
+    def parent_refs(self) -> list[ParentRef]:
```
kayoub5, line 46: Really needed after the database links has been resolved?

odxtools/diaglayers/protocol.py
```diff
@@ -24,14 +26,26 @@ class Protocol(HierarchyElement):
...
+    def comparam_spec_ref(self) -> OdxLinkRef:
+        return self.protocol_raw.comparam_spec_ref
+
     @property
     def comparam_spec(self) -> ComparamSpec:
         return self.protocol_raw.comparam_spec
 
+    @property
+    def prot_stack_snref(self) -> str | None:
+        return self.protocol_raw.prot_stack_snref
```
kayoub5, line 39: Really needed after the database links has been resolved?

odxtools/odxlink.py
```diff
@@ -204,8 +204,10 @@ def resolve(self, ref: OdxLinkRef, expected_type: Any | None = None) -> Any:
             # locate an object exhibiting with the referenced local ID
             # in the ID database for the document fragment
             if (obj := doc_frag_db.get(ref.ref_id)) is not None:
-                if expected_type is not None:
-                    odxassert(isinstance(obj, expected_type))
+                if expected_type is not None and not isinstance(obj, expected_type):
+                    odxraise(f"Referenced object for link {ref.ref_id} is of type "
+                             f"{type(obj).__name__} which is not a subclass of expected "
+                             f"type {expected_type.__name__}")
```
kayoub5, line 210: This happened?

kayoub5:
@andlaus you may want to cherry pick 7bbe195dd550bf4073e28ebe78228926181cdea6 into its own PR for faster fix to users

andlaus:
> @andlaus you may want to cherry pick 7bbe195dd550bf4073e28ebe78228926181cdea6 into its own PR for faster fix to users

I don't think it is highly critical (only medium). do you see a longer holdup of the present PR? (I'll move 7bbe195 to a separate PR if the delay is longer than this Friday...

odxtools/diaglayers/protocol.py
kayoub5: logic: provide the user with the stack itself, not just its ref

odxtools/diaglayers/protocol.py
andlaus: the "logical view" is supposed to expose *all* properties in their "final" form, i.e., if something needs to be computed -- like value inheritance -- non-raw diaglayers should expose the processed sets of objects, if there's no need for processing -- e.g., when specifying the parent layers (`parent_refs`) -- it should just forward these properties from the respective raw layers. (At least that was my philosophy when writing this code, I probably made a few mistakes besides the ones fixed by this PR...)

> not just their refs

be aware that the parent refs are a bit quirky here: they do reference resolution internally, i.e. they provide access to the respective "full" parent layer object via their `.layer` attribute.

odxtools/diaglayers/protocol.py
andlaus: why not have both? That way, raw and "full" diag layers can be used interchangeably (for code which does not use the `.diag_layer_raw` attribute)...

kayoub5: because it's already available as `prot_stack.odx_id`

andlaus: the ODX id is not exactly equivalent to a reference, e.g., the reference may feature a different set of document fragments. Also, all other classes which do reference resolution still provide the references in addition to the resolved objects..

Issue #425, opened 2025-05-08
move the root for odx id collection to `IdentifiableElement`


andlaus:
Currently, the `IdentifiableElement` class -- which represents all objects that can be referenced via the ODXLINK mechanism -- does not feature a `._build_odxlinks()` method, i.e., all classes that derive from `IdentifiableElement` need to register their ODX ids themselfs. This should be moved to `IdentifiableElement` and the derived classes should be changed to call `._build_odxlinks()` of their parent class.

PR #428, opened 2025-05-20
Implement the ECU-CONFIG category

andlaus:
This ODX category is used to hold variant coding information. Just like for the `FLASH` category, the PR is relatively large, but it does not contain much logic on the `odxtools` level.


odxtools/configiditem.py
```diff
@@ -0,0 +1,18 @@
...
+
+from .configitem import ConfigItem
+from .odxdoccontext import OdxDocContext
+from .utils import dataclass_fields_asdict
+
+
+@dataclass(kw_only=True)
+class ConfigIdItem(ConfigItem):
+    """This class represents a CONFIG-ID-ITEM."""
+
```
kayoub5, line 13: No properties?

odxtools/configrecord.py
```diff
@@ -0,0 +1,146 @@
...
+        for system_item in self.system_items:
+            result.update(system_item._build_odxlinks())
+        if self.data_id_item is not None:
+            result.update(self.data_id_item._build_odxlinks())
+        for option_item in self.option_items:
+            result.update(option_item._build_odxlinks())
+        for sdg in self.sdgs:
+            result.update(sdg._build_odxlinks())
+
+        return result
```
kayoub5, line 110: A bit off topic, but maybe `_build_odxlinks`, `_resolve_odxlinks` and `_resolve_snrefs`, can be simplified but have one function called `_get_referenceable(self) -> list[Optional[LinkableElement]]`, that return all the objects that should have `_build_odxlinks`, `_resolve_odxlinks` and `_resolve_snrefs` called on them, the rest of the logic can be implemented in the base class.

odxtools/datarecord.py
```diff
@@ -0,0 +1,73 @@
...
+@dataclass(kw_only=True)
+class DataRecord(NamedElement):
+    rule: str | None = None
+    key: str | None = None
+    data_id: IdentValue | None = None
+    sdgs: list[SpecialDataGroup] = field(default_factory=list)
+
+    # at most one of the following two attributes is not None
+    datafile: Datafile | None = None
+    data: str | None = None
```
kayoub5, line 27: Maybe add helper method that returns actual data from either `data` or `datafile` ?

odxtools/readdiagcommconnector.py
```diff
@@ -0,0 +1,86 @@
...
+    @property
+    def read_diag_comm(self) -> DiagComm | None:
+        return self._read_diag_comm
+
+    # READ-DATA cannot be resolved ahead of time as it points to a
+    # value parameter of the positive responses that are received as
+    # an answer after executing the read diag comm...
+    #@property
+    #def read_data(self) -> ValueParameter:
+    #    return self._read_data
```
kayoub5, line 35: Commented out code?

odxtools/templates/macros/printCompuMethod.xml.jinja2
```diff
@@ -110,7 +110,7 @@
 </PROG-CODE>
 {%- endmacro -%}
 
-{%- macro printCompuMethod(cm) -%}
+{%- macro printCompuMethod(cm) %}
```
kayoub5, line 113: Why?

odxtools/writepdxfile.py
```diff
@@ -225,6 +225,21 @@ def write_pdx_file(
 
             del jinja_vars["flash"]
 
+        # write the ECU-config objects
+        ecu_config_tpl = jinja_env.get_template("ecu_config.odx-e.xml.jinja2")
+        for ecu_config in database.ecu_configs:
+            zf_file_name = f"{ecu_config.short_name}.odx-e"
+            zf_file_cdate = datetime.datetime.now().strftime("%Y-%m-%dT%H:%M:%S")
```
kayoub5, line 232: What is this?

odxtools/configrecord.py
andlaus: I guess the "leafs" of `_build_odxlinks()` should be moved into `IdentifiableElement` (#425). For resolving the references, I don't really see a good alternative to the current solution. Maybe we can introduce a proxy class like [weakref.proxy](https://docs.python.org/3/library/weakref.html) but I'm not sure if this such a terribly good idea as it is quite "magical"...

odxtools/templates/macros/printCompuMethod.xml.jinja2
andlaus: the this screwed up the indentation of the XML output in some cases...

odxtools/writepdxfile.py
andlaus: this is required for the `index.xml` file of the PDX. it might be better to track the creation date with each category object of  the database, but since this information is IMO quite irrelevant (`.admin_data.doc_revisions` is a much better way to track version information), odxtools uses the time the PDX file is written as the timestamp for `index.xml`...

odxtools/dataiditem.py
```diff
@@ -0,0 +1,18 @@
...
+from xml.etree import ElementTree
+
+from .configitem import ConfigItem
+from .odxdoccontext import OdxDocContext
+from .utils import dataclass_fields_asdict
+
+
+@dataclass(kw_only=True)
+class DataIdItem(ConfigItem):
+    """This class represents a DATA-ID-ITEM."""
```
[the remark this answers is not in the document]
kayoub5: if no ConfigItem sub tag define any properties, then we can make it simply a type alias, without the extra logic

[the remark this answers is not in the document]
andlaus: `OptionItem` defines quite a few of them. I think the main point of these different item types is to express intend in different contexts, i.e. that an option, config-ID, system-ID, etc. gets specified and that only the right kind of items can be used in their respective contexts...

odxtools/readdiagcommconnector.py
andlaus: this is meant as a reminder that this should be resolved, but at load time this is impossible as far as I can see...

andlaus:
I made the CI pass again: seems like mypy 1.16 was recently released and it looks like it is quite a bit pickier with `bytes` vs `bytearray` than 1.15. (IIRC the old policy was to tread them as basically equal.)

 @kayoub5: do you see any showstoppers left in this PR?

odxtools/readdiagcommconnector.py
kayoub5: a ticket would be better reminder

andlaus: agreed. fixed: #431, https://github.com/mercedes-benz/odxtools/pull/428/commits/d3106d87e22ecd73f5557397ce4d58667e89b12a

odxtools/decodestate.py
```diff
@@ -21,7 +21,7 @@ class DecodeState:
     """Utility class to be used while decoding a message."""
 
     #: bytes to be decoded
-    coded_message: bytes
+    coded_message: bytes | bytearray
```
kayoub5, line 24: Why?

odxtools/datarecord.py
andlaus: I'm not so sure if having a non-trivial additional dependency is a good idea (this can quickly degenerate into dependency hell like most NPM and Java projects). also, `hexrec` seems to have problems with one of the datasets which have been used internally at our company for ages (it raises a `ValueError('non-contiguous data within range')` exception). `bincopy` can handle the dataset, but if fills these gaps in the record with `0xff` bytes instead of `0x00`. (That said, I'm not sure how "gaps" are supposed to be filled or if it even matters. I guess that using zeros might be advantageous, because the memory controller of the ECU's microchip might compress all-zero pages by pointing it to the same memory location...)

andlaus:
@kayoub5: I switched to using the `bincopy` module? do you see anything which should prevent merging this PR?

(BTW: ODX' process model for flashing is quite arcane IMO. Possibly we should add some additional logic to simplify user code w.r.t. filters, segments and all that jazz? something like:
```python
for address, chunk in my_flashdata.chunks:
  write_to_ecu_flash(address, chunk)
```
)

odxtools/flashdata.py
```diff
@@ -30,31 +30,33 @@ def data_str(self) -> str:
                                   f"by the {type(self).__name__} class")
 
     @property
-    def dataset(self) -> IntelHexDataSet | MotorolaSDataSet | bytearray | None:
+    def dataset(self) -> BinFile | bytearray | None:
```
kayoub5, line 33: You can create a BinFile from bytearray to have consistent return value

odxtools/datarecord.py
```diff
@@ -78,8 +80,8 @@ def blob(self) -> bytearray:
         run time when it is accessed.
```
kayoub5, line 80: What happens if the data have gaps?

odxtools/flashdata.py
andlaus: that's the purpose of the `.blob` property. `.dataset` allows you to access additional information provided by the underlying dataset, e.g. the execution start address...

PR #436, opened 2025-07-07
Implement support for the `VEHICLE-INFORMATION-SPEC` category

andlaus:
`VEHICLE-INFORMATION-SPEC` is used to describe how a given vehicle can be identified in a semi-automatic way. Also, it describes the vehicle's network topology, i.e., how communication with the vehicle's ECUs can be achieved for networking architectures involving gateways. This is described in section 7.3.10 of the ASAM version of the standard.


tests/test_vehicle_info_spec.py
```diff
@@ -0,0 +1,275 @@
...
+from odxtools.memberlogicallink import MemberLogicalLink
+from odxtools.nameditemlist import NamedItemList
+from odxtools.odxlink import DocType, OdxDocFragment
+from odxtools.oem import Oem
+from odxtools.writepdxfile import (jinja2_odxraise_helper, make_bool_xml_attrib, make_ref_attribs,
+                                   make_xml_attrib, set_category_docfrag, set_layer_docfrag)
+
+doc_frags = (OdxDocFragment(doc_name="vehicle_info_test", doc_type=DocType.VEHICLE_INFO_SPEC),)
+
+vehicle_info_spec_xml_str = """<?xml version="1.0" encoding="UTF-8" standalone="no" ?>
```
kayoub5, line 22: Better to move it to its own file

odxtools/modelyear.py
```diff
@@ -0,0 +1,17 @@
...
+from dataclasses import dataclass
+from xml.etree import ElementTree
+
+from .infocomponent import InfoComponent
+from .odxdoccontext import OdxDocContext
+from .utils import dataclass_fields_asdict
+
+
+@dataclass(kw_only=True)
+class ModelYear(InfoComponent):
```
kayoub5, line 11: This looks wrong, sub-classing InfoComponent each time, without added value

tests/test_vehicle_info_spec.py
andlaus: I'd like to keep the unit tests as self contained as possible. The tests for flashing, variant coding and multiple ecu jobs follow the same pattern..

odxtools/writepdxfile.py
```diff
@@ -256,6 +256,21 @@ def write_pdx_file(
 
             del jinja_vars["ecu_config"]
 
+        # write the vehicle info specification objects
+        vehicle_info_spec_tpl = jinja_env.get_template("vehicle_info_spec.odx-v.xml.jinja2")
+        for vehicle_info_spec in database.vehicle_info_specs:
+            zf_file_name = f"{vehicle_info_spec.short_name}.odx-v"
+            zf_file_cdate = datetime.datetime.now().strftime("%Y-%m-%dT%H:%M:%S")
```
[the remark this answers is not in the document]
kayoub5: The line `zf_file_cdate = datetime.datetime.now().strftime("%Y-%m-%dT%H:%M:%S")` can be moved out of the for loop

odxtools/modelyear.py
andlaus: the added value is that `isinstance()` can be used to distinguish these objects and thus it is much clearer to ensure that stuff like `ECU-CONFIG-REF` points to an `EcuConfig` object. (I thought about using type aliases like the derivatives of `ValidityFor`, but IMO using a separate class is better for the mentioned reasons and it should be considered to change the `ValidityFor` family to use separate classes as well.)

andlaus:
@kayoub5: do you still see any showstoppers here?

andlaus:
okay, let's merge this then. thanks for your review!

PR #446, opened 2025-09-22
Clean up the names of the load methods

andlaus:
This makes the naming more consistent and less misleading:
    
- In `Database` the method for internalizing an already parsed XML tree is named `.add_xml_tree()`, not `.load_odx_tree()`. For consistency, the method which internalizes an XML file should thus be called `.add_xml_file()` instead of `.add_odx_file()`.
- For quite some time `load_odx_d_file()` could be used to load an XML file containing any ODX category, not just ones featureing `DIAG-LAYER-CONTAINER`. To make this function name consistent with the corrosponding method of `Database`, it is renamed to `load_xml_file()`.

The old names still exist, but they are marked deprecated and are implemented as shallow shims around the new functions.

### DIFF
--- a/odxtools/__init__.py
+++ b/odxtools/__init__.py
@@ -67,6 +67,7 @@ from .loadfile import load_file as load_file  # noqa: F401
 from .loadfile import load_files  # noqa: F401
 from .loadfile import load_odx_d_file as load_odx_d_file  # noqa: F401
 from .loadfile import load_pdx_file as load_pdx_file  # noqa: F401
+from .loadfile import load_xml_file as load_xml_file  # noqa: F401
 from .version import __version__ as __version__  # noqa: F401
 from .writepdxfile import write_pdx_file as write_pdx_file  # noqa: F401
 
--- a/odxtools/loadfile.py
+++ b/odxtools/loadfile.py
@@ -2,6 +2,8 @@
 import os
 from pathlib import Path
 
+from deprecation import deprecated
+
 from .database import Database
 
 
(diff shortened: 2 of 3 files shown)


odxtools/loadfile.py
```diff
@@ -12,19 +14,24 @@ def load_pdx_file(pdx_file: str | Path) -> Database:
     return db
 
 
-def load_odx_d_file(odx_d_file_name: str | Path) -> Database:
+def load_xml_file(xml_file_name: str | Path) -> Database:
```
kayoub5, line 17: The index file in a PDX is XML, but we don't support loading it

odxtools/__init__.py
```diff
@@ -67,6 +67,7 @@
 from .loadfile import load_files  # noqa: F401
 from .loadfile import load_odx_d_file as load_odx_d_file  # noqa: F401
 from .loadfile import load_pdx_file as load_pdx_file  # noqa: F401
+from .loadfile import load_xml_file as load_xml_file  # noqa: F401
```
kayoub5, line 70: For end-user, they only need `load_file` which does not seem to be already exposed, no need for `load_pdx_file`, `load_pdx_file` and `load_odx_d_file`

odxtools/loadfile.py
andlaus: yes, but `Database.add_odx_tree()` would be weird IMO, (also, in principle any file can be an "ODX" file if it is an auxiliary file.)

Issue #421, opened 2025-04-29
Fix object lifetime management


andlaus:
The odxtools database object currently seem to mess up python's garbage collector: when running
```python
import odxtools
import time

print("loading database", flush=True)
db = odxtools.load_file("my_big_pdx_file.pdx")
print("database loaded", flush=True)
time.sleep(20)
print("deleting database", flush=True)
del db
print("database deleted", flush=True)
time.sleep(20)
print("done", flush=True)
```

the memory used by the `db` object is not freed after the "database deleted" string is printed but only at program exit, i.e., after "done" appears.

I suspect that this is due to python's garbage collector getting confused by cyclic references, but unfortunately I have not found a good way to debug this, and my limited experiments with weak references from python's `weakref` module also did not fix the issue. @kayoub5, @nada-ben-ali, @zariiii9003: any ideas how to get to the bottom of this?

kayoub5:
Those are common mis-conception about garbage collectors:
* Garbage collocation does not run when you are done with the object, quite the opposite, most GCs (I have seen) run when the program ask to allocate more memory, more on why below.
* Some GCs have "scope exit" logic, basically GC auto run on local objects of functions for example, since are those are easier to detect and clean,  your code have global module variables, GC typically don't aggressively run until module logic is over.
* GCs usually require putting the entire application/thread in a pause state in order to run, so typical imlimentations avoid interrupting your code as much as possible (usually wait for a sleep/ io operation), unless RAM is very limited.
* A process does not return the memory it longer needs to the os, the reason is that "it may need it again soon", and requesting more RAM from os is slow (in some cases, running a GC is faster than asking os for RAM). so process just mark RAM as "os free to take it back if it really needs to, but I prefer to keep it if possible".

andlaus:
Isn't python's main memory reference counting? I'm not sure how exactly it is implemented in cpython, but the other refcounting implementation I have experience with (C++'s smart pointers) frees memory of objects which's reference count reaches zero immediately. (This is also a very fast operation...)

> lambda functions keep references to objects simply because the lambda code uses them, which makes them hard to notice.

sure, but as soon as the refcount of the lambda function reaches zero, it should be deleted...

kayoub5:
A memory freed does not mean the process memory goes down immediately, a process always keep the memory pages for fast re-allocation.

andlaus:
> A memory freed does not mean the process memory goes down immediately, a process always keep the memory pages for fast re-allocation.

[...]

PR #447, opened 2025-09-22
Weakrefs

andlaus:
Weak references can now be used (enabled by default): This hopefully avoids cycles in the reference graph thus making odxtools objects play nicely with reference counting. As a consequence, when deleting a database object using `del odx_db_obj`, the memory used by the object seems to be freed immediately (in contrast to after calling `gc.collect()`) and object deletion followed by `gc.collect()` experiences a speed up of factor 4. Also, unexpected pauses in program execution due to python garbage collecting orphaned `Database` objects when it feels like it should be a thing of the past.

This PR fixes https://github.com/mercedes-benz/odxtools/issues/421 and due to a ordering depencency is based atop #446.


odxtools/odxlink.py
```diff
@@ -172,18 +173,31 @@ class OdxLinkDatabase:
     This can resolve ODXLINK references.
     """
 
-    def __init__(self) -> None:
+    def __init__(self, *, use_weakrefs: bool = False) -> None:
```
kayoub5, line 176: Avoid default value here, otherwise you miss a case where the value is not propagated from database correctly.

odxtools/odxlink.py
```diff
@@ -172,18 +173,31 @@ class OdxLinkDatabase:
...
         self._db: dict[OdxDocFragment, dict[str, Any]] = {}
+        self.use_weakrefs = use_weakrefs
 
     @overload
-    def resolve(self, ref: OdxLinkRef, expected_type: None = None) -> Any:
+    def resolve(self,
+                ref: OdxLinkRef,
+                expected_type: None = None,
+                *,
+                use_weakrefs: bool | None = None) -> Any:
```
kayoub5, line 185: Already in init, why in resolve as well?

odxtools/snrefcontext.py
```diff
@@ -20,6 +20,7 @@ class SnRefContext:
     to be resolved
     """
 
+    use_weakrefs: bool = True
```
kayoub5, line 23: Avoid default value, otherwise you miss case where it should be specified

odxtools/odxlink.py
andlaus: because I frequently use the `OdxLink` database to resolve references in my application code and when I wrote this code I expected to get normal references, i.e., it is sometimes necessary to change the default value of the `OdxLinkDatabase` object...

kayoub5: a new `clone(self, use_weakrefs)` function would be better in that case

odxtools/odxlink.py
andlaus: e.g. when you want to restrict resolving to a given ECU variant.

odxtools/nameditemlist.py
```diff
@@ -212,9 +213,15 @@ def _get_item_key(self, item: T) -> str:
         such short names.
 
         """
-        if not isinstance(item, OdxNamed):
+        if isinstance(item, (weakref.ProxyType, weakref.CallableProxyType)):
+            if not isinstance(item.__repr__.__self__, OdxNamed):  # type: ignore[attr-defined]
```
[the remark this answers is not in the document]
andlaus: good point. [done](https://github.com/mercedes-benz/odxtools/pull/447/commits/1ac4f991c02ef599f8a1101f6b02f582d6dc9497)

kayoub5:
> @kayoub5 do you see any showstoppers for merging here (i.e., stuff which cannot be fixed later)?

No

PR #456, opened 2026-05-05
update CONTRIBUTING.md to use the DCO approach

andlaus:
this updates the boiler plate files to the things which have changed over the last few years: i.e., the contributer's license agreement has been replaced by the developer certificate of origin approach (*yay*) and `SECURITY.md` file has become obsolete.


andlaus:
since this concerns Mercedes-Benz internal stuff, there's IMO not much point in reviewing this (you are very much invited to do so if you want), so I it merge directly...

PR #459, opened 2026-05-11
make it work for DIAG-SERVICEs without REQUESTs

### LINKED ISSUE #458: DIAG-SERVICE without REQUEST-REF causes crash
When an ODX file contains a DIAG-SERVICE without a REQUEST-REF, odxtools crashes during parsing.

Although REQUEST-REF is optional according to the ODX specification, the current implementation assumes it is always present and does not handle the missing case.

The following exceptions is thrown:
 File "C:\Users\XXXXXXX\AppData\Local\Programs\Python\Python312\Lib\site-packages\odxtools\odxlink.py", line 225, in resolve
    for ref_frag in reversed(ref.ref_docs):
                             ^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'ref_docs'

andlaus:
(this is allowed by the spec, but IMO its akin to an airplane without wings...)

this fixes https://github.com/mercedes-benz/odxtools/issues/458 .

### DIFF
--- a/odxtools/diagservice.py
+++ b/odxtools/diagservice.py
@@ -27,7 +27,7 @@ class DiagService(DiagComm):
     """
 
     comparam_refs: list[ComparamInstance] = field(default_factory=list)
-    request_ref: OdxLinkRef
+    request_ref: OdxLinkRef | None
     pos_response_refs: list[OdxLinkRef] = field(default_factory=list)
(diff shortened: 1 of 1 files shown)


andlaus:
little point in a review here. let's merge...

Big621 (commented):
🔥

PR #461, opened 2026-06-02
feat(compare): add diagnostic comparison tool with export support

seif-galal:
# PR #461: Refactor and Enhance ODX Diagnostic Comparison Tool

## Overview
This PR refactors and enhances the existing diagnostic comparison functionality in odxtools, introducing structured export formats, intelligent service rename detection, CI/CD integration features, and improved code organization.

### Refactoring
- Consolidates existing comparison logic into dedicated DiffEngine class
- Separates display/output logic into ReportFormatter class
- Introduces Exporter abstract base class with format-specific implementations

### New Features

**Structured Export Formats**
- JSON export for programmatic analysis
- YAML export for human-readable configuration
- CSV export for spreadsheet analysis
- Enhanced text export with full detail section

**Intelligent Service Rename Detection**
- Identifies renamed services by comparing request prefix + response names
- More reliable than simple name matching (handles refactoring)

**CI/CD Integration**
- `--fail-on-diff` flag returns exit code 1 when differences detected
- `--ignore` regex patterns to filter noisy attributes
- `--output-file` for automated report generation

**Developer Experience**
- `--profile` flag for performance timing
- Clear error messages and graceful handling of missing variants

**Change Score Metric**
- Quantitative assessment: New + Deleted + Renamed + Changed Parameters

**Optional Dependencies**
- PyYAML for YAML export support (`pip install odxtools[compare-tool]`)

## Engineering Context
- Processes ODX/PDX diagnostic data
- Full type safety with mypy, ruff, yapf

## Usage Examples

Compare two ECU variants:
```bash
odxtools compare database.pdx -v variant1 variant2

PR #469, opened 2026-06-19
feat(compare): add parallel processing support

seif-galal:
### Overview
This PR adds parallel processing support to the `odxtools compare` command, significantly reducing comparison time for large PDX files by processing diagnostic layers concurrently across multiple CPU cores.

### New Features
- `--parallel`: Enable parallel processing for large PDX files
- `--max-workers N`: Control worker count (auto-detects CPU cores by default)
- Progress callback showing layer completion during comparison

### Files Changed
- `odxtools/cli/compare_parallel.py` (new file, 200+ lines)

### Quality Assurance
| Check | Status |
|-------|--------|
| Syntax | ✅ PASS |
| Ruff | ✅ PASS |
| Mypy | ✅ PASS |
| YAPF | ✅ PASS |

### Related
- Part of the compare tool enhancement series
- Follows PR #461 (base comparison tool)

### Usage Examples
```bash
# Compare with parallel processing
odxtools compare baseline.pdx -db new.pdx --parallel

# Control worker count
odxtools compare baseline.pdx -db new.pdx --parallel --max-workers 4

# Verbose mode with progress
odxtools compare baseline.pdx -db new.pdx --parallel --verbose

### DIFF
--- a/odxtools/cli/compare_parallel.py
+++ b/odxtools/cli/compare_parallel.py
@@ -1,0 +1,129 @@
+#!/usr/bin/env python
+# SPDX-License-Identifier: MIT
+"""
+Parallel processing module for ODX diagnostic comparison.
+
+This module provides parallel comparison capabilities for large PDX files,
+significantly reducing comparison time by processing diagnostic layers
+concurrently across multiple CPU cores.
+"""
+
+import concurrent.futures
+import logging
+from collections.abc import Callable
+from multiprocessing import cpu_count
+
+from ..database import Database
+from .compare import DiffEngine, SpecsChangesVariants, DiagLayerChanges
+
+logger = logging.getLogger(__name__)
+
+
+def compare_databases_parallel(
+    engine: DiffEngine,
+    database_new: Database,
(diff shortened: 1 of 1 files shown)


andlaus:
thanks for your effort, but I don't think that we should trade slightly better runtime performance for significantly more complicated and less maintainable code. IOW, what are the use cases for this? comparing datasets with many layers should be relatively rare and if you have many PDX files which need to be compared pairwise, you can simply call `odxtools compare` in multple processes...

seif-galal:
> thanks for your effort, but I don't think that we should trade slightly better runtime performance for significantly more complicated and less maintainable code. IOW, what are the use cases for this? comparing datasets with many layers should be relatively rare and if you have many PDX files which need to be compared pairwise, you can simply call `odxtools compare` in multple processes...

We can discuss this concept with the Other PRs when #475 is merged

The main objective of this PR is to substatially reduce runtime,to maintain simplicity i will close it and we will discuss it later if we need it again.

PR #473, opened 2026-06-30
`parameter_info()`: handle `Structure` DOPs

andlaus:
This is a small but relevant quality of life improvement for services which use `Structure` objects as the DOP for parameters...

### DIFF
--- a/odxtools/parameterinfo.py
+++ b/odxtools/parameterinfo.py
@@ -30,6 +30,7 @@ from .parameters.tablekeyparameter import TableKeyParameter
 from .parameters.tablestructparameter import TableStructParameter
 from .paramlengthinfotype import ParamLengthInfoType
 from .staticfield import StaticField
+from .structure import Structure
 
 
 def _get_linear_segment_info(segment: LinearSegment) -> str:
@@ -153,6 +154,11 @@ def parameter_info(param_list: Iterable[Parameter], quoted_names: bool = False) 
                 else:
                     dtc_desc = dtc.text and f" (\"{dtc.text}\")"
                     of.write(f"  0x{dtc.trouble_code:06x}{dtc_desc}\n")
+            continue
+        elif isinstance(dop, Structure):
+            of.write(f"{q}{param.short_name}{q}: {{\n")
+            of.write(textwrap.indent(parameter_info(dop.parameters, True), "  "))
+            of.write(f"}}\n")
             continue
         elif isinstance(dop, Multiplexer):
             of.write(f"{q}{param.short_name}{q}: ")
--- a/odxtools/parameterinfo.py
+++ b/odxtools/parameterinfo.py
@@ -175,8 +175,9 @@ def parameter_info(param_list: Iterable[Parameter], quoted_names: bool = False) 
         elif isinstance(dop, EnvironmentDataDescription):
             dtc_ref = dop.param_snref or dop.param_snpathref or "<unknown>"
             of.write(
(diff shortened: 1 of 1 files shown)


andlaus:
Note: I added handling `EnvironmentDataDescription` DOPs...

andlaus:
@kayoub5: any objections/comments?

odxtools/parameterinfo.py
```diff
@@ -165,6 +172,40 @@ def parameter_info(param_list: Iterable[Parameter], quoted_names: bool = False)
                     of.write(textwrap.indent(parameter_info(struc.parameters, True), "    "))
                 of.write(f"   }})\n")
             continue
+        elif isinstance(dop, EnvironmentDataDescription):
+            dtc_ref = dop.param_snref or dop.param_snpathref or "<unknown>"
+            of.write(
+                f"{q}{param.short_name}{q}: environment data description; DTC parameter: '{dtc_ref}': {{\n"
```
kayoub5, line 178: Not sure why with `EnvironmentDataDescription` refs are printed, for every other type, only the resolved data is printed

odxtools/parameterinfo.py
```diff
@@ -165,6 +172,40 @@ def parameter_info(param_list: Iterable[Parameter], quoted_names: bool = False)
...
+                of.write(f"  parameters required for all DTCs:\n")
+                of.write(textwrap.indent(parameter_info(env_data.parameters, True), "    "))
+
+            # then print the DTC-specific parts
+            for env_data in dop.env_datas:
+                if env_data.all_value:
+                    continue
+
+                if env_data.dtc_values:
+                    if len(env_data.dtc_values) == 1:
```
kayoub5, line 197: Why the special logic for length 1 ?, `','.join` would give same result for one element list

odxtools/parameterinfo.py
```diff
@@ -165,6 +172,40 @@ def parameter_info(param_list: Iterable[Parameter], quoted_names: bool = False)
...
+            )
+
+            # first, print the environment datas which are always send
+            # (the ODX standard mandates them to be send before the
+            # DTC-specific ones)
+            for env_data in dop.env_datas:
+                if not env_data.all_value:
+                    continue
+
+                of.write(f"  parameters required for all DTCs:\n")
```
kayoub5, line 188: This line will get printed for each env_data

odxtools/parameterinfo.py
andlaus: the problem is that the `param_snref` cannot be resolved ahead of time because the DOP can be featured by multiple parameters for multiple services. in the `parameter_info()` function we could actually resolve the SNREF (because we know the actual list of parameters), but if we only use it to print the parameter's `short_name`, we can equally well just print the SNREF (SNPATHREFs are currently not supported, but I thought it might be a good idea to include it anyway...)

odxtools/parameterinfo.py
andlaus: the point is actually the proper pluralization of "DTC": `"DTC"` instead of `"DTCs"` if there is only a single DTC for a given `EnvironmentData`

odxtools/parameterinfo.py
andlaus: for each which has the `.all_value` attribute set to `True`. IIRC the standard allows at most one such `EnvironmentData` per `EnvironmentDataDescription`. We could break out of the loop here, but if the dataset is not standard compliant in this way, all the "global" parameters will be printed...

kayoub5 (approved):

PR #477, opened 2026-07-16
make `pyright` happy

andlaus:
`pyright` sometimes flags different issues than `mypy` and having both in their happy zone is thus a good thing. Note that this requires a few `# pyright: ignore` comments where `mypy` and `pyright` are mutually exclusive.

Finally, as a small drive-by, this PR reverts cf1b9cb307043418378fdcc7d882f0b973004128 because both mypy and pyright can infer that `out_param_snpathref_el` cannot be `None` at this place...

### DIFF
--- a/odxtools/dataobjectproperty.py
+++ b/odxtools/dataobjectproperty.py
@@ -120,7 +120,8 @@ class DataObjectProperty(DopBase):
         if not isinstance(physical_value, (int, float, str, BytesTypes)):
             odxraise(f"Invalid type '{type(physical_value).__name__}' for physical value. "
                      f"(Expect atomic type!)")
-        internal_value = self.compu_method.convert_physical_to_internal(physical_value)
+        internal_value = self.compu_method.convert_physical_to_internal(
+            physical_value)  # pyright: ignore[reportArgumentType]
         self.diag_coded_type.encode_into_pdu(internal_value, encode_state)
 
     def decode_from_pdu(self, decode_state: DecodeState) -> ParameterValue:
--- a/odxtools/dataobjectproperty.py
+++ b/odxtools/dataobjectproperty.py
@@ -14,7 +14,7 @@ from .exceptions import EncodeError, odxraise, odxrequire
 from .internalconstr import InternalConstr
 from .odxdoccontext import OdxDocContext
 from .odxlink import OdxLinkDatabase, OdxLinkId, OdxLinkRef
-from .odxtypes import AtomicOdxType, BytesTypes, ParameterValue
+from .odxtypes import AtomicOdxType, ParameterValue
 from .physicaltype import PhysicalType
 from .snrefcontext import SnRefContext
 from .unit import Unit
@@ -117,11 +117,11 @@ class DataObjectProperty(DopBase):
                 f"The value {repr(physical_value)} of type {type(physical_value).__name__}"
                 f" is not a valid.")
 
-        if not isinstance(physical_value, (int, float, str, BytesTypes)):
+        if not isinstance(physical_value, AtomicOdxType):
             odxraise(f"Invalid type '{type(physical_value).__name__}' for physical value. "
                      f"(Expect atomic type!)")
-        internal_value = self.compu_method.convert_physical_to_internal(
-            physical_value)  # pyright: ignore[reportArgumentType]
+            return
+        internal_value = self.compu_method.convert_physical_to_internal(physical_value)
         self.diag_coded_type.encode_into_pdu(internal_value, encode_state)
 
     def decode_from_pdu(self, decode_state: DecodeState) -> ParameterValue:
--- a/odxtools/decodestate.py
+++ b/odxtools/decodestate.py
@@ -118,12 +118,16 @@ class DecodeState:
             str_encoding = get_string_encoding(base_data_type, base_type_encoding,
                                                is_highlow_byte_order)
             if str_encoding is not None:
+                if not isinstance(raw_value, (bytes, bytearray)):
+                    odxraise(f"Expected bytes for string decoding, got {type(raw_value).__name__}")
                 internal_value = raw_value.decode(str_encoding, errors=text_errors)
             else:
                 internal_value = "ERROR"
 
         # ... signed integers, ...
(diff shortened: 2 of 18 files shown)


kayoub5:
Personally, I use https://docs.astral.sh/ty/ , less false positives and faster

odxtools/dataobjectproperty.py
```diff
@@ -120,7 +120,8 @@ def encode_into_pdu(self, physical_value: ParameterValue, encode_state: EncodeSt
         if not isinstance(physical_value, (int, float, str, BytesTypes)):
             odxraise(f"Invalid type '{type(physical_value).__name__}' for physical value. "
                      f"(Expect atomic type!)")
-        internal_value = self.compu_method.convert_physical_to_internal(physical_value)
+        internal_value = self.compu_method.convert_physical_to_internal(
+            physical_value)  # pyright: ignore[reportArgumentType]
```
kayoub5: why would this get flagged?

odxtools/decodestate.py
```diff
@@ -160,6 +164,8 @@ def extract_atomic_value(
 
         # ... unsigned integers, ...
         elif base_data_type == DataType.A_UINT32:
+            if not isinstance(raw_value, int):
+                odxraise(f"Expected integer raw value for A_UINT32, got {type(raw_value).__name__}")
             if base_type_encoding == Encoding.BCD_P:
                 internal_value = self.__decode_bcd_p(raw_value)
             elif base_type_encoding == Encoding.BCD_UP:
```
kayoub5, line 171: I have seen those getting used in byte arrays

odxtools/environmentdatadescription.py
```diff
@@ -189,7 +189,9 @@ def decode_from_pdu(self, decode_state: DecodeState) -> ParameterValue:
                 tmp = env_data.decode_from_pdu(decode_state)
                 if not isinstance(tmp, dict):
                     odxraise()
-                result.update(tmp)
+                    break
+                result.update(tmp)  # pyright: ignore[reportArgumentType,reportCallIssue]
```
kayoub5, line 193: Did the code flow change here, or just me?

odxtools/environmentdatadescription.py
```diff
@@ -199,7 +201,9 @@ def decode_from_pdu(self, decode_state: DecodeState) -> ParameterValue:
...
+
```
kayoub5, line 206: Did the code flow change here, or just me?

odxtools/multiplexer.py
```diff
@@ -167,6 +167,7 @@ def encode_into_pdu(self, physical_value: ParameterValue, encode_state: EncodeSt
         elif case_spec is None:
             if self.default_case is None:
                 raise EncodeError(f"Multiplexer {self.short_name} does not define a default case")
+            mux_case = self.default_case
```
kayoub5, line 170: what is this?

odxtools/dataobjectproperty.py
andlaus: pyright produces this:
```
> pyright .
odxtools/dataobjectproperty.py
  odxtools/dataobjectproperty.py:124:13 - error: Argument of type "ParameterValue" cannot be assigned to parameter "physical_value" of type "AtomicOdxType" in function "convert_physical_to_internal"
    Type "ParameterValue" is not assignable to type "AtomicOdxType"
      Type "DiagnosticTroubleCode" is not assignable to type "AtomicOdxType"
        "DiagnosticTroubleCode" is not assignable to "str"
        "DiagnosticTroubleCode" is not assignable to "int"
        "DiagnosticTroubleCode" is not assignable to "float"
        "DiagnosticTroubleCode" is not assignable to "bytearray"
        "DiagnosticTroubleCode" is not assignable to "bytes"
        "DiagnosticTroubleCode" is incompatible with protocol "SupportsBytes"
    ... (reportArgumentType)
1 error, 0 warnings, 0 informations
```

odxtools/environmentdatadescription.py
andlaus: yes, it changed, but only for the non-strict mode in case of an error...

odxtools/environmentdatadescription.py
andlaus: this only affects the non-strict mode in case of an error: hopefully making the code not trip with another exception.

odxtools/multiplexer.py
andlaus: If no cases were defined, the default case is used. (else pyright complains that `mux_case` can be undefined.) This is a minor fix for a corner case. (IMO a multiplexer without cases is a bit like a piano without keys...)

andlaus: actually, since `case_spec`  is the case selected by the user. it more like playing the piano without pressing a key and thus listening to the woodworms ;)

andlaus:
> Personally, I use https://docs.astral.sh/ty/ , less false positives and faster

I will have a go at `ty` once this one is in...

odxtools/decodestate.py
andlaus: did you have any chance to test if the BCD stuff in odxtools is correct?

kayoub5: Sorry, no

I only cared that files using BCD loaded correctly

odxtools/dataobjectproperty.py
kayoub5: The warning is kind of correct, since ParameterValue can be dict/list, but AtomicOdxType is only for primitive data types, a small `isintance` check would make it happy

odxtools/decodestate.py
andlaus: ok, we're surely correct since we have a  [unit test](https://github.com/mercedes-benz/odxtools/blob/main/tests/test_encoding.py#L150-L194) for this after all *g*

odxtools/dataobjectproperty.py
andlaus: I see. fixed: 354acf80de94aa4c7dc337c9719691d3ac9d6f30 , thanks

kayoub5 (approved):

PR #478, opened 2026-07-17
Fix the `browse` tool

andlaus:
since the `browse` tool is not used very much and it is difficult to unit test due to its interactive nature, it atrophied quite a bit over the years. This pull request should make it functional again...

### DIFF
--- a/odxtools/cli/browse.py
+++ b/odxtools/cli/browse.py
@@ -5,6 +5,7 @@ import sys
 from typing import cast
 
 from InquirerPy.resolver import prompt as IP_prompt
+from rich import print as rich_print
 
 from ..complexdop import ComplexDop
 from ..database import Database
@@ -133,7 +134,7 @@ def encode_message_interactively(codec: Request | Response,
         }]
         answer = IP_prompt(answered_request_prompt)
         answered_request = cast(bytes, answer.get("request"))
-        print(f"Input interpretation as list: {list(answered_request)}")
+        rich_print(f"Input interpretation as list: {list(answered_request)}")
 
     has_settable_param = False
     for param in codec.parameters:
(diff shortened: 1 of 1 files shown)


odxtools/cli/browse.py
```diff
@@ -5,6 +5,7 @@
 from typing import cast
 
 from InquirerPy.resolver import prompt as IP_prompt
+from rich import print as rich_print
```
kayoub5, line 8: Easier use
```suggestion
from rich import print
```

odxtools/cli/browse.py
```diff
@@ -92,25 +93,37 @@ def prompt_single_parameter_value(parameter: Parameter) -> AtomicOdxType | None:
...
-    if (dop := getattr(parameter, "dop", None)) and \
-       (compu_method := getattr(dop, "compu_method", None)):
-        scales = compu_method.internal_to_phys
-        choices = [scale.compu_const for scale in scales if scale is not None]
-        if (cdv := compu_method.compu_default_value) is not None:
-            choices.append(cdv.compu_const)
-        param_prompt[0]["choices"] = choices
+    if (dop := getattr(parameter, "dop", None)) is not None and \
+       (compu_method := getattr(dop, "compu_method", None)) is not None and \
+       (citp := getattr(compu_method, "compu_internal_to_phys", None)) is not None:
```
kayoub5, line 98: that looks like a big hack, that should be replaced with `isinstance`

odxtools/cli/browse.py
andlaus: I prefer immediately seeing that we're not using the normal print. (this the same is approach as in `compare`.)

odxtools/cli/browse.py
andlaus: the problem is that not all DOPs have a `compu_method`, and IIRC not all compumethods have `compu_internal_to_phys` attributes. finding the exhaustive list of which ones do is some effort and -- more importantly -- I think that `isinstance(x, Foo); cm = x.compu_method` conveys the intention of the code (slightly) less clearly...

kayoub5 (approved):

PR #481, opened 2026-07-20
correct file extension and fix resource leak

seif-galal:
- Fix copy-paste error: function dictionaries were written with .odx-v instead of .odx-fd extension
- Fix resource leak when copying template files into PDX archive


andlaus:
while I don't agree that the unclosed file descriptor is a resource leak (it gets deleted when it runs out of scope), I don't mind doing it the way you proposed. (the copy-and-pasto is clearly a mistake.)

thanks for the patch!

PR #467, opened 2026-06-12
Add graceful fallback for missing InquirerPy dependency

seif-galal:
## Description

Wraps the InquirerPy import in `try/except ImportError` so `browse.py` can be imported without the optional dependency installed. Adds explicit `SystemError` guards before each interactive prompt call with a clear install message.

## Changes

- Replace unconditional `import InquirerPy.prompt` with `try/except ImportError` fallback to `None`
- Add `None` checks before all 6 `IP_prompt()` calls
- Remove unused `# type: ignore[assignment]` comment

## Behavior

**No functional changes when InquirerPy is installed.**

When missing, users now see:
SystemError: InquirerPy is required for interactive mode. Install: pip install InquirerPy

Instead of a cryptic:
TypeError: 'NoneType' object is not callable

PR #482, opened 2026-07-20
fix(cli): improve error messages for unavailable tools

seif-galal:
When a tool fails to load due to a missing optional dependency, print a user-friendly message with the correct pip install command instead of a raw exception traceback.

- Map known missing modules to their odxtools extra install hints
- Keep full traceback in --help description for debugging
- Clean up run() output to stderr for better UX
- this PR addresses the suggestion from PR #467


seif-galal:
reopening later due to issues

PR #485, opened 2026-07-22
preserve correct MIME types for ODX files

seif-galal:
Fix logic error where the correct ASAM MIME type for .odx-d and .odx-cs files was immediately overwritten with application/octet-stream.

The condition 'if mime_type is None and guessed_mime_type is not None' evaluated to False when mime_type was already set (e.g., to the correct ASAM type), causing the else branch to clobber it.

Change to nested if: only fall back to guessed or octet-stream when the ASAM-specific type was not set.


seif-galal:
@andlaus back to the writepdxfile.py file, i spotted a mime type error which this PR solves, feel free to check it out in the diff

andlaus:
looks good. thanks for the catch!

PR #487, opened 2026-07-25
fix: do not test ElementTree elements for truthiness

### LINKED ISSUE #419: `xml.etree.ElementTree` Deprecation Warnings
I noticed the following deprecation warning when using odxtools with Python 3.12:

```
odxtools/specialdatagroup.py:25: DeprecationWarning: Testing an element's truth value will always return True in future versions.  Use specific 'len(elem)' or 'elem is not None' test instead.
    if caption_elem := et_element.find("SDG-CAPTION"):
```

which is caused by the following line:

https://github.com/mercedes-benz/odxtools/blob/a602a79e216e8c5e538110ce2048d1a3d08a8ac1/odxtools/specialdatagroup.py#L25

This can be fixed as follows:

```diff
-         if caption_elem := et_element.find("SDG-CAPTION"):
+         if (caption_elem := et_element.find("SDG-CAPTION")) is not None:
```

There are at least two more similar issues:

https://github.com/mercedes-benz/odxtools/blob/a602a79e216e8c5e538110ce2048d1a3d08a8ac1/odxtools/subcomponentparamconnector.py#L46

and

https://github.com/mercedes-benz/odxtools/blob/a602a79e216e8c5e538110ce2048d1a3d08a8ac1/odxtools/subcomponentparamconnector.py#L55

which can be fixed similarly, e.g.:

```diff
-        for elem in et_element.find("IN-PARAM-IF-REFS") or []:
+        for elem in [] if (in_elems := et_element.find("IN-PARAM-IF-REFS")) is None else in_elems:
```

Sanjays2402:
Closes #419

An `Element`'s truth value is deprecated since Python 3.12, and it is also `False` for elements without children — so `et_element.find(...) or <fallback>` both warns and silently discards a found-but-childless element. Replaced the three remaining occurrences with explicit `is None` checks.

The new test asserts that parsing these elements emits no `DeprecationWarning`; it fails 2/2 without the source change and passes with it. Full test suite stays green (157 passed).

### DIFF
--- a/odxtools/dynamicendmarkerfield.py
+++ b/odxtools/dynamicendmarkerfield.py
@@ -39,7 +39,9 @@ class DynamicEndmarkerField(Field):
 
         # ODX 2.0 uses DATA-OBJECT-PROP-REF
         # ODX 2.2 uses DYN-END-DOP-REF
-        dop_ref = et_element.find("DYN-END-DOP-REF") or et_element.find("DATA-OBJECT-PROP-REF")
+        dop_ref = et_element.find("DYN-END-DOP-REF")
+        if dop_ref is None:
+            dop_ref = et_element.find("DATA-OBJECT-PROP-REF")
         dyn_end_dop_ref = DynEndDopRef.from_et(odxrequire(dop_ref), context)
 
         return DynamicEndmarkerField(dyn_end_dop_ref=dyn_end_dop_ref, **kwargs)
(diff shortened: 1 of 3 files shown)


andlaus:
thanks for your contribution. unfortunately, it seems like the timing collided with #488 which, amongst others,  fixed this issue. I'll close the PR for now, but don't let yourself be demotivated from opening new ones!

PR #488, opened 2026-07-27
Miscellaneous fixes

andlaus:
This PR contains a smattering of small but important fixes which Claude Code found while increasing the test coverage. (Maybe I'll revise some of the unit tests which it produced and open PRs for them in the future. Let's see...)

### DIFF
--- a/odxtools/commrelation.py
+++ b/odxtools/commrelation.py
@@ -29,7 +29,7 @@ class CommRelation:
 
     @property
     def diag_comm(self) -> DiagComm:
-        return self._diag_comm
+        return odxrequire(self._diag_comm)
 
     @property
     def in_param_if(self) -> Parameter | None:
@@ -90,6 +90,7 @@ class CommRelation:
         return {}
 
     def _resolve_odxlinks(self, odxlinks: OdxLinkDatabase) -> None:
+        self._diag_comm = None
         if self.diag_comm_ref is not None:
             self._diag_comm = odxlinks.resolve(self.diag_comm_ref, DiagComm)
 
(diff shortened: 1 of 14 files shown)


odxtools/commrelation.py
```diff
@@ -103,6 +103,10 @@ def _resolve_snrefs(self, context: SnRefContext) -> None:
                 DiagComm,
                 use_weakrefs=context.use_weakrefs)
 
+        if not hasattr(self, "_diag_comm"):
```
kayoub5: why not just use `self._diag_comm` ? or set it to None initially ?

odxtools/compumethods/scalelinearcompumethod.py
```diff
@@ -107,7 +107,7 @@ def __post_init__(self) -> None:
             # reference point must be identical
             y0 = s0.convert_internal_to_physical(x)
             y1 = s1.convert_internal_to_physical(x)
-            if abs(y0 - y1) < 1e-10:
+            if abs(y0 - y1) > 1e-10:
```
kayoub5: Use `math.isclose` https://docs.python.org/3/library/math.html#math.isclose

andlaus: fixed: f3ab5a6084d53a28c283d913420636c6aaa00750

odxtools/commrelation.py
andlaus: the problem is that at this point, `self._diag_comm` may not exist yet (if the dataset is not standard compliant.) Anyway, I [fixed](https://github.com/mercedes-benz/odxtools/pull/488/changes/eeaf5afb7261d0e583c33f6c9a816840d62d16b8) it a nicer way..

kayoub5 (approved):

PR #489, opened 2026-07-28
perf(writepdxfile): cache Jinja2 environment to avoid re-parsing temp…

seif-galal:
Creating a new jinja2.Environment on every write_pdx_file() call forces
Jinja2 to re-parse all templates from disk each time. Since the
templates are static files that never change at runtime, this is pure
overhead.

Cache the environment using @functools.cache. The call-specific globals
(set_category_docfrag, set_layer_docfrag, make_ref_attribs) are still
set fresh per call since they close over jinja_vars.

Benchmark: 50 consecutive PDX writes
  Before: 47.50s
  After:  2.9s

Benchmark: 500 consecutive PDX writes
  Before: 465.3s
  After:  29.9s

Per-write performance with repeated use:
  Cold:  85.4ms (10.9x faster than original)
  Warm:  72.6ms (12.8x faster than original)
  Hot:   56.7ms (16.4x faster than original)

### DIFF
--- a/odxtools/writepdxfile.py
+++ b/odxtools/writepdxfile.py
@@ -15,6 +15,7 @@ import odxtools
 from .database import Database
 from .odxlink import DocType, OdxDocFragment, OdxLinkRef
 from .odxtypes import bool_to_odxstr
+from functools import cache
 
 
 def jinja2_odxraise_helper(msg: str) -> None:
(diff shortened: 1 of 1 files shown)


seif-galal:
@andlaus its ready check the diff

odxtools/writepdxfile.py
```diff
@@ -72,6 +72,21 @@ def make_ref_attribs(jinja_vars: dict[str, Any], ref: OdxLinkRef) -> str:
 assert isinstance(__module_filename, str)
 __templates_dir = os.path.sep.join([os.path.dirname(__module_filename), "templates"])
 
+_jinja_env_cache: dict[str, jinja2.Environment] = {}
+
+
+def _get_jinja_env(templates_dir: str) -> jinja2.Environment:
```
kayoub5, line 78: Use `@functools.cache`

https://docs.python.org/3/library/functools.html#functools.cache

odxtools/writepdxfile.py
kayoub5, line 162: This change looks unrelated to PR topic

seif-galal: yes it was merged in my last PR ignore it. i didnt sync my fork and thats why its there

kayoub5: Please rebase your branch then

seif-galal:
PR description updated accordingly as well.

andlaus:
I do not really see how writing performance can be relevant in practical applications, but given that this is a pretty simple fix, there is no harm merging it. thanks for the contribution! (thanks for the review, @kayoub5 !)

seif-galal:
@andlaus youre welcome. its just to speed things up a bit...

PR #490, opened 2026-07-29
eliminate bitstruct in snoop/isotp_state in favor of native bit operations

seif-galal:
perf(isotp): eliminate bitstruct dependency from snoop/isotp state files

Replace bitstruct.unpack() and bitstruct.pack() with native Python
bit-shifting in isotp_state_machine.py. Remove bitstruct from
pyproject.toml dependencies — it is no longer used in this file.

Benchmark: 10,000 CAN frames
  bitstruct (pure Python): 0.671s
  bitstruct.c (C extension): 0.592s
  Native bit-shifting: 0.076s

Note: bitstruct remains in encodestate.py and decodestate.py for
ODX value encoding/decoding. Those use dynamic format strings
(runtime-determined bit_length and format_char) and are significantly
more complex to replace. This PR handles the snoop/ISO-TP path only.

### DIFF
--- a/odxtools/isotp_state_machine.py
+++ b/odxtools/isotp_state_machine.py
@@ -9,7 +9,6 @@ from enum import IntEnum
 from io import TextIOBase
 from typing import TextIO
 
-import bitstruct
 import can
 
 
@@ -56,12 +55,13 @@ class IsoTpStateMachine:
             return  # unknown CAN ID
 
         # decode the isotp segment
-        frame_type, _ = bitstruct.unpack("u4u4", data)
+        frame_type = data[0] >> 4
         assert isinstance(frame_type, int)
 
         telegram_len = None
         if frame_type == IsoTp.FRAME_TYPE_SINGLE:
-            frame_type, telegram_len = bitstruct.unpack("u4u4", data)
+            frame_type = data[0] >> 4
+            telegram_len = data[0] & 0x0F
             assert isinstance(telegram_len, int)
 
             self.on_single_frame(telegram_idx, data[1:1 + telegram_len])
@@ -70,7 +70,8 @@ class IsoTpStateMachine:
             yield (rx_id, bytes(data[1:1 + telegram_len]))
 
         elif frame_type == IsoTp.FRAME_TYPE_FIRST:
-            frame_type, telegram_len = bitstruct.unpack("u4u12", data)
+            frame_type = data[0] >> 4
+            telegram_len = ((data[0] & 0x0F) << 8) | data[1]
             assert isinstance(telegram_len, int)
 
             self._telegram_specified_len[telegram_idx] = telegram_len
@@ -80,7 +81,8 @@ class IsoTpStateMachine:
             self.on_first_frame(telegram_idx, data)
 
         elif frame_type == IsoTp.FRAME_TYPE_CONSECUTIVE:
-            frame_type, rx_segment_idx = bitstruct.unpack("u4u4", data)
+            frame_type = data[0] >> 4
+            rx_segment_idx = data[0] & 0x0F
             assert isinstance(rx_segment_idx, int)
 
             expected_segment_idx = (self._telegram_last_rx_fragment_idx[telegram_idx] + 1) % 16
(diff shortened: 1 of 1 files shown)


odxtools/isotp_state_machine.py
```diff
@@ -9,7 +9,6 @@
 from io import TextIOBase
 from typing import TextIO
 
-import bitstruct
```
andlaus, line 12: why not simply use `bitstruct.c`, i.e.:
```suggestion
try:
    import bitstruct.c as bitstruct
except ModuleNotFoundError:
    import bitstruct
```

I haven't benchmarked it, but I would not be surprised if it was faster than this proposal because it requires fewer interpreter statements.

andlaus (commented):
color me skeptical on this one: the `IsotpStateMachine` is not very time critical (it is only used by the `snoop` tool), and the PR does not address the other uses of `bitstruct` (i.e., encoding and decoding) which are much more important.

I'd also like to see some benchmarks comparing the "bare metal" approach with `bitstruct` and `bitstruct.c`...

odxtools/isotp_state_machine.py
seif-galal: @andlaus Fair points. 

I just tried benchmarking bitstruct.c to get a rough idea, i didnt add how much faster is native python actually is in the PR description but here are the results (per 10000 can frames):

bitstruct (pure python): 0.671s
bitstruct.c: 0.592s
Native bit-shifting (this PR): 0.076s

bitstruct.c: 0.592s vs native 0.076s for 10k frames. Native is 7.8x faster — even the C extension pays boundary call overhead, data[0] >> 4 doesn't.

if you want me to use bitstruct.c let me know.

seif-galal: @andlaus Also you're right that snoop isn't critical path. bitstruct only exists in this file, no encoding/decoding uses to expand to. this is a gain even if snoop isn't the hot path you still get less dependency surface, much simpler code and runs faster.

seif-galal: @andlaus hope all is well just reminding you of this PR to keep the wheels rolling:)

andlaus: sorry, I missed your reply: I think that getting rid of bitstruct might be worthwhile, but it should be done completely, i.e., also for the `encode()` and `decode()` paths; Only doing this for `snoop` is IMO not worth the squeeze...

andlaus: > bitstruct.c: 0.592s vs native 0.076s for 10k frames. Native is 7.8x faster — even the C extension pays boundary call overhead, data[0] >> 4 doesn't.

strange: when I asked claude code this question, it came back with 5-10%...

seif-galal: > sorry, I missed your reply: I think that getting rid of bitstruct might be worthwhile, but it should be done completely, i.e., also for the `encode()` and `decode()` paths; Only doing this for `snoop` is IMO not worth the squeeze...

at least we have seen  some gains by removing it for native.. ill test it out removing all bitstruct entirely from the encode/decode files and see what happens, but this will likely be too big of a PR so its your call whether we need another seperate PR just for that.

So we can either update it to bitstruct.c or remove it (again way too big).


but for now benchmarks look like this at the moment with native Python:

Batch   10:    0.582s    |  58.21ms/write |  17.2/s
Batch   50:    2.871s    |  57.42ms/write |  17.4/s
Batch  100:    5.429s   |  54.29ms/write |  18.4/s
Batch  200:   10.567s  |  52.83ms/write |  18.9/s
Batch  500:   24.387s  |  48.77ms/write |  20.5/s
Batch 1000:   48.547s |  48.55ms/write |  20.6/s

andlaus: My position is that I'd prefer to get rid of bitstruct entirely, but if you don't feel like adapting the encode/decode functionality, `iso_tp_state_machine.py` should be switched to `bitstruct.c`...

seif-galal:
closed. check #492

Issue #497, opened 2026-08-20
Multiplexer


cmydd:
physicalconstantparameter.py   
 def _resolve_snrefs(self, context: SnRefContext) -> None:
        super()._resolve_snrefs(context)

        dop = odxrequire(self.dop)
        if not isinstance(dop, DataObjectProperty):
            odxraise("The type of PHYS-CONST parameters must be a simple DOP")
            return
        base_data_type = dop.physical_type.base_data_type
        self._physical_constant_value = base_data_type.from_string(self.physical_constant_value_raw)

 <STRUCTURE ID="STRUC_DataRecorMeasuLedHeadlCurre">
              <SHORT-NAME>STRUC_DataRecorMeasuLedHeadlCurre</SHORT-NAME>
              <LONG-NAME>Data Record Measurment Led_headlamp_currents</LONG-NAME>
              <DESC>
                <p>Data Record Measurment Led_headlamp_currents</p>
              </DESC>
              <PARAMS>
                <PARAM xsi:type="PHYS-CONST">
                  <SHORT-NAME>Param_HeadlLeftChann1Maste</SHORT-NAME>
                  <LONG-NAME TI="MAS07091">Headlamp_left_channel_1_master</LONG-NAME>
                  <BYTE-POSITION>0</BYTE-POSITION>
                  <BIT-POSITION>0</BIT-POSITION>
                  <PHYS-CONSTANT-VALUE>0</PHYS-CONSTANT-VALUE>
                  <DOP-SNREF SHORT-NAME="MUX_LINEA2ByteCurreMUX" />
                </PARAM>
                <PARAM xsi:type="PHYS-CONST">
                  <SHORT-NAME>Param_HeadlRightChann1Maste</SHORT-NAME>
                  <LONG-NAME TI="MAS07092">Headlamp_right_channel_1_master</LONG-NAME>
                  <BYTE-POSITION>2</BYTE-POSITION>
                  <BIT-POSITION>0</BIT-POSITION>
                  <PHYS-CONSTANT-VALUE>0</PHYS-CONSTANT-VALUE>
                  <DOP-SNREF SHORT-NAME="MUX_LINEA2ByteCurreMUX2" />
                </PARAM>
                <PARAM xsi:type="PHYS-CONST">
                  <SHORT-NAME>Param_HeadlLeftChann1Maste2</SHORT-NAME>
                  <LONG-NAME TI="MAS07093">Headlamp_left_channel_1_master_2</LONG-NAME>
[...]

PR #498, opened 2026-08-20
improve robustness of the non-strict mode

andlaus:
with this, the file featured by https://github.com/mercedes-benz/odxtools/issues/497 can be loaded in non-strict mode. (although the file is very incomplete and the resulting database is thus probably of limited usefulness.)

### DIFF
--- a/odxtools/diaglayers/diaglayer.py
+++ b/odxtools/diaglayers/diaglayer.py
@@ -100,6 +100,10 @@ class DiagLayer:
             for import_ref in self.import_refs:
                 imported_dl = odxlinks.resolve(import_ref)
 
+                if imported_dl is None:
+                    # in non-strict mode, odxlinks.resolve() can fail...
+                    continue
+
                 # TODO: ensure that the imported diagnostic layer has
                 # not been referenced in any PARENT-REF of the current
                 # layer or any of its parents.
--- a/odxtools/diaglayers/diaglayerraw.py
+++ b/odxtools/diaglayers/diaglayerraw.py
@@ -234,6 +234,10 @@ class DiagLayerRaw(IdentifiableElement):
         for dc_proxy in self.diag_comms_raw:
             if isinstance(dc_proxy, OdxLinkRef):
                 dc = odxlinks.resolve(dc_proxy, DiagComm)
+
+                if dc is None:
+                    # in non-strict mode, odxlinks.resolve() can return None
+                    continue
             else:
                 dc = dc_proxy
                 dc._resolve_odxlinks(odxlinks)
--- a/odxtools/diaglayers/hierarchyelement.py
+++ b/odxtools/diaglayers/hierarchyelement.py
@@ -243,9 +243,17 @@ class HierarchyElement(DiagLayer):
         self._state_charts = NamedItemList(state_charts)
 
     def _get_parent_refs_sorted_by_priority(self, reverse: bool = False) -> Iterable[ParentRef]:
+
+        def get_inheritance_priority(layer: DiagLayer) -> int:
+            if layer is None:
+                # in non-strict mode, a parent layer ref sometimes
+                # cannot be resolved
+                return 0
+            return layer.variant_type.inheritance_priority
+
         return sorted(
             getattr(self.diag_layer_raw, "parent_refs", []),
-            key=lambda pr: pr.layer.variant_type.inheritance_priority,
+            key=lambda parent_ref: get_inheritance_priority(parent_ref.layer),
             reverse=reverse)
 
     def _compute_available_objects(
@@ -278,6 +286,11 @@ class HierarchyElement(DiagLayer):
         # populate the result dictionary with the inherited objects
         for parent_ref in self._get_parent_refs_sorted_by_priority(reverse=True):
             parent_dl = parent_ref.layer
+
+            if parent_dl is None:
+                # in non-strict mode, parent references sometimes
+                # cannot be resolved
+                continue
 
             # retrieve the set of short names of the objects which we
             # are not supposed to inherit
--- a/odxtools/multiplexer.py
+++ b/odxtools/multiplexer.py
@@ -85,8 +85,12 @@ class Multiplexer(ComplexDop):
             self.default_case._resolve_odxlinks(odxlinks)
 
         for mux_case in self.cases:
+            dop = self.switch_key.dop
+            if dop is None:
+                # the DOP can be None in non-strict mode
+                continue
             mux_case._mux_case_resolve_odxlinks(
-                odxlinks, key_physical_type=self.switch_key.dop.physical_type.base_data_type)
+                odxlinks, key_physical_type=dop.physical_type.base_data_type)
 
     @override
     def _resolve_snrefs(self, context: SnRefContext) -> None:
--- a/odxtools/nameditemlist.py
+++ b/odxtools/nameditemlist.py
@@ -52,6 +52,12 @@ class ItemAttributeList(list[T]):
 
         \return The name under which item is accessible
         """
+
+        if not hasattr(item, "short_name"):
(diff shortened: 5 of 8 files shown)


odxtools/parameters/physicalconstantparameter.py
```diff
@@ -63,12 +63,16 @@ def _resolve_odxlinks(self, odxlinks: OdxLinkDatabase) -> None:
 
     @override
     def _resolve_snrefs(self, context: SnRefContext) -> None:
+        self._physical_constant_value: ParameterValue
```
kayoub5, line 66: what's this for ?

odxtools/nameditemlist.py
```diff
@@ -52,6 +52,12 @@ def append(self, item: T) -> None:
 
         \return The name under which item is accessible
         """
+
+        if not hasattr(item, "short_name"):
```
kayoub5: failing short name is a fundamental structure issue of the document (not an unfinished document / edge case kind of issue)

odxtools/diaglayers/diaglayer.py
```diff
@@ -100,6 +100,10 @@ def _resolve_odxlinks(self, odxlinks: OdxLinkDatabase) -> None:
             for import_ref in self.import_refs:
                 imported_dl = odxlinks.resolve(import_ref)
 
+                if imported_dl is None:
+                    # in non-strict mode, odxlinks.resolve() can fail...
```
kayoub5, line 104: no logged warning ?

odxtools/diaglayers/diaglayerraw.py
```diff
@@ -234,6 +234,10 @@ def _resolve_odxlinks(self, odxlinks: OdxLinkDatabase) -> None:
         for dc_proxy in self.diag_comms_raw:
             if isinstance(dc_proxy, OdxLinkRef):
                 dc = odxlinks.resolve(dc_proxy, DiagComm)
+
+                if dc is None:
+                    # in non-strict mode, odxlinks.resolve() can return None
```
kayoub5, line 239: no logged warning ?

odxtools/diaglayers/hierarchyelement.py
```diff
@@ -243,9 +243,17 @@ def _compute_value_inheritance(self, odxlinks: OdxLinkDatabase) -> None:
         self._state_charts = NamedItemList(state_charts)
 
     def _get_parent_refs_sorted_by_priority(self, reverse: bool = False) -> Iterable[ParentRef]:
+
+        def get_inheritance_priority(layer: DiagLayer) -> int:
+            if layer is None:
+                # in non-strict mode, a parent layer ref sometimes
+                # cannot be resolved
+                return 0
```
kayoub5, line 251: no logged warning ?

odxtools/diaglayers/hierarchyelement.py
```diff
@@ -279,6 +287,11 @@ def _compute_available_objects(
         for parent_ref in self._get_parent_refs_sorted_by_priority(reverse=True):
             parent_dl = parent_ref.layer
 
+            if parent_dl is None:
+                # in non-strict mode, parent references sometimes
+                # cannot be resolved
+                continue
```
kayoub5, line 293: no logged warning ?

odxtools/parameters/physicalconstantparameter.py
```diff
@@ -81,6 +85,10 @@ def _encode_positioned_into_pdu(self, physical_value: ParameterValue | None,
                 f"only be specified as {self.physical_constant_value!r} (is: {physical_value!r})",
                 EncodeError)
 
+        if self.dop is None:
+            # in non-strict mode, the DOP-REF might not be resolvable
```
kayoub5, line 89: no logged warning ?

odxtools/multiplexer.py
```diff
@@ -85,8 +85,12 @@ def _resolve_odxlinks(self, odxlinks: OdxLinkDatabase) -> None:
             self.default_case._resolve_odxlinks(odxlinks)
 
         for mux_case in self.cases:
+            dop = self.switch_key.dop
+            if dop is None:
+                # the DOP can be None in non-strict mode
+                continue
```
kayoub5, line 91: instead of silently ignore the issue, covert the code comments to a logged warning

odxtools/parameters/physicalconstantparameter.py
andlaus: without this `mypy` complaines that `self._physical_constant_value` is `Any` even with the cast-assignment on the next hunk...

odxtools/nameditemlist.py
andlaus: the problem is that if a reference resolution fails in non-strict mode `None` is returned and thus the lists created using list-comprehensions of resolved references may contain `None` objects...

odxtools/diaglayers/diaglayer.py
andlaus: `odxlinks.resolve()` already does this...

odxtools/diaglayers/diaglayerraw.py
andlaus: same as above: `odxlinks.resolve()` will print a warning instead of raising an exception on resolution failure in non-strict mode.

odxtools/diaglayers/hierarchyelement.py
andlaus: ditto

odxtools/diaglayers/hierarchyelement.py
andlaus: ditto

odxtools/parameters/physicalconstantparameter.py
andlaus: ditto

odxtools/multiplexer.py
andlaus: the switch key should already have produced a warning when it tried to resolve the DOP-REF...

odxtools/nameditemlist.py
andlaus: (for strict mode, I agree with your point 100%. On the other hand, the point of non-strict mode is to be able to load and handle non-compliant datasets, even if they are *severely* non-compliant.)

kayoub5: then filter on None, not on short_name

andlaus: I don't mind going that route (I originally even did it like this), but I'm not sure if this is such a good idea: what we actually need for `.append()` not to raise an exception is a `.short_name` attribute. (performance-wise, `is None` is a bit faster, but the difference is far from relevant IMO...)

kayoub5: I would prefer to avoid hasattr, since this function is a hot path called for almost every object in the database

andlaus: in odxtools, adding items to `NamedItemList`s is not a hot path: it only needs  to be done once per item and after these lists have been created, they are basically immutable. (*accessing* the list items *is* a hot path though...)

kayoub5: we have a different definition of hot path, I work with pdx files that are few hundred megabytes in size (contains ECUs definition for entire lineup (multiple models, same year/oem)

including all internal/development operations

so for me loading one pdx can take multiple seconds, and I would like to avoid making that minutes.

andlaus: okay, I relent: https://github.com/mercedes-benz/odxtools/pull/498/changes/77b0d070363c666f16b864accc7a02176c42892c

for funsies, it would be cool if you measured the performance of the current version of this PR vs the previous version:
```
time odxtools list MY_FILE.pdx > /dev/null
```

in one of the biggest files I need to handle, this was 13.15 (`is None` verstion) vs 13.25 seconds with a variance of about 0.5 seconds over 10 runs...

kayoub5 (approved):

PR #499, opened 2026-08-22
Name the offending element when a reference points to the wrong kind of DOP

Poseidonas:
While looking at #497 I noticed that the message which reports the problem does not say which parameter caused it. On that user's file it is printed 86 times, and only two of those messages differ from each other, so finding the 71 offending parameters in an 877 KB file means searching by hand. That is roughly what happened in the issue thread.

This PR leaves the behaviour alone and only changes the text of the diagnostics.

**Before**

```
The type of PHYS-CONST parameters must be a simple DOP
The type of PHYS-CONST parameters must be a simple DOP
... 86 lines, 2 distinct
```

**After**

```
PHYS-CONST parameter 'Param_HeadlLeftChann1Maste' references DOP
'MUX_LINEA2ByteCurreMUX' which is a Multiplexer, but a simple DOP is required

PHYS-CONST parameter 'Param_DiagnSessiType' references DOP
'DOP_TEXTTABLEDiagnSessiTypes' which could not be resolved, but a simple
DOP is required
```

86 lines, 67 distinct. The second example is worth pointing out: an unresolved reference and a reference of the wrong type used to produce the same sentence, and they are different problems. In non-strict mode `odxrequire()` returns `None`, so the two cases reach the same branch.

The same treatment is applied to the three other places with this constraint, since they share the shortcoming:

- `PhysicalConstantParameter._resolve_snrefs`
- `ValueParameter._resolve_snrefs`
- `TableRow._resolve_odxlinks` (DOP-REF)
- `TableRow._resolve_snrefs` (DOP-SNREF)

The shared part sits in `ParameterWithDOP._dop_description`, which names the reference as written in the file rather than the resolved object, so it also works when resolution failed.

**Verification**

Numbers above are from loading the file attached to #497 in non-strict mode before and after the change. `yapf`, `ruff` and `mypy` are clean, and the test suite goes from 155 to 159 passing — four new cases in `tests/test_dop_type_diagnostics.py` covering both parameter types, the unresolved case and the reference naming.

I did not touch what is accepted or rejected, so no existing test changed.

Happy to adjust the wording if you would phrase these differently, or to drop the `TableRow` part if you would rather keep the change to the parameters only.

### DIFF
--- a/odxtools/parameters/parameterwithdop.py
+++ b/odxtools/parameters/parameterwithdop.py
@@ -32,7 +32,7 @@ class ParameterWithDOP(Parameter):
         return self._dop
 
     @property
-    def _dop_description(self) -> str:
+    def dop_info(self) -> str:
         """Describe the referenced DOP for diagnostic messages.
 
         Names the reference as it appears in the ODX file and, if it could be
@@ -40,10 +40,9 @@ class ParameterWithDOP(Parameter):
         """
         name = self.dop_snref if self.dop_snref is not None else (
             self.dop_ref.ref_id if self.dop_ref is not None else "<unspecified>")
-        dop = getattr(self, "_dop", None)
-        if dop is None:
-            return f"DOP '{name}' which could not be resolved"
-        return f"DOP '{name}' which is a {type(dop).__name__}"
+        if self.dop is None:
+            return f"unresolvable DOP '{name}'"
+        return f"{type(self.dop).__name__} DOP '{name}'"
 
     @staticmethod
     @override
--- a/odxtools/parameters/parameterwithdop.py
+++ b/odxtools/parameters/parameterwithdop.py
@@ -31,6 +31,20 @@ class ParameterWithDOP(Parameter):
 
         return self._dop
 
+    @property
+    def _dop_description(self) -> str:
+        """Describe the referenced DOP for diagnostic messages.
+
+        Names the reference as it appears in the ODX file and, if it could be
+        resolved, the type of the object which it points to.
+        """
+        name = self.dop_snref if self.dop_snref is not None else (
+            self.dop_ref.ref_id if self.dop_ref is not None else "<unspecified>")
+        dop = getattr(self, "_dop", None)
+        if dop is None:
+            return f"DOP '{name}' which could not be resolved"
+        return f"DOP '{name}' which is a {type(dop).__name__}"
+
     @staticmethod
     @override
     def from_et(et_element: ElementTree.Element, context: OdxDocContext) -> "ParameterWithDOP":
--- a/tests/test_dop_type_diagnostics.py
+++ b/tests/test_dop_type_diagnostics.py
@@ -1,0 +1,94 @@
+# SPDX-License-Identifier: MIT
+"""Diagnostics emitted when a reference points to the wrong kind of DOP.
+
+The ODX standard only allows simple DOPs in a few places, e.g. for the value
+of a PHYS-CONST parameter. When a file violates this, the message needs to
+name the offending element so that it can be found in a large database.
+"""
+import unittest
+
+from odxtools.exceptions import OdxError
+from odxtools.odxlink import (DocType, OdxDocFragment, OdxLinkDatabase, OdxLinkId, OdxLinkRef)
+from odxtools.parameters.physicalconstantparameter import PhysicalConstantParameter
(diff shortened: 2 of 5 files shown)


tests/test_dop_type_diagnostics.py
```diff
@@ -0,0 +1,94 @@
...
+from odxtools.structure import Structure
+
+doc_frags = (OdxDocFragment("test", DocType.CONTAINER),)
+
+
+def _structure(short_name: str) -> Structure:
+    return Structure(
+        odx_id=OdxLinkId(short_name, doc_frags),
+        short_name=short_name,
+        parameters=[],  # type: ignore[arg-type]
```
andlaus:
```suggestion
        parameters=NamedItemList(),
```

should do the trick

odxtools/parameters/parameterwithdop.py
```diff
@@ -31,6 +31,20 @@ def dop(self) -> DopBase:
 
         return self._dop
 
+    @property
+    def _dop_description(self) -> str:
```
andlaus: no need to make this a non-public method. Also this should probably better be named differently because I would associate the term "description" with an explanation what this DOP does, e.g.
```suggestion
    def dop_info(self) -> str:
```

odxtools/parameters/parameterwithdop.py
```diff
@@ -31,6 +31,20 @@ def dop(self) -> DopBase:
...
+        """Describe the referenced DOP for diagnostic messages.
+
+        Names the reference as it appears in the ODX file and, if it could be
+        resolved, the type of the object which it points to.
+        """
+        name = self.dop_snref if self.dop_snref is not None else (
+            self.dop_ref.ref_id if self.dop_ref is not None else "<unspecified>")
+        dop = getattr(self, "_dop", None)
+        if dop is None:
+            return f"DOP '{name}' which could not be resolved"
```
andlaus: this is slightly too verbose for my taste
```suggestion
            return f"unresolvable DOP '{name}'"
```

odxtools/parameters/parameterwithdop.py
```diff
@@ -31,6 +31,20 @@ def dop(self) -> DopBase:
...
+        return f"DOP '{name}' which is a {type(dop).__name__}"
```
andlaus:
```suggestion
        return f"{type(dop).__name__} DOP '{name}'"
```

odxtools/parameters/parameterwithdop.py
```diff
@@ -31,6 +31,20 @@ def dop(self) -> DopBase:
...
+    def _dop_description(self) -> str:
+        """Describe the referenced DOP for diagnostic messages.
+
+        Names the reference as it appears in the ODX file and, if it could be
+        resolved, the type of the object which it points to.
+        """
+        name = self.dop_snref if self.dop_snref is not None else (
+            self.dop_ref.ref_id if self.dop_ref is not None else "<unspecified>")
+        dop = getattr(self, "_dop", None)
+        if dop is None:
```
andlaus: one can safely assume that `self._dop` exists here:
```suggestion
        if self.dop is None:
```

andlaus (commented):
looks pretty good. I found the usual number of nitpicks...

Poseidonas:
Thanks for the quick look. All five applied in 4398026.

On the `getattr` — you were right, and I checked what it was actually guarding against before removing it. `_dop` is only annotated in `__post_init__`, so it is genuinely absent before resolution, but after a *failed* resolution in non-strict mode it exists and holds `None`. Since `dop_info` is only reached after resolution has run, `self.dop` covers it. My test was the only caller that hit the missing-attribute case, and it was reaching for the property without resolving first, which was not representative; it now resolves against an empty database instead.

The wording is better short. On the file from #497 the messages now read:

```
PHYS-CONST parameter 'Param_HeadlLeftChann1Maste' references Multiplexer DOP
'MUX_LINEA2ByteCurreMUX', but a simple DOP is required

PHYS-CONST parameter 'Param_DiagnSessiType' references unresolvable DOP
'DOP_TEXTTABLEDiagnSessiTypes', but a simple DOP is required
```

`yapf`, `ruff` and `mypy` are clean and the suite is at 159 passing.

andlaus:
thanks for the contribution and the quick fixes; it looks good now. merging!

PR #501, opened 2026-08-22
Name the element in the remaining load-time diagnostics

Poseidonas:
Follow-up to #499. That one covered the four places which require a simple DOP; while looking at them I noted five more messages in the load path that name neither the element they are about nor what was found instead.

| file | message |
| --- | --- |
| `ecumemconnector.py` | `Invalid type of referenced object` |
| `commrelation.py` | `No DIAG-COMM referenced by COMM-RELATION` |
| `matchingparameter.py` | `Output parameter must not be left unspecified` |
| `subcomponentparamconnector.py` (×2) | `Currently, only SNREFS are supported for …-PARAM-IF-REFS` |

The first one is the reason I bothered: on a database with several ECU-MEM-CONNECTORs it says nothing at all. And in `commrelation.py` the message two lines below already reads `DIAG-VARIABLE references non-service {type} diagnostic communication`, so two adjacent checks on the same object report very differently.

Each message now names the element, and where the file gives something to compare against, what was found:

```
LAYER-REF 'layer.42' of ECU-MEM-CONNECTOR 'MyConnector' references a
DiagLayerContainer, but a base or ECU variant is required

SUB-COMPONENT-PARAM-CONNECTOR 'MyConnector' uses OUT-PARAM-IF-REF in
OUT-PARAM-IF-REFS; only OUT-PARAM-IF-SNREF is currently supported

MATCHING-PARAMETER for DIAG-COMM 'ReadIdent' specifies neither
OUT-PARAM-IF-SNREF nor OUT-PARAM-IF-SNPATHREF
```

**One site deliberately left alone**

`dynenddopref.py:32` reports a missing mandatory `DYN-END-DOP-REF` tag. The element does not exist at that point, so there is nothing to name — adding the parent would mean threading it through `from_et`, which seemed disproportionate for one message.

**Verification**

`yapf`, `ruff` and `mypy` clean; suite 159 → 161. The two new tests in `tests/test_load_error_context.py` both fail on `main` and pass with the change — the `SUB-COMPONENT-PARAM-CONNECTOR` one because the message carried no name at all, so there was nothing to assert on.

Behaviour is unchanged; only the text of the diagnostics. Same caveat as last time on wording — happy to adjust if you would phrase any of these differently.


odxtools/ecumemconnector.py
```diff
@@ -104,7 +104,9 @@ def _resolve_odxlinks(self, odxlinks: OdxLinkDatabase) -> None:
         for ref in self.layer_refs:
             x = odxlinks.resolve(ref)
             if not isinstance(x, (BaseVariant, EcuVariant)):
-                odxraise("Invalid type of referenced object")
+                odxraise(f"LAYER-REF '{ref.ref_id}' of ECU-MEM-CONNECTOR "
+                         f"'{self.short_name}' references a {type(x).__name__}, but a "
```
andlaus: in this context using `x.__class__.__name__` works better than `type(x).__name__` because it works fine with `weakref.proxy` objects...

tests/test_load_error_context.py
```diff
@@ -0,0 +1,54 @@
+# SPDX-License-Identifier: MIT
```
andlaus: I think this it is better not to test the specifics of the exception messages (i.e., remove the unit tests file)

odxtools/ecumemconnector.py
Poseidonas: Done, and thank you for pointing at it twice over — this is the second time the same thing has caught me, so it has landed properly now.

Confirmed why it matters here rather than taking it on trust: `layer_refs` are resolved through `odxlinks.resolve()`, and `OdxLinkDatabase` carries `use_weakrefs`, so `x` really can arrive as a proxy.

```
type(proxy).__name__      -> weakproxy
proxy.__class__.__name__  -> BaseVariant
```

Which is precisely the case the message exists to describe, so it would have gone blank exactly when it was needed.

One thing I have deliberately left alone: the same pattern is in about thirty other files, including two lines in `commrelation.py` and `subcomponentparamconnector.py` that this PR touches but did not introduce. That felt like a separate change rather than something to fold in here without asking. Happy to open it as its own PR if it is worth having.

tests/test_load_error_context.py
Poseidonas: Removed. You are right, and the reasoning applies more widely than I had it: asserting on the wording pins the exact thing the PR is trying to leave room to improve. The next person to reword a message would have had to edit a test to say the same thing twice.

`mypy .`, `ruff check .` and the suite are clean after both changes: 325 files, 159 passed, 1 skipped.

Poseidonas:
Both points are addressed and the sign-off I had dropped is back, so DCO is green again.

`linting (3.10)` is still red, and it is not from this branch — it comes from `main`. The revert in e636302 took the Jinja2 environment caching back out but left its import behind:

```
odxtools/writepdxfile.py:9:  from functools import cache
```

Nothing in the file uses it any more, so ruff 0.16.4 reports F401. Checking both trees with the same ruff the CI installs:

```
my branch, on its own   -> All checks passed!
upstream/main           -> Found 1 error
```

The pull request build merges into `main`, which is where it picks the failure up. #502 is green only because its run predates the revert; anything re-run now will hit it.

I have deliberately not touched it here — it is unrelated to this change and folding it in would make the diff say two things at once. Happy to send it as a one-line PR of its own if that is useful, or leave it with you.

andlaus:
thanks for the patch, merging.

> I have deliberately not touched it here — it is unrelated to this change and folding it in would make the diff say two things at once. Happy to send it as a one-line PR of its own if that is useful, or leave it with you.

go wild :)

PR #502, opened 2026-08-22
Add a consistency checker with a first rule

Poseidonas:
Towards #3. I would like to be clear about what this is: a starting point, not an answer to the whole issue. One rule, the structure to hang further ones on, and a CLI entry point — small enough to discuss the shape before more rules are written on top of it.

If the direction is right, I am happy to keep adding rules over time, including working through annex B. If it is not, little has been invested and it is easy to discard.

**What it does today**

```console
$ odxtools check ./somersault.pdx
no findings

$ odxtools check ./some-real-database.pdx
error: MyEcu/DiagnServi_InputOutpuContrByIdent/Resp_InputOutpuContrByIdent: parameters
'Param_IOContrIdent' (bits 8..23) and 'Param_MatchIOContrIdent' (bits 8..23) overlap
[overlapping-parameters]

1 error(s), 0 warning(s)
```

Non-zero exit when errors are found, so it can sit in a pipeline; `--warnings-as-errors` for the stricter case.

**The one rule**

Parameters of the same request, response or structure which cover the same bits. It only considers parameters whose position *and* length are both known from the database — one which follows whatever precedes it, or whose length depends on the data it carries, is passed over rather than guessed at. That keeps it quiet about layouts it cannot reason about.

Requests, responses and structures all carry parameters without sharing a base class, so they are collected through the `CompositeCodec` protocol rather than by enumerating types.

**How another rule gets added**

A rule is an object with a `name` and a `check(database)` method yielding `Finding`s, registered in `checks.RULES`. A `Finding` carries a severity, the path of short names where it was found, and a sentence naming the objects involved. Adding one means a new module and one line in the registry — no changes to the CLI or the framework.

The severity split is `ERROR` for what a conforming database cannot contain, `WARNING` for what the specification permits but which is worth a second look. Only `ERROR` is used so far.

**Verification**

- `somersault.pdx` and `somersault_modified.pdx`: no findings, so it is not noisy on conforming input.
- The file attached to #497: four findings, each a response placing a parameter and its `MatchingRequestParameter` on the same bits. That file is invalid for other reasons too, but this is not something a conforming database should contain and nothing in the loader mentions it today.
- 11 tests, covering bit-span arithmetic, adjacency, partial and total overlap, parameters that cannot be placed, and that the reference database stays clean.
- `yapf`, `ruff` and `mypy .` clean; suite 159 → 170.

**One thing I deliberately left out**

The description also names ambiguous request/response IDs. I prototyped it and it did not hold up: on a real database it flagged eleven pairs, all `ReadDataByIdentifier`-style services whose identifier is a `ValueParameter` rather than a coded constant, so `coded_const_prefix()` stops after the SID for both. Strictly the peer cannot tell them apart from the prefix alone, but a database modelling a generic service that way is making a choice, not a mistake, and reporting all of them would bury the real problems.

So that rule needs a definition first — probably narrower than "identical prefixes" — and that seemed like your call rather than something to decide inside a PR.

**Two questions**

1. Does `check` as a subcommand fit how you see the tools, or would you rather the rules were only a library function that other tools call?
2. Is `ERROR`/`WARNING` the right split, or would you prefer findings to carry the annex B rule identifier once those are implemented?

Happy to rework any of this, and equally happy to step back if you would rather own the checker yourself.


odxtools/checks/__init__.py
```diff
@@ -0,0 +1,82 @@
...
+    #: e.g. ``("MyEcu", "MyService", "MyResponse")``
+    location: tuple[str, ...] = field(default_factory=tuple)
+
+    #: what is wrong, in one sentence, naming the objects involved
+    message: str
+
+    def __str__(self) -> str:
+        where = "/".join(self.location)
+        prefix = f"{where}: " if where else ""
+        return f"{self.severity.value}: {prefix}{self.message} [{self.rule}]"
```
andlaus: for easier integration with editors it is probably better to print the location first (compilers and linters usually do it like this and e.g. emacs parses compilation output using this format to make the location clickable):
```suggestion
        return f"{prefix}{self.severity.value}: {self.message} [{self.rule}]"
```

this becomes more useful once a file name and a line number is added. Unfortunately, that requires to use an XML parser other than `ElementTree`.  (e.g., `lxml` works, but it comes with subtle performance and behavioral differences opening a can of worms which I would prefer to leave closed.)

odxtools/checks/overlapping_parameters.py
```diff
@@ -0,0 +1,102 @@
+# SPDX-License-Identifier: MIT
+"""Parameters of the same object which occupy the same bits."""
```
andlaus:
```suggestion
"""Check that Parameters of the same codec object do not occupy the same bits in a PDU."""
```

odxtools/checks/overlapping_parameters.py
```diff
@@ -0,0 +1,102 @@
...
+    A parameter without a byte position follows whatever precedes it, and one
+    without a static length is sized by its value, so neither can be placed
+    without decoding an actual message.
+    """
+    if param.byte_position is None:
+        return None
+
+    try:
+        bit_length = param.get_static_bit_length()
+    except AttributeError:
```
andlaus: ` param.get_static_bit_length()` is not supposed to raise an exception (please fix it the non-strict fallback code of that method instead.)

odxtools/checks/overlapping_parameters.py
```diff
@@ -0,0 +1,102 @@
...
+    if bit_length is None:
+        return None
+
+    start = param.byte_position * 8 + (param.bit_position or 0)
+    return start, start + bit_length
+
+
+def _overlaps_in(codec: CompositeCodec) -> Iterator[tuple[Parameter, Parameter]]:
+    placed: list[tuple[tuple[int, int], Parameter]] = []
+    for param in codec.parameters:
```
andlaus, line 35: can't `param` be `None` for non-conforming databases?
```suggestion
    for param in codec.parameters:
        if param is None: 
            # this can happen in non-strict mode
            return
```

odxtools/checks/__init__.py
```diff
@@ -0,0 +1,82 @@
...
+
+class Severity(Enum):
+    """How much a finding matters.
+
+    ``ERROR`` is for something a conforming database cannot contain and which
+    will make some operation on it fail or produce a wrong result. ``WARNING``
+    is for something which is suspect but which the specification permits.
+    """
+
+    ERROR = "error"
```
andlaus: INFO and DEBUG levels would probably useful as well.

odxtools/checks/overlapping_parameters.py
```diff
@@ -0,0 +1,102 @@
...
+            if a_start < b_end and b_start < a_end:
+                yield a, b
+
+
+class OverlappingParameters:
+    """Two parameters of one structure which cover the same bits.
+
+    Only parameters whose position and length are both known from the database
+    are considered, so a structure whose layout depends on the data it carries
+    is passed over rather than guessed at.
```
andlaus: instead of silently ignoring such codecs, a low-priority log message could be produced. (maybe DEBUG?)

odxtools/cli/check.py
```diff
@@ -0,0 +1,47 @@
...
+    errors = warnings = 0
+    for finding in run_checks(odxdb):
+        print(finding)
+        if finding.severity is Severity.ERROR:
+            errors += 1
+        else:
+            warnings += 1
+
+    if errors or warnings:
+        print(f"\n{errors} error(s), {warnings} warning(s)")
```
andlaus:
```suggestion
        print()
        print(f"{errors} error(s), {warnings} warning(s)")
```

odxtools/checks/__init__.py
```diff
@@ -0,0 +1,82 @@
...
+
+
+def run_checks(database: Database, rules: Iterable[Rule] | None = None) -> Iterator[Finding]:
+    """Apply ``rules`` to ``database`` and yield what they object to.
+
+    Args:
+        database: the database to inspect
+        rules: the rules to apply, defaulting to :data:`RULES`
+    """
+    for rule in RULES if rules is None else rules:
```
andlaus: better do this in two lines:
```suggestion
    selected_rules = RULES if rules is None else rules
    for rule in selected_rules:
```

odxtools/checks/overlapping_parameters.py
```diff
@@ -0,0 +1,102 @@
...
+def _overlaps_in(codec: CompositeCodec) -> Iterator[tuple[Parameter, Parameter]]:
+    placed: list[tuple[tuple[int, int], Parameter]] = []
+    for param in codec.parameters:
+        span = _bit_span(param)
+        if span is not None:
+            placed.append((span, param))
+
+    for i, ((a_start, a_end), a) in enumerate(placed):
+        for (b_start, b_end), b in placed[i + 1:]:
+            if a_start < b_end and b_start < a_end:
```
andlaus: `b` can be physically be located before `a`:
```suggestion
            if (a_start < b_end and b_start < a_end) or (b_start < a_end and a_start < b_end):
```

odxtools/checks/overlapping_parameters.py
```diff
@@ -0,0 +1,102 @@
...
+                    severity=Severity.ERROR,
+                    location=location,
+                    message=(f"parameters '{a.short_name}' (bits {a_span[0]}..{a_span[1] - 1}) "
+                             f"and '{b.short_name}' (bits {b_span[0]}..{b_span[1] - 1}) "
+                             f"overlap"),
+                )
+
+
+def _codecs(database: Database) -> Iterator[tuple[CompositeCodec, tuple[str, ...]]]:
+    """Every parameter-carrying object reachable from ``database``.
```
andlaus:
```suggestion
    """Find every parameter-carrying object reachable from ``database``.
```

odxtools/checks/overlapping_parameters.py
```diff
@@ -0,0 +1,102 @@
...
+            for attribute in ("request", "positive_responses", "negative_responses"):
+                value = getattr(service, attribute, None)
+                candidates = value if isinstance(value, (list, tuple)) else [value]
+                for candidate in candidates:
+                    if not isinstance(candidate, CompositeCodec) or id(candidate) in seen:
+                        continue
+                    seen.add(id(candidate))
+                    yield candidate, (layer_name, service.short_name, candidate.short_name)
+
+        ddd_spec = getattr(diag_layer, "diag_data_dictionary_spec", None)
```
andlaus:
```suggestion
        ddd_spec = diag_layer.diag_data_dictionary_spec
```

odxtools/checks/overlapping_parameters.py
```diff
@@ -0,0 +1,102 @@
...
+        for structure in getattr(ddd_spec, "structures", []):
```
andlaus: explicitly checking for overlaps in structures is arguably only necessary for the those structures that have not been caught during inspection of the requests and responses. (i.e., if the structure is not part of a codec which has a static bit length) Regardless, do not use `getattr` here:
```suggestion
        for structure in ddd_spec.structures:
```

finally, if structures are checked for internal overlaps, so should the environment datas, i.e.,
```suggestion
        for structure in chain(ddd_spec.structures, ddd_spec.environment_datas):
```

odxtools/checks/__init__.py
```diff
@@ -0,0 +1,82 @@
...
+
+    #: identifier used to name the rule in output and to select it
+    name: str
+
+    def check(self, database: Database) -> Iterable[Finding]:
+        """Yield a finding for every object this rule objects to."""
+        ...
+
+
+from .overlapping_parameters import OverlappingParameters  # noqa: E402
```
andlaus: move this to the top of the file with the other imports. this should also get rid of the `noqa:`

odxtools/checks/__init__.py
```diff
@@ -0,0 +1,82 @@
...
+    will make some operation on it fail or produce a wrong result. ``WARNING``
+    is for something which is suspect but which the specification permits.
+    """
+
+    ERROR = "error"
+    WARNING = "warning"
+
+
+@dataclass(frozen=True, kw_only=True)
+class Finding:
```
andlaus: I'd prefer if these classes were moved to separate files...

odxtools/checks/__init__.py
```diff
@@ -0,0 +1,82 @@
...
+    def check(self, database: Database) -> Iterable[Finding]:
+        """Yield a finding for every object this rule objects to."""
+        ...
+
+
+from .overlapping_parameters import OverlappingParameters  # noqa: E402
+
+#: The rules which :func:`run_checks` applies by default. New rules are added
+#: here once they are known not to report conforming databases.
+RULES: tuple[Rule, ...] = (OverlappingParameters(),)
```
andlaus: better name appreciated. how about `DEFAULT_RULES`?

odxtools/checks/overlapping_parameters.py
```diff
@@ -0,0 +1,102 @@
...
+
+    Requests, responses and structures all carry parameters without sharing a
+    base class, so they are identified through the CompositeCodec protocol.
+    """
+    seen: set[int] = set()
+
+    for diag_layer in database.diag_layers:
+        layer_name = diag_layer.short_name
+
+        for service in getattr(diag_layer, "services", []):
```
andlaus: avoid `getattr` if possible:

```suggestion
        for diag_comm in diag_layer.diag_comms:
            if not isinstance(diag_comm, DiagService):
                continue
```

odxtools/checks/overlapping_parameters.py
```diff
@@ -0,0 +1,102 @@
...
+            for attribute in ("request", "positive_responses", "negative_responses"):
+                value = getattr(service, attribute, None)
+                candidates = value if isinstance(value, (list, tuple)) else [value]
+                for candidate in candidates:
```
andlaus: use `chain()` here and rename `candidate` to `codec_obj`:
```suggestion
            for attribute in chain([diag_comm.request], diag_comm.positive_responses, diag_comm.negative_responses):
```

odxtools/checks/overlapping_parameters.py
```diff
@@ -0,0 +1,102 @@
...
+                    rule=self.name,
+                    severity=Severity.ERROR,
+                    location=location,
+                    message=(f"parameters '{a.short_name}' (bits {a_span[0]}..{a_span[1] - 1}) "
+                             f"and '{b.short_name}' (bits {b_span[0]}..{b_span[1] - 1}) "
+                             f"overlap"),
+                )
+
+
+def _codecs(database: Database) -> Iterator[tuple[CompositeCodec, tuple[str, ...]]]:
```
andlaus: it might be worthwhile to consider to allow any object here and recursively go through its attributes. (possibly it is a bad idea because there might be issues due to reference resolution...)

andlaus (commented):
looks fine for a first shot. (it will be quite a bit of effort to add checks for the roughly 250 checker rules defined by the ODX standard in appendix B.2.)

Poseidonas:
Thank you for going through it that thoroughly — this was a lot of reading on your part. Everything is in except three points, which are below with what I measured rather than what I think.

**Applied**

Location first in the message, so it reads `layer/service/response: error: … [rule]`. `Finding`, `Severity` and `Rule` are in their own modules, which lets the rule import sit at the top with the others and takes the `noqa` with it. `RULES` is now `DEFAULT_RULES`. The CLI prints a blank line before the summary. `diag_comms` filtered by `isinstance(…, DiagService)` instead of `services`, `chain()` for the request and responses, `codec_obj` instead of `candidate`, `ddd_spec` read directly with no `getattr`, and the two-line form for the selected rules.

`INFO` and `DEBUG` are in. A codec whose parameters cannot all be placed now produces a `DEBUG` finding saying how many were left out, so a database with nothing wrong can be told apart from one that could not be inspected. The CLI takes `--severity` to decide what to print and counts only errors and warnings towards the exit status, so the new levels cannot change anyone's pipeline.

**`get_static_bit_length()` — you were right, and the guard is gone**

I checked before removing it rather than just deleting it, and could not make it raise:

```
somersault.pdx           164 parameters   0 raised
somersault_modified.pdx  212 parameters   0 raised
```

in strict and non-strict mode alike. So the `try/except AttributeError` was defending against something I had never shown could happen, and I have taken it out rather than move it into the method. If you know a shape of database that does make it raise, I would rather see that case and fix the method properly than put the guard back.

Your point about `param` being `None` I have taken on your word, since none of the databases in `examples/` produce it — the check is in with your comment.

**`env_datas`, not `environment_datas`**

Applied, with one correction: the attribute is `env_datas`.

```
ddd_spec attributes: … env_data_descs  env_datas  … structures  tables …
```

`environment_datas` is not there on any of the four layers in `somersault.pdx`. So it reads `chain(ddd_spec.structures, ddd_spec.env_datas)`.

**The overlap condition — I think this one is already right**

The suggestion was:

```python
if (a_start < b_end and b_start < a_end) or (b_start < a_end and a_start < b_end):
```

The second half is the first half with the operands swapped, so the whole thing reduces to `a_start < b_end and b_start < a_end` — the same expression that is there now. The test is symmetric, and `b` being physically before `a` is covered:

| case | a | b | expected | current |
| --- | --- | --- | --- | --- |
| a before b, disjoint | (0,8) | (8,16) | False | False |
| **b before a, disjoint** | (8,16) | (0,8) | False | False |
| a before b, overlapping | (0,8) | (4,12) | True | True |
| **b before a, overlapping** | (4,12) | (0,8) | True | True |
| b inside a | (0,16) | (4,8) | True | True |
| a inside b | (4,8) | (0,16) | True | True |
| identical | (0,8) | (0,8) | True | True |

I have left it as it was. If the concern was about something other than ordering I have misread it, and I am happy to be told so.

**Left alone**

The recursive walk over arbitrary attributes: you flagged the reference-resolution risk yourself, and I would rather not open that in a first rule.

On the 250 rules in appendix B.2 — understood, and the point of this shape is that the next one is a module and a line in `DEFAULT_RULES`. I am not proposing to write them all, but I am happy to keep adding the ones where I can produce a database that the rule catches and a conforming one that it stays quiet about.

`mypy .`, `ruff check .` and the suite are clean: 331 files, 174 passed, 1 skipped.

andlaus:
@Poseidonas:  please deal down the verbosity of your LLM a bit, it is a lot of reading and I don't want to go through it apologizing for mistakes...

> Your point about param being None I have taken on your word, since none of the databases in examples/ produce it — the check is in with your comment.

that's because all example PDX files ought to be valid (or at least loadable in strict mode)

> The second half is the first half with the operands swapped, so the whole thing reduces to a_start < b_end and b_start < a_end — the same expression that is there now. The test is symmetric, and b being physically before a is covered:

that would only be true if the inner loop went over all parameters, not just `[i+1:]`. right?

andlaus:
> @Poseidonas: please deal down the verbosity of your LLM a bit, it is a lot of reading and I don't want to go through it apologizing for mistakes...

or better still: make it reply to the individual review comments and close the ones which are resolved...

Poseidonas:
Fair enough on the length, I'll keep it shorter.

You're right about the examples being valid, that's why I couldn't reproduce it. The None check is in.

On the loop, I think it does hold. `[i+1:]` means each pair gets looked at once, but the comparison itself doesn't care which of the two was declared first. I put two overlapping parameters in reverse order to check:

```
declared in position order  -> found
declared in reverse         -> found
reverse, not overlapping    -> nothing
```

Happy to switch to your version if you'd rather have it written out explicitly.

andlaus:
> Happy to switch to your version if you'd rather have it written out explicitly.

please do

odxtools/checks/overlapping_parameters.py
```diff
@@ -0,0 +1,121 @@
...
+def _overlaps(placed: list[tuple[tuple[int, int],
+                                 Parameter]]) -> Iterator[tuple[Parameter, Parameter]]:
+    for i, ((a_start, a_end), a) in enumerate(placed):
+        for (b_start, b_end), b in placed[i + 1:]:
+            if a_start < b_end and b_start < a_end:
+                yield a, b
+
+
+class OverlappingParameters:
+    """Two parameters of one structure which cover the same bits.
```
kayoub5: as far as I know, there is nothing against overlapping parameters in the standard.

in practice, I have seen multiple pdx files that put multiple response/decode parameters for the same bits

Example:
* One byte for temperature
* The same byte is mapping to a text warning parameter, based on the temperature value range

odxtools/checks/overlapping_parameters.py
kayoub5: a better approach would be to use the OdxLinkDatabase, since it has all objects in a flat dict, plus the objects are keyed by their origin XML document

odxtools/checks/finding.py
```diff
@@ -0,0 +1,42 @@
...
+    will make some operation on it fail or produce a wrong result. ``WARNING``
+    is for something which is suspect but which the specification permits.
+    ``INFO`` and ``DEBUG`` are for things worth saying but not worth acting on,
+    such as an object a rule decided not to inspect.
+    """
+
+    ERROR = "error"
+    WARNING = "warning"
+    INFO = "info"
+    DEBUG = "debug"
```
kayoub5: typically a severity is assigned to a number, since that allows comparing levels, and easier logic when filtering by severity ( filter on info and above for example)

odxtools/checks/finding.py
```diff
@@ -0,0 +1,42 @@
...
+
+
+@dataclass(frozen=True, kw_only=True)
+class Finding:
```
kayoub5, line 32: the finding text should include the doc, short_name and id of the problematic element

odxtools/checks/rule.py
```diff
@@ -0,0 +1,19 @@
...
+from collections.abc import Iterable
+from typing import Protocol, runtime_checkable
+
+from ..database import Database
+from .finding import Finding
+
+
+@runtime_checkable
+class Rule(Protocol):
+    """A single consistency rule."""
```
kayoub5, line 16: a rule should include a field that describes what part of the spec it validates

Poseidonas:
@kayoub5 you're right, and I should have checked before proposing this. The codebase itself says it plainly in `nrcconstparameter.py` — an NRC-CONST is meant to overlap with a VALUE parameter, with a reference to the standard. I built the rule on an assumption I never verified, and your temperature example is the same pattern.

I'm closing this. A checker that reports false errors on real files is worse than no checker. If the idea is ever picked up again, your comments already sketch the right shape for it: rules taken from appendix B.2, each declaring which part of the spec it checks — not invented ones.

@andlaus sorry for the review time this cost you both. I came here to learn how a real ODX codebase works, not to point a tool at someone else's repository — and getting the premise right was my job, not yours.

odxtools/checks/overlapping_parameters.py
andlaus: > as far as I know, there is nothing against overlapping parameters in the standard.

anyway, I think this should be flagged on at least INFO level...

odxtools/checks/finding.py
andlaus: the issue with this is that the internalized dataclasses do not provide that info (yet?). IMO that's a topic for a future PR...

andlaus:
@Poseidonas: I want to have this functionality; that said, I am more interested in the framework to implement checks and the shrubbery to plug it into the CLI than in the overlapping parameters check. Having a check for overlapping parameters is not a bad idea either as long as it does not produce errors and the check can be disabled by the user...

andlaus:
also, the logic for `NrcConstParameter`s could be inverted: produce a warning if such a parameter does *not* overlap with a `ValueParameter`...

odxtools/checks/__init__.py
Poseidonas: Done — `path: severity: message [rule]`.

odxtools/checks/overlapping_parameters.py
Poseidonas: Applied, reworded to "Check whether …" since the rule no longer asserts a violation.

odxtools/checks/overlapping_parameters.py
Poseidonas: The guard is gone. Measured across both example files (164 + 212 parameters, strict and non-strict): zero raises, so there was nothing to fix in the method either.

odxtools/checks/overlapping_parameters.py
Poseidonas: Applied.

odxtools/checks/__init__.py
Poseidonas: Both added, with numeric values following the `logging` convention.

odxtools/checks/overlapping_parameters.py
Poseidonas: Done — a DEBUG finding reports how many parameters were not compared.

odxtools/cli/check.py
Poseidonas: Applied.

odxtools/checks/__init__.py
Poseidonas: Applied.

odxtools/checks/overlapping_parameters.py
Poseidonas: Written out as given.

odxtools/checks/overlapping_parameters.py
Poseidonas: Applied, adjusted for the link-database enumeration.

odxtools/checks/overlapping_parameters.py
Poseidonas: This code is gone — replaced by the link-database enumeration.

odxtools/checks/overlapping_parameters.py
Poseidonas: Handled by the link-database enumeration: objects are deduplicated, so each is inspected exactly once.

odxtools/checks/__init__.py
Poseidonas: Done — imports at the top, `noqa` gone.

odxtools/checks/__init__.py
Poseidonas: Done — `finding.py`, `rule.py`, one module per rule.

odxtools/checks/__init__.py
Poseidonas: Renamed.

odxtools/checks/overlapping_parameters.py
Poseidonas: This code is gone — replaced by the link-database enumeration.

odxtools/checks/overlapping_parameters.py
Poseidonas: This code is gone — replaced by the link-database enumeration.

odxtools/checks/overlapping_parameters.py
Poseidonas: Went with the link database instead — it reaches the same objects without the reference-resolution risk you mention.

odxtools/checks/overlapping_parameters.py
Poseidonas: You were right. The rule now reports overlaps at INFO, can be disabled, and its docstring says the standard permits them — somersault itself now carries one as a live example.

odxtools/checks/overlapping_parameters.py
Poseidonas: Done — and it fixed a real bug: the attribute walk handed out weakref proxies which fail `isinstance` against the runtime-checkable protocol, so only 3 of 25 codecs were inspected. A test pins 25/25 now.

odxtools/checks/finding.py
Poseidonas: Done — `IntEnum` with the `logging` module's values; the CLI filters with `--severity`.

odxtools/checks/finding.py
Poseidonas: The location carries the short-name path; doc and id can follow once the dataclasses expose them.

odxtools/checks/rule.py
Poseidonas: Done — `spec` on the protocol and on both rules.

Poseidonas:
Reopened with the rework on top. Overlaps are now INFO and any rule can be disabled with `--disable-rule`; severities are numeric following the `logging` convention, rules carry a `spec` field, and objects are enumerated through the link database — which also fixed a real bug: the attribute walk handed out weakref proxies that fail `isinstance` against the runtime-checkable protocol, so only 3 of 25 codecs were being inspected. A test pins 25/25 now.

The inverted NRC-CONST rule is in as `nrc-const-without-value` — and it immediately flagged somersault itself: `flips_not_done`'s `reason` has no VALUE parameter over its byte, so no reason can actually be encoded (a directly set value raises `EncodeError`, otherwise the byte stays zero). The second commit gives it an overlapping `reason_value`, which also gives the overlap rule a live INFO example in the reference file. If you'd rather cure the example differently, that commit stands alone and is easy to drop.

The individual review comments are answered and resolved.

odxtools/checks/finding.py
```diff
@@ -0,0 +1,48 @@
...
+    WARNING = 30
+    ERROR = 40
+
+
+@dataclass(frozen=True, kw_only=True)
+class Finding:
+    """A single thing a rule reports."""
+
+    #: name of the rule which produced this finding
+    rule: str
```
kayoub5, line 36: why not the rule object itself ?

odxtools/checks/finding.py
```diff
@@ -0,0 +1,48 @@
...
+
+    severity: Severity
```
kayoub5, line 41: severity is a property of the rule or the finding ?

odxtools/checks/finding.py
```diff
@@ -0,0 +1,48 @@
...
+
+    #: where in the database the finding is, as a path of short names
+    #: starting with the name of the document which defines the object,
+    #: e.g. ``("somersault", "flips_not_done")``
+    location: tuple[str, ...] = field(default_factory=tuple)
```
kayoub5: type hint here too vague

odxtools/checks/finding.py
```diff
@@ -0,0 +1,48 @@
...
+
+    #: what was found, in one sentence, naming the objects involved
+    message: str
+
+    def __str__(self) -> str:
+        where = "/".join(self.location)
+        prefix = f"{where}: " if where else ""
```
kayoub5: why would location be empty ?

odxtools/checks/nrc_const_without_value.py
```diff
@@ -0,0 +1,61 @@
...
+    left at zero — measured on ``somersault.pdx``, where encoding
+    ``flips_not_done`` cannot choose a reason.
+    """
+
+    name = "nrc-const-without-value"
+
+    #: The expectation comes from the NRC-CONST semantics of the standard:
+    #: the parameter constrains bits which an overlapping VALUE parameter
+    #: is supposed to expose (ASAM MCD-2 D, p. 77-79).
+    spec = "NRC-CONST semantics, ASAM MCD-2 D p. 77-79"
```
kayoub5: I think using chapter is better than page

ODX have two active standards, 2.0 and 2.2, which one is this referencing ?

odxtools/checks/overlapping_parameters.py
```diff
@@ -0,0 +1,117 @@
...
+
+def _overlaps(placed: list[tuple[tuple[int, int],
+                                 Parameter]]) -> Iterator[tuple[Parameter, Parameter]]:
+    for i, ((a_start, a_end), a) in enumerate(placed):
+        for (b_start, b_end), b in placed[i + 1:]:
+            if (a_start < b_end and b_start < a_end) or (b_start < a_end and a_start < b_end):
+                yield a, b
+
+
+def iter_codecs(database: Database) -> Iterator[tuple[str, CompositeCodec]]:
```
kayoub5: you can make this generic for future checks by making it accept a type argument

odxtools/checks/__init__.py
```diff
@@ -0,0 +1,64 @@
...
+from .finding import Finding, Severity
+from .nrc_const_without_value import NrcConstWithoutValue
+from .overlapping_parameters import OverlappingParameters
+from .rule import Rule
+
+__all__ = [
+    "DEFAULT_RULES",
+    "Finding",
+    "NrcConstWithoutValue",
+    "OverlappingParameters",
```
kayoub5: why are the individual rules exposed ?

odxtools/checks/finding.py
Poseidonas: A name keeps `Finding` a plain value — printable, comparable, serializable — and it is what `--disable-rule` and the `[rule]` suffix already use, the way linters name their rules. The rule object stays reachable through `DEFAULT_RULES` when a caller needs it.

odxtools/checks/finding.py
Poseidonas: Of the finding: both current rules emit two levels each — their main severity plus a DEBUG note for what could not be inspected — so it cannot live on the rule.

odxtools/checks/finding.py
Poseidonas: It would not be — you are right, and by the same standard as the removed `try/except`: no rule produces an empty location. It is required now and the guard is gone.

odxtools/checks/overlapping_parameters.py
Poseidonas: Done — `iter_objects(database, of_type)` in its own module, with `iter_codecs` kept as the typed convenience both rules use.

odxtools/checks/__init__.py
Poseidonas: No good reason — removed. The public names are now `run_checks`, `DEFAULT_RULES`, `Finding`, `Severity`, `Rule`; the rule classes remain importable from their own modules.

odxtools/checks/finding.py
Poseidonas: What shape would you prefer? The location already gives the document name and the short-name path; the id is what the internalized dataclasses do not expose yet — the part @andlaus parked for a future PR — so I did not want to design that type prematurely.

odxtools/checks/nrc_const_without_value.py
Poseidonas: The page reference mirrors `nrcconstparameter.py`, which says "See ASAM MCD-2 D (ODX), p. 77-79" — I work from the codebase, so I cannot name the chapter or the revision myself. If you can, I will gladly put the chapter into the rule, and the docstring it mirrors could follow.

odxtools/checks/finding.py
kayoub5: typically linters assign a severity to the rule, and allow users to customize a rule severity

Poseidonas: That fits what is here more than it first seems: each rule's main findings already use a single level — INFO for overlaps, WARNING for the NRC rule — so that level can become a rule attribute with a user override, say `--rule-severity overlapping-parameters=warning`, while the finding keeps recording what was actually emitted, since the skip-notes stay at DEBUG regardless of configuration. Happy to add the override in this PR or as a follow-up, whichever you prefer.

odxtools/checks/finding.py
kayoub5: OdxLinkDatabase provides extra info that can be used to locate the element

namely

* OdxDocFragment: which XML file the element originated from
* ODX ID

Most objects in the database are OdxNamed, even if the validation applies to a non-named element, one of its parents would certainly be a OdxNamed

so instead of a generic location, specific doc, element id, element name fields would be better.

Poseidonas: Done — findings now carry `doc`, `element_name` and `element_id`, taken from the link database as you suggested:

```
somersault/flips_not_done (somersault.NR.flips_not_done): warning: ...
```

One note: the fragment gives the document name, not the physical file name — that is what the link database has.

odxtools/checks/finding.py
kayoub5: I am ok to defer rule override to another PR

odxtools/checks/nrc_const_without_value.py
kayoub5: @andlaus can you help here? I am on mobile only (taking a vacation)

kayoub5 (approved):

odxtools/checks/nrc_const_without_value.py
andlaus: > @andlaus can you help here?

that's section 7.3.5.4 of ODX 2.2 (ASAM version, not ISO), but since it is pretty long, keeping the relevant page (79) is not the worst idea. the relevant paragraph reads:

> The NRC-CONST is located at the same PDU position as the VALUE parameter. As a VALUE parameter is never used to match a RESPONSE, the parameter (and its associated TEXTTABLE) does not suffice to select the response.

> (taking a vacation)

enjoy! :)

odxtools/checks/finding.py
andlaus: > typically linters assign a severity to the rule, and allow users to customize a rule severity

I agree with @Poseidonas here: if the rule specifies a severity (like most checker rules from appendix B.2), this simply means that all findings stemming from it exhibit the given severity. On the other hand it is easy to imagine rules which produce findings of different severity, e.g., a maximum size checker could produce a debug or info finding if the maximum size of a structure could not be determined ahead of time and an error if it can be and it is larger than the maximum allowed...

odxtools/checks/finding.py
```diff
@@ -0,0 +1,51 @@
...
+    severity: Severity
+
+    #: name of the document which defines the element the finding is about
+    doc: str
+
+    #: short name of that element
+    element_name: str
+
+    #: the ODX ID the element is registered under in its document
+    element_id: str
```
andlaus: make this `odx_id` like for identifiable elements:
```suggestion
    odx_id: OdxLinkId|None
```

(`|None` since not all relevant objects might come with an ODX ID.)

Even better: Link to the relevant object directly, i.e.,

```suggestion
    object: Any
```

(all attributes like `odx_id`, `short_name`, ... can then be retrieved from that. maybe provide some helper functions to extract commonly-used attributes like `.odx_id` or `.short_name`)

besides this, having a `.short_name_path` attribute would be helpful as it helps to locate the object in question more quickly...

odxtools/odxlink.py
```diff
@@ -191,6 +191,18 @@ def __init__(self, *, use_weakrefs: bool = False) -> None:
         self._db: dict[OdxDocFragment, dict[str, Any]] = {}
         self.use_weakrefs = use_weakrefs
 
+    def objects(self) -> Iterator[tuple[OdxDocFragment, str, Any]]:
```
andlaus: All objects contained in the `OdxLinkDatabase` are `IdentifiableElement`s, i.e.,
```suggestion
    def objects(self) -> Iterator[IdentifiableElement]:
```

the remaining tuple entries can be retrieved from a yielded object `x` via `x.odx_id`,  `x.odx_id.local_id`, and `x.odx_id.doc_fragments`.

Besides this, each object should only be yielded once. This is currently not the case because some objects can be present in multiple document fragments (e.g., DIAG-LAYER-CONTAINER as well as DIAG-LAYER)

odxtools/checks/objects.py
```diff
@@ -0,0 +1,32 @@
...
+from collections.abc import Iterator
+from typing import TypeVar
+
+from ..compositecodec import CompositeCodec
+from ..database import Database
+
+T = TypeVar("T")
+
+
+def iter_objects(database: Database, of_type: type[T]) -> Iterator[tuple[str, str, T]]:
```
andlaus: IMO it is better to iterate over `db.odxlinks.objects()` directly.

odxtools/checks/__init__.py
```diff
@@ -0,0 +1,62 @@
...
+    OverlappingParameters(),
+    NrcConstWithoutValue(),
+)
+
+
+def run_checks(
+        database: Database,
+        rules: Iterable[Rule] | None = None,
+        *,
+        disabled: Iterable[str] = (),
```
andlaus: specifying the applicable `rules` only to disable some of them in the same call is a good recipe for a Rube-Goldstein machine: only pass the `rules` argument here and create this list in the calling code. (Not passing the `rules` argument should IMO be seen as a convenience for users which quickly want to run all checks. the `check` CLI tool can invest the effort to do this properly...

andlaus (commented):
haven't looked at the unit test, but we're getting closer...

odxtools/checks/nrc_const_without_value.py
Poseidonas: Put the section into `spec` (`ASAM MCD-2 D (ODX) 2.2, section 7.3.5.4, p. 79`) and the quoted paragraph into the rule's docstring; the overlap rule cites the same section now.

odxtools/checks/finding.py
Poseidonas: Nothing to change in the model then: the rule declares no severity and each finding carries its own. I added a one-line comment on `Finding.severity` saying so.

odxtools/checks/finding.py
Poseidonas: `Finding` now holds `object`, with `odx_id`, `short_name` and `short_name_path` as accessors that return `None` where the object has none. Since objects do not know their parents, `short_name_path` is the doc-fragment names from the ODX ID followed by the short name, e.g. `somersault.somersault_base_variant.flips_not_done`; if there is a better source for the path I am happy to switch.

odxtools/odxlink.py
Poseidonas: Done: `Iterator[IdentifiableElement]`, each object yielded once, with a test for an object registered in a CONTAINER and a LAYER fragment. One observation: on somersault, `EcuVariant`, `BaseVariant`, `Protocol` and `TableKeyParameter` are in the database but are not `IdentifiableElement` subclasses, though they carry `odx_id` and `short_name`; I kept the annotation you suggested.

odxtools/checks/objects.py
Poseidonas: Removed the module; both rules iterate `database.odxlinks.objects()` directly and select codecs by their concrete classes — `isinstance` against the runtime-checkable protocol evaluates its properties on Python ≤ 3.11 and silently skipped malformed objects, which CI caught on 3.10.

odxtools/checks/__init__.py
Poseidonas: `run_checks` takes only `rules` now (`DEFAULT_RULES` unless given). The `check` command builds the list from `--disable-rule`, and unknown names are rejected by argparse via `choices`.

odxtools/cli/check.py
```diff
@@ -49,16 +50,13 @@ def run(args: argparse.Namespace) -> None:
     odxdb = _parser_utils.load_file(args)
 
     threshold = Severity[args.severity.upper()]
+    rules = [rule for rule in DEFAULT_RULES if rule.name not in args.disable_rule]
```
andlaus: the `check` tool should also have a command line argument to print all available rules and a one-line description for each of them, e.g., `--list-available-rules`. when specified, the tool should print the list and exit.

odxtools/cli/check.py
```diff
@@ -39,6 +39,7 @@ def add_subparser(subparsers: SubparsersList) -> None:
         action="append",
         default=[],
         metavar="NAME",
+        choices=[rule.name for rule in DEFAULT_RULES],
```
andlaus: listing all rule names here is not a good idea as there will potentially dozends to hundreds. Also an argument to specify a list of rules a la `--disable-rules param_overlap odx12 odx23 odx175` would be better IMO.

odxtools/cli/check.py
Poseidonas: Done in 329b5ed: `odxtools check --list-available-rules` prints one line per rule — the name, then a one-line `description` each rule now carries next to its `spec` — and exits without needing a file.

odxtools/cli/check.py
Poseidonas: Done in fb7b0e1: the option is now `--disable-rules NAME [NAME ...]`; the names are checked in code against the registered rules and every unknown one is listed in the usage error, and the help no longer enumerates the rules but points at `--list-available-rules`.

Issue #505, opened 2026-08-23
write_pdx_file() emits a corrupt PDX from the second call onwards (regression in 11.4.2)


ken-otf:
## Summary

`write_pdx_file()` produces a **corrupt PDX from the second call onwards** within the same process. The first call is fine; every subsequent call emits a file whose `COMPARAM-SUBSET-REF` has lost its `DOCREF`/`DOCTYPE` attributes, so loading it back fails with an unresolvable ODXLINK reference.

This is a regression: **last good = 11.4.1, first bad = 11.4.2.**

It is independent of whether the database was modified — writing the same unmodified database twice is enough to trigger it.

## Reproducer

```python
import sys, tempfile, os, odxtools

src = sys.argv[1]          # any PDX; we used the somersault example
d = tempfile.mkdtemp()
print("odxtools", odxtools.__version__)
for i in (1, 2, 3):
    db = odxtools.load_pdx_file(src)
    out = os.path.join(d, f"out{i}.pdx")
    odxtools.write_pdx_file(out, db)
    try:
        odxtools.load_pdx_file(out)
        print(f"  write #{i}: reload OK")
    except Exception as e:
        print(f"  write #{i}: reload FAILED -> {type(e).__name__}: {str(e)[:90]}")
```

Output on 11.5.3:

```
odxtools 11.5.3
  write #1: reload OK
  write #2: reload FAILED -> KeyError: "ODXLINK reference OdxLinkRef('SAE_J2411_SWCAN') could not be resolved for any of the docu
  write #3: reload FAILED -> KeyError: "ODXLINK reference OdxLinkRef('SAE_J2411_SWCAN') could not be resolved for any of the docu
```

Note that each iteration loads a **fresh** database from the original file, so no state is carried over on the caller's side.

## What differs in the output

Diffing `ISO_15765_3_on_ISO_15765_2.odx-c` between write #1 and write #2:

```diff
@@ -14 +14 @@
-      <TEAM-MEMBER-REF ID-REF="CPS_ISO_15765_3_on_ISO_15765_2.Softing.HC" />
+      <TEAM-MEMBER-REF ID-REF="CPS_ISO_15765_3_on_ISO_15765_2.Softing.HC" DOCREF="ISO_15765_3_on_ISO_15765_2" DOCTYPE="COMPARAM-SPEC" />
@@ -138 +138 @@
[...]

PR #506, opened 2026-08-24
Revert "perf(writepdxfile): cache Jinja2 environment to avoid re-parsing templates"

andlaus:
This reverts commit 413e3036aa241a05feb42973496820d19c897d73. Turns out that this [broke the correctness of repeated writes](https://github.com/mercedes-benz/odxtools/issues/505) from within the same proccess (but at least they were fast ;))


seif-galal:
@andlaus well its fast but now i have to adjust the approach in PRs to maintain balance.

Issue #508, opened 2026-08-25
bare except swallows KeyboardInterrupt and SystemExit


seif-galal:
```markdown
**Bug:** `except:` in `_validate_string_value` (line 58) catches `BaseException`, including `KeyboardInterrupt` and `SystemExit`.

**Location:** `odxtools/cli/browse.py`, function `_validate_string_value`

**Root Cause:** Python's bare `except:` clause catches `BaseException`, which is the superclass of **all** exceptions including:
- `KeyboardInterrupt` — raised when user presses Ctrl+C
- `SystemExit` — raised by `sys.exit()` calls
- `GeneratorExit` — raised when a generator's `close()` is called
- `Exception` — the normal errors we actually want to catch

**Reproducer:**
```python
>>> try:
...     try:
...         raise KeyboardInterrupt
...     except:
...         print("caught")  # <-- This branch runs
... except KeyboardInterrupt:
...     print("propagated")  # <-- Never reached
caught
```

**Impact on `odxtools browse`:**
When a user runs `odxtools browse` and is prompted for parameter values, pressing Ctrl+C to abort is silently swallowed. The validation returns `False`, and the CLI continues prompting for the next parameter instead of exiting. This violates the [Principle of Least Astonishment](https://en.wikipedia.org/wiki/Principle_of_least_astonishment) — users expect Ctrl+C to terminate the program immediately.

Additionally, any genuine bugs (e.g., `AttributeError` from a missing attribute, `TypeError` from incorrect API usage) are masked, making debugging impossible.

**What exceptions CAN legitimately occur here:**
1. `ValueError` — from `_convert_string_to_odx_type` when input is invalid (e.g., non-hex string for `A_BYTEFIELD`)
2. `TypeError` — from `odxrequire` when `parameter.physical_type` is `None`

**What exceptions MUST NOT be caught:**
- `KeyboardInterrupt` — user intent to abort
- `SystemExit` — program termination request
- `RuntimeError`, `AttributeError`, etc. — real bugs that should crash loudly

**Fix:**
```diff
-    except:  # noqa: E722
+    except (ValueError, TypeError):  # noqa: E722
         return False
```

**Why not `except Exception:`?**
[...]

PR #509, opened 2026-08-28
replace bare except with OdxError, ValueError, TypeError

seif-galal:
Bare except swallowed KeyboardInterrupt and SystemExit, preventing Ctrl+C from working during parameter validation prompts.

Fixes: issue #508

### DIFF
--- a/odxtools/cli/browse.py
+++ b/odxtools/cli/browse.py
@@ -14,7 +14,7 @@ from ..diaglayers.diaglayer import DiagLayer
 from ..diaglayers.hierarchyelement import HierarchyElement
 from ..diagservice import DiagService
 from ..dopbase import DopBase
-from ..exceptions import odxraise, odxrequire
+from ..exceptions import OdxError, odxraise, odxrequire
 from ..odxlink import resolve_snref
 from ..odxtypes import AtomicOdxType, DataType, ParameterValueDict
 from ..parameters.matchingrequestparameter import MatchingRequestParameter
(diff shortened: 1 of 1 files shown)


odxtools/cli/browse.py
```diff
@@ -55,7 +55,7 @@ def _validate_string_value(input: str, parameter: Parameter) -> bool:
         try:
             phys_type = odxrequire(parameter.physical_type)
             val = _convert_string_to_odx_type(input, phys_type.base_data_type)
-        except:  # noqa: E722
+        except (ValueError, TypeError):  # noqa: E722
```
kayoub5: are you sure odxrequire doesn't throw another type of exception ?

seif-galal: good catch - ill double check if it is the case ill amend and push if not ill be back here with an explanation

seif-galal: i updated the commit just now + removed the # noqa: E722 ignore, check diff

andlaus:
color me skeptical here: `browse` is an interactive tool; good luck pressing Ctrl+C during the 0.5 miliseconds when this function is called. On the flipside, I really do not want to risk making the code less robust, as demonstrated by #489.

@kayoub5: please merge if you are in favor of this, else close...

seif-galal:
> color me skeptical here: `browse` is an interactive tool; good luck pressing Ctrl+C during the 0.5 miliseconds when this function is called. On the flipside, I really do not want to risk making the code less robust, as demonstrated by #489.
> 
> @kayoub5: please merge if you are in favor of this, else close...

@andlaus I understand. I added this to help cover any possible scenarios i can think of. So in order to avoid any further demonstrations such as #489 ill do my analysis then open an issue……

+ i have another question and it is use my arduino and a can module as an ECU implementation (with some software..) on my side to test odxtools, is it possible with this repo?

PR #511, opened 2026-08-28
fix(cli/browse): use sys.stdout.isatty() instead of sys.__stdout__

### LINKED ISSUE #510: browse.py crashes in headless environments instead of raising SystemError
**Bug:** `browse.py` should raise `SystemError` in non-interactive mode, but instead crashes with `NoConsoleScreenBufferError`.

**Problem:** `IP_prompt(selection)` calls `prompt_toolkit` which requires a real Windows console buffer. In headless environments (pytest, CI, SSH), `GetConsoleScreenBufferInfo` fails and raises `NoConsoleScreenBufferError`. The code does not check for interactive mode before calling `IP_prompt`.

**Expected behavior:** The test expects `SystemError`:
tests\test_cli.py:140:
with self.assertRaises(SystemError):
    # browse can only be run interactively
    browse.run(browse_args)

**Actual Behavior**
prompt_toolkit.output.win32.NoConsoleScreenBufferError: No Windows console found. Are you running cmd.exe?

fix by adding isatty() check before IP_prompt:
import sys

In browse() function, before IP_prompt:
if not sys.stdout.isatty():
    raise SystemError("browse can only be run interactively")

seif-galal:
browse.py checked sys.__stdout__.isatty() but prompt_toolkit checks sys.stdout.isatty(). In headless environments (CI, pytest), the test mocks sys.stdout.isatty to False, but sys.__stdout__ was still True, so IP_prompt ran and crashed with NoConsoleScreenBufferError.

Fix: Use sys.stdout.isatty() to match what prompt_toolkit checks.

Fixes: #510

### DIFF
--- a/odxtools/cli/browse.py
+++ b/odxtools/cli/browse.py
@@ -129,7 +129,7 @@ def prompt_single_parameter_value(parameter: Parameter) -> AtomicOdxType | None:
 def encode_message_interactively(codec: Request | Response,
                                  ask_user_confirmation: bool = False) -> None:
     if sys.__stdin__ is None or sys.__stdout__ is None or not sys.__stdin__.isatty(
-    ) or not sys.__stdout__.isatty():
+    ) or not sys.stdout.isatty():
         raise SystemError("This command can only be used in an interactive shell!")
 
     answered_request = b''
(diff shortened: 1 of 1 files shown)


andlaus:
this seems fine. thanks for the catch!

PR #513, opened 2026-09-04
`browse`: add support for [Home], [End], [PageUp], and [PageDown] keys for list selection

andlaus:
this makes navigating long lists much more convenient.

### DIFF
--- a/odxtools/cli/browse.py
+++ b/odxtools/cli/browse.py
@@ -2,13 +2,10 @@
 import argparse
 import logging
 import sys
-from typing import Any, cast
+from typing import cast
 
-from InquirerPy.prompts.list import ListPrompt
 from InquirerPy.resolver import prompt as IP_prompt
 from InquirerPy.resolver import question_mapping
-from InquirerPy.separator import Separator
-from prompt_toolkit.keys import Keys
 from rich import print as rich_print
 
 from ..complexdop import ComplexDop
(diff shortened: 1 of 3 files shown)


tests/test_dop_type_diagnostics.py
```diff
@@ -8,12 +8,11 @@
 import unittest
 
 import odxtools.exceptions as exceptions
-
 from odxtools.exceptions import OdxError
-from odxtools.odxlink import (DocType, OdxDocFragment, OdxLinkDatabase, OdxLinkId, OdxLinkRef)
+from odxtools.nameditemlist import NamedItemList
```
andlaus, line 12: note: the changes to this file were caused by #509 not running the `format-source.sh` script. IMO this does not warrant a separate PR...

odxtools/cli/browse.py
```diff
@@ -31,6 +35,57 @@
 _odxtools_tool_name_ = "browse"
 
 
+class _ListPromptWithCustomKeys(ListPrompt):  # type: ignore[misc,unused-ignore]
```
kayoub5, line 38: browse.py file is not small, so having this class here feels like a big file trying to take even more responsibility

maybe move this class to its own file?

odxtools/cli/browse.py
```diff
@@ -31,6 +35,57 @@
...
+    """InquirerPy prompt for lists that adds support for PageDown, PageUp, Home and End keys."""
```
kayoub5, line 39: Why not contribute this back to InquirerPy ?

odxtools/cli/browse.py
andlaus: I can do this, but I would like to do it thoroughly, so I suggest a separate refactoring PR...

odxtools/cli/browse.py
andlaus: the [InquirerPy repo](https://github.com/kazhala/InquirerPy)  seems to be pretty dead: the last commit is from 2022 and there are open PRs from 2023 and 2024...

odxtools/cli/browse.py
kayoub5: why refactor? only this newly added class goes into its own file

odxtools/cli/browse.py
kayoub5: I would question using it as a dependency then.

maybe consider modern alternatives like https://rich.readthedocs.io/en/stable/prompt.html

odxtools/cli/browse.py
andlaus: my intention is to split the file into multiple sub-modules and put all of them in the `odxtools/cli/browse` directory...

odxtools/cli/browse.py
andlaus: sure. but that's a much larger project.

kayoub5: others

https://github.com/magmax/python-inquirer

https://github.com/tmbo/questionary

odxtools/cli/browse.py
kayoub5: not worth it, after new class moved, only about 500 lines remain, which is acceptable file size

odxtools/cli/browse.py
andlaus: I played a bit with `rich.Prompt`: IMO it is not fit for our purpose because it does not support to interactively selecting a choice -- instead, you either have to print a enumerated list of choices and the user has to type the number, or the user has to type the entire string of the choice. 

 the last commit in `python-inquirer` is from October 2025, not sure if we would gain much if we switched to it.

For `questionary`, it seems to be a relatively small project with just 652 commits, but I guess I should look into it to determine if it is feasible.

andlaus: my LLM suggested to use [`prompt_toolkit`](https://github.com/prompt-toolkit/python-prompt-toolkit/commits/main/) directly.  AFAICS, `InquirerPy` is just an abstraction layer on top of it, so this might be an alternative...

odxtools/cli/browse.py
andlaus: ok, done: https://github.com/mercedes-benz/odxtools/pull/513/changes/b399ec88748601dbfd159174d41689c055a0e6d4 . Thanks for bringing this up!

odxtools/cli/browse.py
kayoub5: that sounds like something that you should do eventually, keeping a dead library as dependency is problematic, maybe new issue for that?

andlaus: IMO, we should lean back and do nothing as long as `InquirerPy` still fits our purpose, i.e., once it breaks we should spend the mental energy of thinking about how to replace it...

andlaus:
do you see any issues preventing this from getting merged, @kayoub5?

odxtools/cli/browse.py
kayoub5: I am not saying do it now, just a ticket in the backlog

kayoub5:
> do you see any issues preventing this from getting merged, @kayoub5?

The PR changes look good to merge

odxtools/cli/browse.py
andlaus: done: https://github.com/mercedes-benz/odxtools/issues/518

Issue #515, opened 2026-09-04
Bug: incorrect SNREF/SNPATHREF validation in StateTransitionRef and PreConditionStateRef


seif-galal:
In both `StateTransitionRef` and `PreConditionStateRef`, the validation performed when `VALUE` is specified checks `in_param_if_snref` twice:

self.in_param_if_snref is not None or self.in_param_if_snref is not None

The second operand appears to be intended to check `in_param_if_snpathref`.

As a result, a reference provided exclusively through `IN-PARAM-IF-SNPATHREF` does not satisfy this validation.

Files:
- `odxtools/statetransitionref.py`
- `odxtools/preconditionstateref.py`

A local correction changes the second operand to `in_param_if_snpathref`.

andlaus:
sounds correct, please make a PR. (everything related to SNPATHREFs is currently basically untested, even the error checks...)

PR #516, opened 2026-09-06
Fix SNPATHREF validation in PreConditionStateRef

seif-galal:
Description:

Fix the SNPATHREF validation in `PreConditionStateRef`.

When `VALUE` is specified, the validation checks `in_param_if_snref` twice. The second check should be `in_param_if_snpathref`.

This allows a reference supplied exclusively through `IN-PARAM-IF-SNPATHREF` to satisfy the validation.

Related issue: #515


andlaus:
thanks

Issue #514, opened 2026-09-04
Bug: Incorrect tuple-length validation in statetransitionref.py


seif-galal:
In statetransitionref.py, the validation for TableStructParameter currently contains:

if not isinstance(inner_param_value, tuple) or len(inner_param_value) == 2:
    odxraise("Invalid value for table struct parameter")

The len(...) == 2 condition is inverted.

A TableStructParameter expects a tuple containing exactly two values, so the validation should reject tuples whose length is not 2.

Proposed fix

Change:

len(inner_param_value) == 2

to:

len(inner_param_value) != 2

andlaus:
true, make it an unequal.

PR #517, opened 2026-09-06
Fix table parameter tuple validation

seif-galal:
Description

Fix the tuple validation for `TableStructParameter` values in `StateTransitionRef`.

The current validation incorrectly rejects valid 2-element tuples because it checks `len(inner_param_value) == 2`.

This changes the condition to reject tuples whose length is not 2.

Issue #514 and #514

### DIFF
--- a/odxtools/statetransitionref.py
+++ b/odxtools/statetransitionref.py
@@ -60,7 +60,7 @@ def _resolve_in_param_helper(
 
     # deal with table parameters
     if isinstance(inner_param, TableStructParameter):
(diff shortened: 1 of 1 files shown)


kayoub5:
The code change looks ok, but it would be nice if it was covered by a unit test

andlaus:
To avoid tailoring the unit tests to the existing code, I think we should only test the most common error conditions in the unit tests, so IMO having this fix without a unit test is okay...

seif-galal:
@kayoub5 true. well ill create a regression test locally for the sake of it anyway.. ill comment the results soon



update: python -m pytest -q returns pass with test file

image

PR #519, opened 2026-09-07
`browse`: implement support for querying nested parameter values

andlaus:
e.g., allow specifying values for parameters which use a structure containing structures as their DOP.

### DIFF
--- a/odxtools/cli/browse.py
+++ b/odxtools/cli/browse.py
@@ -8,14 +8,16 @@ from InquirerPy.resolver import prompt as IP_prompt
 from InquirerPy.resolver import question_mapping
 from rich import print as rich_print
 
-from ..complexdop import ComplexDop
+from ..basicstructure import BasicStructure
 from ..database import Database
 from ..dataobjectproperty import DataObjectProperty
 from ..diaglayers.diaglayer import DiagLayer
 from ..diaglayers.hierarchyelement import HierarchyElement
 from ..diagservice import DiagService
-from ..dopbase import DopBase
-from ..exceptions import OdxError, odxraise, odxrequire
+from ..environmentdatadescription import EnvironmentDataDescription
+from ..exceptions import odxraise
+from ..field import Field
+from ..multiplexer import Multiplexer
 from ..odxlink import resolve_snref
 from ..odxtypes import AtomicOdxType, DataType, ParameterValueDict
 from ..parameters.matchingrequestparameter import MatchingRequestParameter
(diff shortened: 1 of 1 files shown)


odxtools/cli/browse.py
```diff
@@ -51,26 +53,30 @@ def _convert_string_to_bytes(string_value: str) -> bytes:
         return int(string_value, 16).to_bytes((int(string_value, 16).bit_length() + 7) // 8, "big")
 
 
-def _validate_string_value(input: str, parameter: Parameter) -> bool:
-    if not parameter.is_required and input == "":
+def _validate_chosen_value(input_val: AtomicOdxType, parameter: Parameter) -> bool:
+    if not parameter.is_required and (input_val == "" or input_val is None):
```
kayoub5, line 57: this creates an issue, if param has non empty default value, and user wants to send an empty string

odxtools/cli/browse.py
```diff
@@ -51,26 +53,30 @@ def _convert_string_to_bytes(string_value: str) -> bytes:
...
+                    return False
+                return dop.is_valid_physical_value(converted_val)
+            return dop.is_valid_physical_value(input_val)
         else:
-            raise NotImplementedError("Validation of complex DOPs")
+            raise NotImplementedError(f"Validation of {dop.__class__.__name__} DOPs")
     else:
-        logging.info("This value is not validated precisely: Parameter {parameter}")
-        return input != ""
+        logging.info(f"Parameter's '{parameter}' value not validated")
```
kayoub5, line 75: you mean value is not valid or value can't be validated ?

odxtools/cli/browse.py
```diff
@@ -51,26 +53,30 @@ def _convert_string_to_bytes(string_value: str) -> bytes:
...
+            if isinstance(input_val, str):
+                base_data_type = dop.physical_type.base_data_type
+                if base_data_type is None:
+                    return input_val != ""
+                try:
+                    converted_val = _convert_string_to_odx_type(input_val, base_data_type)
+                except ValueError:
+                    return False
+                return dop.is_valid_physical_value(converted_val)
+            return dop.is_valid_physical_value(input_val)
```
kayoub5, line 71: really needed ? if value can't be converted, why try to use it as physical value ?

odxtools/cli/browse.py
andlaus: true, but the issue was present before and I am not sure how it can be fixed in a non-clunky way. Possibly: if the default exists and is not the same as the empty object, ask the user if he wants the default or an empty object?

odxtools/cli/browse.py
andlaus: `dop.is_valid_physical_value()` takes a physical value and checks if it is allowed. E.g. one can specify ranges which are valid or invalid using the `ScaleConstr` class...

odxtools/cli/browse.py
andlaus: it means that we encountered a parmeter type which we do not know how to specify values for because it does not specify a DOP. The error message is probably a bit too terse...

odxtools/cli/browse.py
andlaus: ok, I implemented the "ask the user" approach for my next PR in the `browse` cleanup series. (I think that this is out of scope for this PR...)

odxtools/cli/browse.py
kayoub5: the question why the function gets passed both converted and raw input value ?

andlaus: it doesn't: it only gets converted to the physical value if the input value is a string. (if it isn't, the physical value has been directly selected from the list. This not yet used, but it will become relevant soon when implementing DTCs properly.)

kayoub5 (approved):

PR #520, opened 2026-09-10
`browse`: some cleanups

andlaus:
most of these cleanups are prompting-related:

- the validation function does not need the parameter object; the DOP and whether it is mandatory or not is sufficient.
- empty input for parameters where it is valid (i.e., strings and byte arrays) are handled properly (hopefully).
- some general cleanups for `prompt_primitive_parameter_value()`.
- enable specifying empty values for parameters exhibiting a default value.
- in lists, preselect the default value.
- remove unused function `encode_message_from_string_values()`

### DIFF
--- a/odxtools/cli/browse.py
+++ b/odxtools/cli/browse.py
@@ -232,7 +232,7 @@ def encode_message_interactively(codec: Request | Response,
 
     param_values: ParameterValueDict = {}
     answered_request = b''
-    if has_settable_param:
+    if has_settable_param or has_matching_request_param:
         # Ask whether user wants to encode a message
         if ask_user_confirmation:
             encode_message_prompt = [{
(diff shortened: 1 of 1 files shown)


odxtools/cli/browse.py
```diff
@@ -53,81 +51,119 @@ def _convert_string_to_bytes(string_value: str) -> bytes:
...
-            param_prompt[0]["choices"] = choices
+            param_prompt[0]["choices"] = texttable_choices
 
+    # query user for answer
     answer = IP_prompt(param_prompt)
     raw_answer = answer.get(parameter.short_name)
-    if raw_answer == "" and not parameter.is_required:
+
+    if raw_answer in ("", None):
+        if raw_answer == "" and parameter.is_required:
```
kayoub5, line 124: `raw_answer in ("", None)` or `raw_answer == ""` ?

odxtools/cli/browse.py
```diff
@@ -53,81 +51,119 @@ def _convert_string_to_bytes(string_value: str) -> bytes:
...
+            # value of the parameter's data type (e.g., b'' for byte fields).
+            if parameter.physical_type.base_data_type is not None:
+                try:
+                    return _convert_string_to_odx_type("", parameter.physical_type.base_data_type)
+                except (OdxError, ValueError, TypeError):
+                    return None
+            return None
+
+        # For optional parameters, determine the default value and ask
+        # the user whether they meant the default or the empty value.
```
kayoub5, line 135: Since you can catch keyboard keys, it would be more user friendly to have a keyboard shortcut for "skip" parameter

odxtools/cli/browse.py
andlaus: I'm also always confused about this. I think `("", None)` is more failsafe because it allows `None` to be specified as a list value....

odxtools/cli/browse.py
andlaus: I looked into this: Specifying a shortcut only works for lists and even for these, I could not make it work satisfactorily...

PR #522, opened 2026-09-11
`browse`: add proper support for specifying parameters using `DtcDop`

andlaus:
So far, DTCs were handled as integer parameters which made them very unintuitive to specify. Now, an entry from the list of the DTCs specified by the DOP can be selected.

### DIFF
--- a/odxtools/cli/browse.py
+++ b/odxtools/cli/browse.py
@@ -12,8 +12,10 @@ from ..database import Database
 from ..dataobjectproperty import DataObjectProperty
 from ..diaglayers.diaglayer import DiagLayer
 from ..diaglayers.hierarchyelement import HierarchyElement
+from ..diagnostictroublecode import DiagnosticTroubleCode
 from ..diagservice import DiagService
 from ..dopbase import DopBase
+from ..dtcdop import DtcDop
 from ..environmentdatadescription import EnvironmentDataDescription
 from ..exceptions import OdxError, odxraise, odxrequire
 from ..field import Field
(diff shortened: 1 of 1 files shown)


kayoub5:
the logic looks fine, but it would be best to unit test it

andlaus:
the problem with `browse` and unit tests is that it is an interactive tool. one can use MagicMock to get around this, but I don't like this because it "calcifies" the current state of the art and I think that interactive tools should be relatively open for experimentation. Anyway if you *really* want a unit test anyway, I can ask my LLM to create one...

kayoub5:
> the problem with `browse` and unit tests is that it is an interactive tool. one can use MagicMock to get around this, but I don't like this because it "calcifies" the current state of the art and I think that interactive tools should be relatively open for experimentation. Anyway if you _really_ want a unit test anyway, I can ask my LLM to create one...

some functions are pure, and have no interactive part
* _validate_chosen_value
* _convert_string_to_bytes
* _convert_string_to_odx_type

The massive prompt_primitive_parameter_value function can be split in two functions (before/after `answer = IP_prompt(param_prompt)` call)

Another approach is to pass `IP_prompt` as argument to the functions (no mocking needed)
browse pass normal `IP_prompt`, unit test pass a function that return items from hard-coded list

seif-galal:
@andlaus @kayoub5 I created a test covering the relevant `DtcDop` handling in `browse`, including the conversion logic. It should provide a useful regression test without mocking the interactive UI itself. Let me know if i should add it to the PR.

andlaus:
I just uploaded what my LLM spit out: https://github.com/andlaus/odxtools/pull/3 . I haven't gone through it with a fine comb, but from skimming over it, the result looks pretty good IMO. That said, just because we *can* do this is IMO not indicative of whether we *should*...

seif-galal:
> I just uploaded what my LLM spit out: [andlaus#3](https://github.com/andlaus/odxtools/pull/3) . I haven't gone through it with a fine comb, but from skimming over it, the result looks pretty good IMO. That said, just because we _can_ do this is IMO not indicative of whether we _should_...

Agreed. Could also add edge cases for `_convert_string_to_bytes()`, especially `"00"` / `"0001"`, to confirm the leading-zero behavior is intended. But yeah, looks quite good

andlaus:
@kayoub5: How shall we proceed here? merge as is, add a full unit test or reduce the proposed test to the non-interactive parts of `browse`?

kayoub5:
reduce the proposed test to the non-interactive parts, that should be enough to at least have some code coverage

andlaus:
okay, done. There are a few uses of `MagicMock` left, but IMO it is not too bad because they are pretty simple...

kayoub5 (approved):

PR #523, opened 2026-09-15
`browse`: implement support for prompting more parameter values

andlaus:
This is the last PR in the `browse` improvement series. With this, `browse` should hopefully be able to specify values for any parameter using any DOP. The only exception is `NRC-CONST`, but these are IMO pretty exotic, quite hacky and difficult to implement in `browse`. (IOW, not worth it as far as I am concerned.)

Note that this PR implements multiple DOP and parameter types at once. If desired, I can split it up, but given that all of these follow pretty much the same template, it IMO is less work for everybody involved to do them all at once.


kayoub5:
The code has become quite tangled and difficult to follow.

It looks correct, but unmaintainable and untestable

This means you are starting the slippery slope of:
* Higher bug risk — changes can introduce regressions that aren't caught automatically, no linter will help you if you use `getattr`
* Slower development — more manual testing and debugging are needed.
* Harder maintenance — because the code behavior is difficult to verify and visually too complex to follow.

kayoub5:
There are few patterns that you can apply, to make code testable and more maintainable

* Make `IP_prompt` an input function to the functions/wrapping class
* Use Visitor design pattern https://refactoring.guru/design-patterns/visitor/python/example

andlaus:
Ok, i removed the use of `getattr()`: https://github.com/mercedes-benz/odxtools/pull/523/changes/ac3678fdf44d7f6c29eef451155243d645d820fc .

Regarding the increase in complexity: I think that, while this PR adds quite a few lines, it does not make the `browse` tool meaningfully more complicated/entangled. (IMO, ~700 lines of code is still manageable in a single file.) Anyway, fixing this would be another PR or two...

andlaus:
(also, IMO ´browse´ is not a critical part of `odxtools` because I doubt that many people build their own stuff on top of it...)

kayoub5:
> Ok, i removed the use of `getattr()`: [ac3678f](https://github.com/mercedes-benz/odxtools/commit/ac3678fdf44d7f6c29eef451155243d645d820fc) .
> 
> Regarding the increase in complexity: I think that, while this PR adds quite a few lines, it does not make the `browse` tool meaningfully more complicated/entangled. (IMO, ~700 lines of code is still manageable in a single file.) Anyway, fixing this would be another PR or two...

Low number of lines true, but extremely high Cognitive Complexity/ Cyclomatic Complexity

See https://www.baeldung.com/java-cognitive-complexity

andlaus:
> Low number of lines true, but extremely high Cognitive Complexity/ Cyclomatic Complexity

Isn't the additional prompting functionality nicely encapsulated in quite self-contained functions which are not exactly rocket science? (`prompt_field_parameter_value()`, `prompt_env_data_desc_parameter_value()`, etc).

kayoub5:
> > Low number of lines true, but extremely high Cognitive Complexity/ Cyclomatic Complexity
> 
> Isn't the additional prompting functionality nicely encapsulated in quite self-contained functions which are not exactly rocket science? (`prompt_field_parameter_value()`, `prompt_env_data_desc_parameter_value()`, etc).

Typical recommendation is to have a complexity of 15 or less (some teams make it even as strict as 10)
| Function | Complexity |
|---|---|
| `prompt_primitive_parameter_value` | 57 |
| `encode_message_interactively.has_settable_or_matching_request_param` | 27 |
| `browse` | 26 |
| `_validate_chosen_value` | 22 |
| `prompt_all_parameter_values` | 16 |

andlaus:
> Typical recommendation is to have a complexity of 15 or less (some teams make it even as strict as 10)

IMHO this metric is flawed: it is perfectly fine to see these these functions as a black box from the outside, and it is not too difficult to understand their inner workings if something goes belly-up because they do not call any functions which cannot be considered as a black box. (is it even possible to implement complex functionality like a PDE solver adhering to the metric not only in letter but in spirit?)

andlaus:
(also note that none of the functions which you outlined were added by this PR.)

andlaus:
A part of odxtools where this kind of complexity actually is a relevant problem IMO is the encoding and decoding functionality. That said, I have spend most of 2023 and 2024 trying to improve it, but I have to confess that I only partially succeeded. I believe that the fundamental issue there is the complexity of the specification...

kayoub5 (approved):

PR #524, opened 2026-09-16
Maintenance

andlaus:
This PR falls under the general maintenance category: It enables `test_browse` on github actions, fixes a recent regression of that test (the `browse` utility was always fine, but the test got out of sync and this was not discovered because GH actions did not run it routinely), and it fixes a few typos/grammos in `README.md` as well as  synchronizing it with the current state of affairs.


andlaus:
I suppose there is not too much point in a review; merging...

PR #526, opened 2026-09-22
Fix English in source code

andlaus:
This fixes typos, grammos and language style (as well as a few factual errors) in code comments, string literals and jinja templates.

This was heavily AI-assisted using github copilot-CLI.


andlaus:
IMO there is not too much point in a review here. @kayoub5, if you disagree with some of the changes, please drop some comments.

merging.

